PR #228, opened 2024-12-12
BM-305: Integrate SetInclusionReceipt

### LINKED ISSUE #121: Add support in the client SDK to fetch the root of the `SetReceipt` proofs from the chain events

capossele:
Closes #121

### DIFF
--- a/crates/broker/src/rpc_retry_policy.rs
+++ b/crates/broker/src/rpc_retry_policy.rs
@@ -20,8 +20,9 @@ impl RetryPolicy for CustomRetryPolicy {
     fn should_retry(&self, error: &TransportError) -> bool {
         let should_retry = match error {
             TransportError::Transport(TransportErrorKind::Custom(err)) => {
-                let msg = err.to_string();
-                msg.contains("os error 104")
+                // easier to match against the debug format string because this is what we see in the logs
+                let err_debug_str = format!("{:?}", err);
+                err_debug_str.contains("os error 104") || err_debug_str.contains("reset by peer")
             }
             _ => false,
         };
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -49,7 +49,7 @@ hex = { version = "0.4", default-features = false, features = ["alloc"] }
 postcard = { version = "1.0" }
 rand = { version = "0.9" }
 reqwest = "0.12"
-risc0-aggregation = "0.2.1"
+risc0-aggregation = { version = "0.2" }
```
Wollac, line 52:
```suggestion
risc0-aggregation = { version = "0.2.2" }
```
Looking at the tag of the guest-set-builder, it seems that is locked to v0.2.2 so we should assure here to at least use version 0.2.2 risc0-aggregator (maybe even fix it to that exact version).

Cargo.toml
```diff
@@ -59,7 +59,7 @@ risc0-zkp = { version = "1.2" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 serde_yaml = "0.9"
-guest-set-builder = { git = "https://github.com/risc0/risc0-ethereum", tag = "aggregation-v0.2.1" }
+guest-set-builder = { git = "https://github.com/risc0/risc0-ethereum", tag = "aggregation-v0.2.2" }
```
Wollac, line 62: I don't understand, why this points to a special aggregation tag instead of to the general latest version of risc0-ethereum

Wollac (approved):

Cargo.toml
capossele: I think the idea was to keep it more independent from the rest of the repo

Wollac (approved):

PR #248, opened 2025-01-09
BM-79: Chain monitor service

austinabell:
Closes BM-79

Just used to monitor chain head for now.

In testing, seems like the current block number is polled constantly, so maybe worth considering switching this from being lazily polled for simplicity or increasing the poll time of this specifically instead of tying it in with `status_poll_ms`.


diamondval74687:
git checkout release-0.12

PR #266, opened 2025-01-23
BM-474: broker stress test

austinabell:
Just opening for visibility, not ready to be reviewed yet

Setting up some logic to be able to test the broker with random inputs to shake out inconsistent errors. Primary goal is to resolve sqlite edge case.

What I am planning to change about this:
- [x] Modify broker test utils to allow configuring things like the sqlite URL
- [x] Create wrapper around anvil node to chaos test with configurable latency and status codes to simulate network failures
- [ ] Query chain state to get a pulse on orders fulfilled, failed, and queued to have more visibility into the test results


crates/broker-stress/src/main.rs
```diff
@@ -0,0 +1,224 @@
...
+    Ok((broker_task, config_file))
+}
+
+/// Basic handler that responds with a static string.
+async fn serve_elf() -> &'static [u8] {
+    ECHO_ELF
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
```
mothran, line 137: Any reason to have this be a new crate vs a `example` in broker crate?

crates/broker-stress/src/toxiproxy.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    // Try to start toxiproxy
+    let mut child = Command::new("toxiproxy-server")
+        .args(["-port", "8474"])
+        .stdout(Stdio::piped())
+        .spawn()
+        .map_err(|e| {
+            if e.kind() == io::ErrorKind::NotFound {
+                io::Error::new(io::ErrorKind::Other, "Toxiproxy is not installed or not in the PATH. See https://github.com/Shopify/toxiproxy")
```
mothran, line 30: I think a README in this crate might be helpful to get someone spun up with using this tooling.

mothran (approved):

crates/broker-stress/src/main.rs
austinabell: iirc my rationale was that this is not something that any external user would be running, and having as an example might be confusing to someone browsing. I don't feel strongly or have any reason I remember for why it can't be an example -- so happy to switch if an opinion is shared in favour.

PR #304, opened 2025-02-06
BM-524: Add journal size check after preflight

willemolding:
Closes #294 

- Adds a new field to the config for setting a max journal size
- Checks after preflight that the resulting journal for an order is less than this, otherwise drops it
- Adds a test that an order is correctly dropped
- Adds a new config field for max batch journal size
- Triggers finalizing a batch once this is exceeded

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2468,6 +2468,7 @@ dependencies = [
  "bytemuck",
  "chrono",
  "clap",
+ "elsa",
  "futures",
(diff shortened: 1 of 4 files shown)


willemolding:
@capossele I went with adding a new trigger to finalize a batch, which is when it exceeds a configurable threshold. Ready for re-review

capossele (approved):

PR #306, opened 2025-02-07
BM-459: Explore options for delivering unaggregated proofs

capossele:
**Part 1**

This first pass only covers the contracts side of things. The idea is to add an optional field `Selector` to the `Requirements` of a proof request. The `selector` is also committed by the  assessor in the `AssessorJournal` (it only commits the ones actually required as otherwise any selector would suffice).  **Currently the broker skips any request that specifies a selector.**

**Part 2** should instead focus on the prover side: the broker must check that it can support the required selector (if set) and generate the seal accordingly


contracts/src/BoundlessMarket.sol
```diff
@@ -214,52 +216,78 @@ contract BoundlessMarket is
...
             claimDigests[i] = ReceiptClaimLib.ok(fills[i].imageId, sha256(fills[i].journal)).digest();
+            // Verification that the provided seal matches the required selector.
+            for (uint8 j = 0; j < assessorFill.selectors.indices.length; j++) {
+                if (assessorFill.selectors.indices[j] == i) {
+                    if (assessorFill.selectors.values[j] != bytes4(fills[i].seal[0:4])) {
+                        revert SelectorMismatch(assessorFill.selectors.values[j], bytes4(fills[i].seal[0:4]));
+                    }
+                    break;
+                }
+            }
```
willpote, line 253: Hmm we should really try and avoid O(n^2). Not sure which of these would be more gas efficient, but a couple ideas:

1/
Can we guarantee that the indices committed to the asssesor are sorted? If so I think we can just do something like:
```
j = 0
for (i=0; i<fills.length; i++)
    if (assessorFill.selectors.indices[j] == i) {
        // check seal
        j++;
    }
    // VERIFIER stuff
}
```


2/
We could have two loops where we first iterative fills and verifiy, then iterator assessor selectors and check those.

3/
Commit something to selectors array for every fill (some null value if theres no selector requirement)

contracts/src/types/Fulfillment.sol
```diff
@@ -3,6 +3,7 @@
 // All rights reserved.
 pragma solidity ^0.8.20;
 
+import {Selector} from "./Selector.sol";
```
willpote, line 6: Not used

contracts/src/types/ProofRequest.sol
```diff
@@ -41,7 +42,8 @@ library ProofRequestLibrary {
             InputLibrary.INPUT_TYPE,
             OfferLibrary.OFFER_TYPE,
             PredicateLibrary.PREDICATE_TYPE,
-            RequirementsLibrary.REQUIREMENTS_TYPE
+            RequirementsLibrary.REQUIREMENTS_TYPE,
+            SelectorLib.SELECTOR_TYPE
```
willpote, line 46: Looks like something from an old version?

contracts/src/types/Selector.sol
```diff
@@ -0,0 +1,108 @@
...
+using SelectorsLib for Selectors global;
+
+/// @title Selector - A type for handling optional bytes4 values
+/// @notice Provides a safe way to handle optional values in Solidity
+struct Selector {
+    /// @notice The wrapped value, only valid when isSome is true
+    bytes4 value;
+    /// @notice Indicates whether the Option contains a value (true) or is None (false)
+    bool isSome;
+}
```
willpote, line 16: Does this mean 0 is a valid selector? Curious why we need this

contracts/src/types/Selector.sol
```diff
@@ -0,0 +1,108 @@
...
+struct Selectors {
+    /// @notice Indices where selectors are required
+    uint8[] indices;
+    /// @notice The actual required selectors, corresponding to indices
+    bytes4[] values;
+}
+
+/// @title SelectorLib - Helper library for working with Selector types
+/// @notice Provides utility functions for creating and manipulating Selector values
+library SelectorLib {
```
willpote, line 29: Are the utility functions here and in SelectorsLib just for tests? 

If so we should probably just move them to tests files just to reduce boilerplate in the main code base. Also not sure if the compiler is smart enough to remove unused code so wondering if its inflating bytecode size?

contracts/src/BoundlessMarket.sol
willpote: Also would be good to have some gas tests that test this for large batch sizes with lots of selector predicates, something similar to the Benchmark tests we have now

contracts/src/BoundlessMarket.sol
```diff
@@ -214,52 +216,78 @@ contract BoundlessMarket is
...
+                    selectors: assessorFill.selectors,
+                    root: claimDigest,
+                    prover: assessorFill.prover
+                })
+            )
+        );
+        // Verification that the provided seal matches the required selector.
+        if (assessorFill.selectors.indices.length == 1 && assessorFill.selectors.values[0] != bytes4(fill.seal[0:4])) {
+            revert SelectorMismatch(assessorFill.selectors.values[0], bytes4(fill.seal[0:4]));
+        }
```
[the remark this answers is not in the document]
capossele: So here we check that the indices length is 1 because we are in the fulfill method (no batch), and if that is 1 it means the request has defined the selector to match. We match that against the fill.seal because that is the (application) seal that must start with the specified selector

contracts/src/BoundlessMarket.sol
capossele: Good call, I've dropped the second loop as well as added a benchmark

contracts/src/types/Selector.sol
capossele: Yeah, my idea was to enable testing it with the MockVerifier that typically uses 0 as selector. But tbh can be set with any selector as well, so I've dropped this type in favor of just a `bytes4`

contracts/src/types/Selector.sol
capossele: Yeah I've dropped all of it and added to TestUtils

contracts/src/BoundlessMarket.sol
```diff
@@ -249,19 +249,28 @@ contract BoundlessMarket is
...
-                        revert SelectorMismatch(assessorFill.selectors.values[j], bytes4(fills[i].seal[0:4]));
-                    }
-                    break;
+
+        uint256 fillsLength = fills.length;
+        uint256 selectorsLength = assessorFill.selectors.indices.length;
+        uint8 selectorIdx = 0;
+
+        for (uint256 i = 0; i < fillsLength; i++) {
+            // Cache the current fill to avoid multiple calldata lookups.
```
willpote, line 258: Nit: don't think we need these comments here and on L266, this is pretty standard

contracts/src/BoundlessMarket.sol
```diff
@@ -249,19 +249,28 @@ contract BoundlessMarket is
...
-            claimDigests[i] = ReceiptClaimLib.ok(fills[i].imageId, sha256(fills[i].journal)).digest();
-            // Verification that the provided seal matches the required selector.
-            for (uint8 j = 0; j < assessorFill.selectors.indices.length; j++) {
-                if (assessorFill.selectors.indices[j] == i) {
-                    if (assessorFill.selectors.values[j] != bytes4(fills[i].seal[0:4])) {
-                        revert SelectorMismatch(assessorFill.selectors.values[j], bytes4(fills[i].seal[0:4]));
-                    }
-                    break;
+
+        uint256 fillsLength = fills.length;
```
willpote, line 253: Can move this up and use it for initializing the arrays

contracts/src/BoundlessMarket.sol
```diff
@@ -249,19 +249,28 @@ contract BoundlessMarket is
...
+
+            requestDigests[i] = fill.requestDigest;
+            claimDigests[i] = ReceiptClaimLib.ok(fill.imageId, sha256(fill.journal)).digest();
+
+            // If the current index is flagged for selector verification, process it.
+            if (selectorIdx < selectorsLength && assessorFill.selectors.indices[selectorIdx] == i) {
+                // Cache the computed bytes4 value from seal.
+                bytes4 fillSealSelector = bytes4(fill.seal[0:4]);
+                if (assessorFill.selectors.values[selectorIdx] != fillSealSelector) {
+                    revert SelectorMismatch(assessorFill.selectors.values[selectorIdx], fillSealSelector);
```
willpote, line 269: Nit: here I would actually recommend we don't cache, in order to make the happy case path cost less gas

contracts/src/types/Selector.sol
```diff
@@ -3,18 +3,6 @@
...
-struct Selector {
-    /// @notice The wrapped value, only valid when isSome is true
-    bytes4 value;
-    /// @notice Indicates whether the Option contains a value (true) or is None (false)
-    bool isSome;
-}
-
 /// @title Selectors - A sparse representation of bytes4 selectors
 /// @notice Efficiently stores selectors when most are zero
 struct Selectors {
```
willpote, line 8: Nit rename file to Selectors

willpote (approved):

contracts/src/BoundlessMarket.sol
```diff
@@ -224,78 +225,113 @@ contract BoundlessMarket is
...
+                AssessorJournal({
+                    requestDigests: requestDigests,
+                    selectors: assessorFill.selectors,
+                    root: claimDigest,
+                    prover: assessorFill.prover
+                })
+            )
+        );
+        // Verification that the provided seal matches the required selector.
+        if (assessorFill.selectors.length == 1 && assessorFill.selectors[0].value != bytes4(fill.seal[0:4])) {
```
nategraf, line 239: There is a weird case here in that if `assessorFill.selectors.length > 1`, this check is not applied.

contracts/src/BoundlessMarket.sol
```diff
@@ -224,78 +225,113 @@ contract BoundlessMarket is
...
+            requestDigests[i] = fill.requestDigest;
+            claimDigests[i] = ReceiptClaimLib.ok(fill.imageId, sha256(fill.journal)).digest();
+
+            // If the current index is flagged for selector verification, process it.
+            if (selectorIdx < selectorsLength && assessorFill.selectors[selectorIdx].index == i) {
+                if (assessorFill.selectors[selectorIdx].value != bytes4(fill.seal[0:4])) {
+                    revert SelectorMismatch(assessorFill.selectors[selectorIdx].value, bytes4(fill.seal[0:4]));
+                }
+                selectorIdx++;
+            }
```
nategraf, line 273: If `assessorFill.selectors[selectorIdx]` is not sorted properly, some of the selector constraints will never be applied.

contracts/src/types/FulfillmentAssessor.sol
```diff
@@ -0,0 +1,23 @@
...
+pragma solidity ^0.8.20;
+
+import {Selector} from "./Selector.sol";
+import {RequestId} from "./RequestId.sol";
+
+using FulfillmentAssessorLibrary for FulfillmentAssessor global;
+
+/// @title FulfillmentAssessor Struct and Library
+/// @notice Represents the information posted by the prover that fulfills an Assessor.
+struct FulfillmentAssessor {
```
nategraf, line 13: `FulfillmentAssessor` doesn't really make sense to me, although this struct is hard to name since it is the proof and _part_ of the `AssessorJournal` (the parts we can't calculate from the other fulfillment args). Here is the best suggestion I could think of in a minute or two of thinking. 
```suggestion
/// @notice Represents the output of the assessor and proof of correctness, allowing request fulfillment.
struct AssesorReceipt {
```

capossele: Yeah bad naming, also the only reason why it has been created was to silence clippy warning about the number of arguments. AssessorReceipt feels also not ideal but better than the other

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1872,6 +1967,38 @@ contract BoundlessMarketBench is BoundlessMarketTest {
...
+        benchFulfillBatchWithSelector(32, "032");
+    }
+
+    function testBenchFulfillBatchWithSelector064() public {
+        benchFulfillBatchWithSelector(64, "064");
+    }
+
+    function testBenchFulfillBatchWithSelector128() public {
+        benchFulfillBatchWithSelector(128, "128");
+    }
```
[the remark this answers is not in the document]
willpote: I asked for this as in the first version the selector requirement path was quite a lot more expensive at scale than the non-selector path, and it wasn't clear from the gas tests. Imo worth keeping something here.

contracts/src/BoundlessMarket.sol
```diff
@@ -224,78 +225,113 @@ contract BoundlessMarket is
...
-        for (uint256 i = 0; i < fills.length; i++) {
-            requestDigests[i] = fills[i].requestDigest;
-            claimDigests[i] = ReceiptClaimLib.ok(fills[i].imageId, sha256(fills[i].journal)).digest();
-            VERIFIER.verifyIntegrity{gas: FULFILL_MAX_GAS_FOR_VERIFY}(Receipt(fills[i].seal, claimDigests[i]));
+        uint256 fillsLength = fills.length;
+        // We can't handle more than 65535 fills in a single batch.
+        // This is a limitation of the current Selectors implementation, and can be increased in the future.
+        if (fillsLength > type(uint16).max) {
+            revert BatchSizeExceedsLimit(fillsLength, type(uint16).max);
+        }
```
willpote, line 254: Why do we have this limitation?
Also can we just check this in the assessor?

contracts/src/BoundlessMarket.sol
```diff
@@ -224,78 +225,114 @@ contract BoundlessMarket is
...
+                AssessorJournal({
+                    requestDigests: requestDigests,
+                    selectors: assessorReceipt.selectors,
+                    root: claimDigest,
+                    prover: assessorReceipt.prover
+                })
+            )
+        );
+        // Verification that the provided seal matches the required selector.
+        if (assessorReceipt.selectors.length == 1 && assessorReceipt.selectors[0].value != bytes4(fill.seal[0:4])) {
```
nategraf, line 239: It's true that there would need to be a bug in the assessor, or a vulnerability somewhere in the proving system for `> 0` and `== 1` to be functionally different here, but I still think we should use the slightly more robust version of this check, and add a note.
```suggestion
        // NOTE: Assessor guest ensures that the number of selectors <= the number of request digests in the journal.
        if (assessorReceipt.selectors.length > 0 && assessorReceipt.selectors[0].value != bytes4(fill.seal[0:4])) {
```

nategraf (approved):
Looks good to me. I did provide one suggestion here, and one in a PR against this branch. Both are intended to make the make the logic a bit more robust against errors between the contract and verifier code. The PR happens to also make things a little easier to read and reduces gas and contract size by a very small amount.

PR #307, opened 2025-02-07
BM-96: Add precheck if prover has image before download and upload

willemolding:
- Adds a new function to the prover trait for checking if an image_id already exists
- Implements this for MockProver and BonsaiProver
- Test that the downloading is skipped on the second task with the same image_id


willemolding:
Ok looks to be working with Bonsai (and presumably Bento) as well as with the mock prover in tests

crates/broker/src/provers.rs
```diff
@@ -135,6 +136,12 @@ impl Prover for Bonsai {
         Ok(self.client.upload_img(image_id, image).await.map(|_| ())?)
     }
 
+    async fn has_image(&self, image_id: &str) -> Result<bool, ProverError> {
+        // upload_img will early return true if the image_id already exists
+        // so calling this with empty data is a way to check this
+        Ok(self.client.upload_img(image_id, vec![]).await?)
```
capossele, line 142: What happens though if the image does not exist yet? Would this fail as we are sending an empty Vec, or will it somehow save it?

willemolding: If it doesn't exist then bento/bonsai will abort when it checks that the calculated image_id matches the passed one. So no worry about saving an empty image

PR #321, opened 2025-02-18
BM-534: Add configurable warning and error logs for broker balance

willemolding:
Closes issue: BM-535 Alert on low balance

Adds a new provider layer which can be configured to log warnings and errors on balance thresholds. Implementing this as a layer on the RPC Provider means all calls anywhere in the broker will have these checks added automatically.

- Adds layer implementation
- Adds new config fields to set threshold
- Adds layer when initializing the broker provider


willpote:
Nice! We'd also like to use this for crates/order-generator and crates/slasher without having them depend on the broker. Maybe this should live in its own util crate? Or in boundless-market?

willemolding:
I think its own crate makes the most sense

crates/balance-alerts-layer/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    ///
+    /// See [`send_transaction`](Self::send_transaction) for more details.
+    async fn send_raw_transaction(
+        &self,
+        encoded_tx: &[u8],
+    ) -> TransportResult<PendingTransactionBuilder<T, Ethereum>> {
+        let res = self.inner.send_raw_transaction(encoded_tx).await;
+        let balance = self.inner.get_balance(self.config.watch_address).await?;
+
+        if balance < self.config.error_threshold.unwrap_or(U256::MAX) {
```
willpote, line 104: Shouldn't this be `unwrap_or(0)`? If config is not set we wouldn't want to log errors/warnings

Would be good to have a test for the no config case also

crates/order-generator/src/bin/zeth.rs
```diff
@@ -91,6 +91,10 @@ struct Args {
     offchain: bool,
     #[clap(long, default_value = "3")]
     max_retries: u32,
+    #[clap(long, value_parser = parse_ether, default_value = "1")]
```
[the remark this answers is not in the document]
willpote: Seems reasonable to me!

willpote (commented):
Think we may have missed this in the original task, but we also need a way to alert on low stake token balances for the broker. I'm guessing it will need a different solution so its fine for a separate PR, but just wanted to flag

willemolding:
As I understand this would be for stake tokens locked in the market?

If so I think we want a different solution for that since it doesn't relate to every transaction like gas balance does

crates/balance-alerts-layer/src/lib.rs
willemolding: Yep that does makes sense rather than for always to alert when there is no config. 

Will update and add a test

willemolding:
@willpote Updated to default to no warnings/errors and adds a test showing this

crates/balance-alerts-layer/src/lib.rs
willemolding: @willpote Updated to default to no warnings/errors and adds a test showing this

crates/order-generator/src/bin/zeth.rs
```diff
@@ -115,8 +122,16 @@ async fn main() -> Result<()> {
     let args = Args::parse();
 
     let wallet = EthereumWallet::from(args.private_key.clone());
+
+    let balance_alerts_layer = BalanceAlertLayer::new(BalanceAlertConfig {
+        watch_address: wallet.default_signer().address(),
+        warn_threshold: args.warn_balance_below,
+        error_threshold: args.error_balance_below,
+    });
```
willpote, line 130: There are actually 2 order generators in this package, theres another one in main.rs that we should update

willpote (approved):
Thanks for doing this! Whats the plan for checking stake balances? Note we only need to check stake in the broker, so the solution could be less generic.

crates/boundless-market/src/client.rs
```diff
@@ -570,12 +580,14 @@ impl Client<Http<HttpClient>, ProviderWallet, BuiltinStorageProvider> {
         set_verifier_address: Address,
         order_stream_url: Option<Url>,
         storage_provider: Option<BuiltinStorageProvider>,
+        balance_alerts: Option<BalanceAlertConfig>,
```
austinabell, line 583: Opinion: maybe this should be done as a builder pattern to avoid breaking changes? I feel like this `from_parts` public method is already unweildy. Also possible we can just refactor and address in the future, just noting.

crates/broker/src/bin/broker.rs
```diff
@@ -31,8 +34,15 @@ async fn main() -> Result<()> {
     );
     let client = RpcClient::builder().layer(retry_layer).http(args.rpc_url.clone()).boxed();
 
+    let balance_alerts_layer = BalanceAlertLayer::new(BalanceAlertConfig {
+        watch_address: wallet.default_signer().address(),
+        warn_threshold: config.market.balance_warn_threshold.and_then(|s| parse_ether(&s).ok()),
+        error_threshold: config.market.balance_error_threshold.and_then(|s| parse_ether(&s).ok()),
```
austinabell, line 40: Silently ignoring the parse failure seems like a footgun. Perhaps at least warn if invalid? This is just on startup so also don't seen an issue with erroring if invalid?

willemolding: Thanks for the reminder. This is fixed in its own branch which also fixes another instance of this same issue https://github.com/boundless-xyz/boundless/pull/411/files

PR #324, opened 2025-02-19
BM-552: fix: avoid upload image id check on empty bytes case

austinabell:
This seems to match the behavior needed for `Prover::has_image`, and was not ideal to make the http calls directly to avoid this hack so decided to just handle this case specifically for now

Also using the following env vars to hit testnet:
```
BOUNDLESS_MARKET_ADDR=0x69c7943DA0D7e45D44Bd0cE7a2412DCdAe423788
SET_VERIFIER_ADDR=0xEf0A93B2310d52358F1eCA0C946aD7D25596e7dd
```


capossele (approved):

austinabell:
second guessing that this is the correct fix. Need to double check the bonsai backend with this

austinabell:
Closing in favour of #327

PR #334, opened 2025-02-21
BM-565: fix: order stream connections mapping

austinabell:
There is an edge case currently where a broker address will be marked in the db as connected, even though there is no active websocket connection.

There are at least a few ways this could have happened:
- Write to DB to lower the connections failed
- Get balance on chain failed, panicked, which does not update the db for that pending connection
- Service crashed for some other reason, and these marked connections in the db aren't handled in this case

Decided to just simplify the logic such that all the data around pending connections is just kept in memory, as there doesn't seem to be any reason to persist it. I kept the pending connections in a separate map to reduce lock contention on the actual connections (don't want people spamming to DOS the active connections). The logic also is changed such that it will not overwrite an existing active connection, so less worried about this having race conditions to replace connections.

Tested on test suites (fixed justfile as well) and tested manually as best I could, although integration tests for this don't really exist _yet_.

### DIFF
--- a/crates/order-stream/src/lib.rs
+++ b/crates/order-stream/src/lib.rs
@@ -2,7 +2,7 @@
 //
 // All rights reserved.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use alloy::{
     primitives::{utils::parse_ether, Address, U256},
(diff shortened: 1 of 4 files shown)


zeroecco (approved):
as long as this works it looks okay to me.

PR #335, opened 2025-02-21
BM-476: Callbacks

willpote:
* Callbacks are only ever executed once, the first time a request is fulfilled
* Add a Callback interface and base contract for apps.
* In the non-locked path, callbacks are stored along with the price, when the request is "priced"


contracts/src/BoundlessMarket.sol
```diff
@@ -358,11 +366,29 @@ contract BoundlessMarket is
...
+        // the new request.
         if (lock.lockDeadline >= block.number) {
-            return _payLockedCurrently(id, assessorProver, lock.prover, lock.price, lock.stake);
+            if (lock.fingerprint != bytes8(requestDigest)) {
+                revert RequestLockFingerprintDoesNotMatch({
+                    requestId: id,
+                    provided: bytes8(requestDigest),
+                    locked: lock.fingerprint
+                });
+            }
```
willpote, line 383: Moving this here fixes a bug that was missed in a previous PR

contracts/src/types/RequestLock.sol
```diff
@@ -49,6 +49,13 @@ struct RequestLock {
     /// based on request digest instead of index. As a friction, this would introduce a second
     /// user-facing concept of what identifies a request.
     bytes8 fingerprint;
+    ///
+    /// Storage slot 3
+    ///
+    /// @notice The address of the contract to call when the proof is delivered. Address is 0 if no callback is required.
+    address callbackAddress;
+    /// @notice If a callback is requested, the maximum gas to use for the callback.
+    uint96 callbackGasLimit;
```
nategraf, line 58: We could put this into the assessor journal instead, to save on storage costs.

contracts/src/types/FulfillmentContext.sol
```diff
@@ -0,0 +1,77 @@
...
+/// @dev This struct is designed to be packed into a single uint256 for efficient transient storage
+struct FulfillmentContext {
+    /// @notice Boolean set to true to indicate the request was validated
+    bool valid;
+    /// @notice The validated price for the request
+    uint96 price;
+    /// @notice The address to call during fulfillment. If set to address(0), no callback is performed
+    address callback;
+    /// @notice The gas limit to use for the callback. Must be non-zero if callback is set
+    uint96 callbackGaslimit;
```
nategraf, line 19: If we put the callback info into the assessor journal, we shouldn't need this here.

contracts/src/IBoundlessMarketCallback.sol
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2025 RISC Zero, Inc.
+//
+// All rights reserved.
+pragma solidity ^0.8.24;
+
+/// @title IBoundlessMarketCallback
+/// @notice Interface for handling proof callbacks from BoundlessMarket with proof verification
+/// @dev Inherit from this contract to implement custom proof handling logic for BoundlessMarket proofs
+interface IBoundlessMarketCallback {
+    /// @notice Handles submitting proofs with RISC Zero proof verification
```
nategraf, line 10: I don't understand this description.

contracts/src/IBoundlessMarketCallback.sol
```diff
@@ -0,0 +1,16 @@
...
+    /// @dev If not called by BoundlessMarket, the function MUST verify the proof before proceeding.
```
nategraf, line 11: Even if the BoundlessMarket is the source, we should be strongly encouraging receivers to verify the seal. If the BoundlessMarket is compromised, we want to avoid this potentially leading to a soundness/safety failure in consuming applications (e.g. it should not be possible to drain the funds of a validity rollup that uses Boundless).

This does cost extra gas, however it is not a huge amount since the proof will usually be a Merkle inclusion path.

We will also want to be extra careful from now on about what external calls the BoundlessMarket contract makes.

contracts/src/BoundlessMarketCallback.sol
```diff
@@ -0,0 +1,53 @@
...
+    constructor(IRiscZeroVerifier verifier, address boundlessMarket) {
+        VERIFIER = verifier;
+        BOUNDLESS_MARKET = boundlessMarket;
+    }
+
+    /// @inheritdoc IBoundlessMarketCallback
+    function handleProof(bytes32 imageId, bytes calldata journal, bytes calldata seal) public {
+        if (msg.sender == BOUNDLESS_MARKET) {
+            _handleProof(imageId, journal, seal);
+        } else {
```
nategraf, line 39: As I mentioned in the other comment, I think we should always verify the proof.

nategraf (commented):
Here is a partial review. I will come back again later.

There is a potential optimization to the lock action, which would be to "Merkle-ize" the request structure, pruning fields that are not required and instead just providing the hashes for reconstruction of the EIP712 digest for signature checking. This would reduce calldata and hashing costs. As we add more to the request, this becomes more attractive.

nategraf (commented):
Thinking about this a bit more, the fact that the requestor can control the contents of the callback, specifically the journal, feels somewhat risky to me. I can't think of an immediate attack against the current construction, but it feels somewhat close in nature to a attack I have seen against a contract that made user-specified calls and the attacker used this to call `transferFrom` and steal funds that had been approved to the contract.

It's very possible this is fine, but I do have two ideas for mitigations.
* Pass the callbacks through a proxy that is intended to be untrusted (i.e. no one should ever provide is an ERC20 approval or grant it a privileged role in any way). The market contract itself will only ever call the proxy.
* Remove the journal from the directly sent data and instead stash it in transient storage. A call from the called contract can get journal's digest or contents. This has a few challenges though, including storing arbitrary length byte strings in transient storage.

Not sure what the right level of caution is here, or the right mitigation if caution is warranted.

contracts/src/types/RequestLock.sol
willpote: Thats a great idea

contracts/src/IBoundlessMarketCallback.sol
willpote: I'm fine with us recommending an ultra-safe approach in our sample callback, but not sure I follow the reasoning here, lets chat offline

contracts/src/BoundlessMarket.sol
```diff
@@ -358,11 +366,29 @@ contract BoundlessMarket is
...
 
         // At this point the request has been fulfilled. The remaining logic determines whether
-        // payment should be sent and to whom.
+        // payment should be sent and to whom, and determines the callback to execute, if specified.
+        //
+        // In the case of a request that _was_ locked and the lock expired, the callback to be executed
+        // must be the one specified in the fulfillment context, not the one specified in the request lock.
+        // This is because it is possible for multiple requests to have the same id, and the callback
+        // stored in the request lock from an original request may not match the callback specified by
+        // the new request.
```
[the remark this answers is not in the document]
willpote: > It seems like we might want to check if the lock is expired before entering _fufillAndPayLocked.

That would require us to SLOAD the requestLock from storage when deciding to go down the locked or unlocked path, rather than using the request flag as we do today. I think at that point we may not need the request flag at all.

Not against this though, see: https://linear.app/risczero/issue/BM-546/consider-removing-lockedrequest-flag

willpote:
> Thinking about this a bit more, the fact that the requestor can control the contents of the callback, specifically the journal, feels somewhat risky to me. I can't think of an immediate attack against the current construction, but it feels somewhat close in nature to a attack I have seen against a contract that made user-specified calls and the attacker used this to call `transferFrom` and steal funds that had been approved to the contract.
> 
> It's very possible this is fine, but I do have two ideas for mitigations.
> 
> * Pass the callbacks through a proxy that is intended to be untrusted (i.e. no one should ever provide is an ERC20 approval or grant it a privileged role in any way). The market contract itself will only ever call the proxy.
> * Remove the journal from the directly sent data and instead stash it in transient storage. A call from the called contract can get journal's digest or contents. This has a few challenges though, including storing arbitrary length byte strings in transient storage.
> 
> Not sure what the right level of caution is here, or the right mitigation if caution is warranted.

Is this not a problem with the image that the callback implementer has chosen to use? If it is possible for the image to generate a journal that leads to tokens being transferred to the wrong person, then the image is the problem?

One thing this did make me think of is, would make sense to pass in the Predicate also used for the journal into the callback? Then even for more general purpose images, the callback can verify the journal meets some requirements? [I guess this would require us making the assessor output the predicate]

Re: the two mitigations, I don't fully see how they help. Even with a proxy, or fetching the journal from storage, ultimately there is a contract that needs to read the journal and decide what to do as a result of it- and if the journal can be  manipulated, that contract will do the wrong thing.

nategraf:
> > Thinking about this a bit more, the fact that the requestor can control the contents of the callback, specifically the journal, feels somewhat risky to me. I can't think of an immediate attack against the current construction, but it feels somewhat close in nature to a attack I have seen against a contract that made user-specified calls and the attacker used this to call `transferFrom` and steal funds that had been approved to the contract.
> > It's very possible this is fine, but I do have two ideas for mitigations.
> > 
> > * Pass the callbacks through a proxy that is intended to be untrusted (i.e. no one should ever provide is an ERC20 approval or grant it a privileged role in any way). The market contract itself will only ever call the proxy.
> > * Remove the journal from the directly sent data and instead stash it in transient storage. A call from the called contract can get journal's digest or contents. This has a few challenges though, including storing arbitrary length byte strings in transient storage.
> > 
> > Not sure what the right level of caution is here, or the right mitigation if caution is warranted.
> 
> Is this not a problem with the image that the callback implementer has chosen to use? If it is possible for the image to generate a journal that leads to tokens being transferred to the wrong person, then the image is the problem?
> 
> One thing this did make me think of is, would make sense to pass in the Predicate also used for the journal into the callback? Then even for more general purpose images, the callback can verify the journal meets some requirements? [I guess this would require us making the assessor output the predicate]
> 
> Re: the two mitigations, I don't fully see how they help. Even with a proxy, or fetching the journal from storage, ultimately there is a contract that needs to read the journal and decide what to do as a result of it- and if the journal can be manipulated, that contract will do the wrong thing.

My description of the potential issue is not super clear, but the issue I was thinking of would require an coincidental collision in the value of the 4 byte selector, and sender-based authentication, as well as meeting other constraints on how it interprets the received data. It is unlikely that this will ever be the case.

contracts/src/BoundlessMarketCallback.sol
```diff
@@ -0,0 +1,49 @@
...
+    /// @param imageId The image ID to accept proofs of.
+    constructor(IRiscZeroVerifier verifier, address boundlessMarket, bytes32 imageId) {
+        VERIFIER = verifier;
+        BOUNDLESS_MARKET = boundlessMarket;
+        IMAGE_ID = imageId;
+    }
+
+    /// @inheritdoc IBoundlessMarketCallback
+    function handleProof(bytes32 imageId, bytes calldata journal, bytes calldata seal) public {
+        require(msg.sender == BOUNDLESS_MARKET, "Invalid sender");
```
nategraf, line 35: I think this is fine to include, and possibly beneficial, but its not clear to me that this is case. Let's keep this check for now, and clearly folks can always override this function to remove it if they decide it's unnecessary.

Keep in mind that anyone can specify a callback to any contract, and anyone can create a request on the market. So the amount of constraint this applies to the call is limited, beyond the fact that the receipt is already checked, which is rechecked based on our discussion earlier.

crates/boundless-cli/src/bin/boundless-cli.rs
```diff
@@ -252,6 +253,12 @@ struct SubmitOfferRequirements {
     /// Journal prefix to use as the predicate in the requirements.
     #[clap(long)]
     journal_prefix: Option<String>,
+    /// Address of the callback to use in the requirements.
+    #[clap(long)]
+    callback_addr: Option<Address>,
+    /// Gas limit of the callback to use in the requirements.
+    #[clap(long)]
+    callback_gas_limit: Option<u64>,
```
nategraf, line 261: Adding `requires` options will prevent someone from specifying one, but forgetting the other. In the code below, that would result in no callback being added, which they may not figure out right away (bad UX).
```suggestion
    /// Address of the callback to use in the requirements.
    #[clap(long, requires = "callback_gas_limit")]
    callback_addr: Option<Address>,
    /// Gas limit of the callback to use in the requirements.
    #[clap(long, requires = "callback_addr")]
    callback_gas_limit: Option<u64>,
```

nategraf (approved):
Looks good to me!

I was staring at the logic in `fulfillBatch` again wondering if we really need 3 branches there. I think I have a nice simplification in that regard by moving the callback before fulfillment is set and using the fact that if anything fails, the whole call will revert anyway.

https://github.com/boundless-xyz/boundless/pull/364

PR #338, opened 2025-02-21
BM-568: run CI on release branches, enable zeth compile check

austinabell:


willpote (approved):

austinabell:
Disabling auto-merge for now, to figure out if the increased CI time is worth it for zeth order generator compile check

PR #346, opened 2025-02-26
BM-544: Support smart contract signatures

willpote:
Adds smart contract signature support

One funny thing, due to the way our "Client" object was created in our tests we were essentially sending requests from Smart Contracts already. This also meant that all the tests started failing once we had real SC support via EIP-1271. As a result I had to refactor Client, and took the opportunity to split things into different files while I was doing this. So lots of changes to tests.

One other note, to make this work for lockRequestWithSignature, it required adding a new field "proverAddress" to the interface. Without this it is not possible to support smart contract signatures for that method.


contracts/src/types/ProofRequest.sol
```diff
@@ -61,32 +62,36 @@ library ProofRequestLibrary {
         );
     }
 
-    /// @notice Verifies the client's signature over the proof request.
+    /// @notice Verifies a signature over a proof request.
+    /// @dev Supports EIP-1271 for smart contract signatures. If the signature is empty and the client is a smart
+    /// contract, calls isValidSignature with the abi encoded request as the signature.
```
nategraf, line 67: This way of deciding whether to use ECDSA verification vs smart contract signatures is restrictive of the client application logic. E.g. this cannot be used to allow a generic smart-contract wallet (e.g. a multisig) an ECDSA signature as part of the data sent in the `isValidSignature` request.

Another way of implementing this would be to use a flag somewhere in the request which indicates how it should be validated. I'd recommend using one of the bits that is currently reserved in the encoded request ID for this purpose. This works well because it allows separation of the request domain between those authorized with ECDSA and those authorized with EIP1271 and makes it evident which method to use given the request ID alone.

If we do it this way, we can forward the given `signature` bytes as is to allow the authorizing contract to define what it needs to be sent.

nategraf, line 67: When I first read this comment, I understood it to mean that EIP1271 is used if and only if `signature.len == 0 && addr.code.length > 0`. I then misread the code with this in mind. Now I understand this is simply stating that when the signature is empty, there is a special case to use the request as the "signature".

contracts/src/types/ProofRequest.sol
```diff
@@ -61,32 +62,36 @@ library ProofRequestLibrary {
...
+            }
+        } else if (signature.length == 0) {
+            // If the signature is empty and the client is a smart contract, we provide the request as the signature.
+            // This flow is intended to enable anyone to submit requests on behalf of a client smart contract.
+            // The client smart contract is expected to validate the request was constructed correctly.
+            if (IERC1271(addr).isValidSignature(structHash, abi.encode(request)) != IERC1271.isValidSignature.selector)
+            {
+                revert IBoundlessMarket.InvalidSignature();
+            }
+        } else {
```
nategraf, line 88: Do we have a good idea of the value of this case? I would be inclined to remove it and always forward the passed signature data. In the case where the client does use `abi.encode(request)` as its EIP1271 signature, this would result in larger calldata. So this is really just an optimization for that case.

This is likely not a major issue, but this also introduces a bit of a completeness hole in that it is impossible to use an empty string as a signature, which might be relevant if e.g. the authorizing contract works by saving a set of validated request hashes.

nategraf (commented):
I marked by old comment as outdated as I misread the actual implementation. I still think it would make sense to use a flag in the request ID to indicate which signature check is expected. Generally, using whether or not code is deployed to an address to determine how to check the signature feels too malleable to me (crypto-engineer spidey-sense is tingling). One concrete issue is that this construction prevents offline checking of whether a request is valid, even when signed with an EOA, since you need to always check the chain state for whether code has been deployed to an address. This also opens up the possibility for weird things when this changes, either by deployment of code to a previously empty address of the self-destruction of a contract. This raises questions about how to handle this in the Assessor.

contracts/src/types/ProofRequest.sol
willpote: Removed

willpote:
To summarize where we landed:

For locked requests:
* Smart contract signatures are validated at lock time only.
* EOA signatures are validated at lock time, and in the assessor (strictly this second check is not needed)

For unlocked requests:
* EOA signatures are validated off-chain in the assessor (because its cheaper)
* Smart contract signatures are validated on-chain at pricing time.

With this new design the changes at the contract level are very minimal, however the logic for that ties this all together is quite disjointed. For example, checking the fill's digest against the one stored in the lock or the FulfillmentContext is an important check that is nested in _fulfillAndPay. 

Refactoring this out in a gas efficient way is a very large refactor, which I'm leaving to a separate PR- mostly because I think we should also consider rethinking the "price" endpoints at the same time.

crates/boundless-market/src/contracts/artifacts/IBoundlessMarket.sol
```diff
@@ -246,6 +246,7 @@ interface IBoundlessMarket {
     function lockRequestWithSignature(
         ProofRequest calldata request,
         bytes calldata clientSignature,
+        address prover,
```
nategraf, line 249: Related to another comment, we don't need to add this if we don't support smart contract signatures for provers. The `lockRequest` method above supports prover contracts authorizing a lock.
```suggestion
```

contracts/src/types/ProofRequest.sol
```diff
@@ -64,32 +65,54 @@ library ProofRequestLibrary {
...
+    /// @param addr The address of the client.
+    /// @param signature The signature to validate.
+    /// @param isSmartContractSig Whether the signature should be validated as a smart contract signature.
+    function verifyClientSignature(
+        ProofRequest calldata,
+        bytes32 structHash,
+        address addr,
+        bytes calldata signature,
+        bool isSmartContractSig
+    ) internal view {
```
nategraf, line 103: From a cryptography API perspective, this is problematic. We shouldn't pass the hash into this function, as the hash is a core part of the signature check. If the hash is not calculated correctly, the signature is unsound. This is why before we passed the hash as a return arg to still allow reuse of the calculation. Additionally, passing in `isSmartContractSig` is less likely to be unsafe, but is kind of odd. Checking that flag should only take a small amount of gas, so even if it ends up repeated, that seems ok.

contracts/src/types/RequestLock.sol
```diff
@@ -65,24 +58,24 @@ library RequestLockLibrary {
...
 
     function setProverPaidAfterLockDeadline(RequestLock storage requestLock, address prover) internal {
         requestLock.prover = prover;
         requestLock.requestLockFlags = PROVER_PAID_AFTER_LOCK_FLAG;
-        // We don't zero out the second slot as stake information is required for slashing.
+        // We don't zero out slot 1 as stake information is required for slashing.
+        // Zero out slot 2 for gas refund.
+        // Clearing via requestLock.requestDigest = bytes32(0)
+        // would require touching slot 1 as well.
+        clearSlot2(requestLock);
```
nategraf, line 72: Since slashing will clear both slots anyway, why not leave this until slashing occurs?

> Clearing via requestLock.requestDigest = bytes32(0) would require touching slot 1 as well.

Is this true? I think the Solidity layout rules mean that, because `requestDigest` does not fit in slot1 entirely, it is aligned into slot2. Assuming I'm not mistaken`requestLock.requestDigest = bytes32(0)` should only touch slot2.

contracts/src/BoundlessMarket.sol
```diff
@@ -509,8 +516,9 @@ contract BoundlessMarket is
...
 
         // If no fulfillment context was stored for this request digest (via priceRequest),
-        // then payment cannot be processed. This check also serves as an expiration check since
-        // fulfillment contexts cannot be created for expired requests.
+        // then payment cannot be processed. This check also serves as
+        // 1/ an expiration check since fulfillment contexts cannot be created for expired requests.
+        // 2/ a smart contract signature check, since signatures are validated when a request is priced.
         FulfillmentContext memory context = FulfillmentContextLibrary.load(requestDigest);
         if (!context.valid) {
             return abi.encodeWithSelector(RequestIsNotPriced.selector, RequestId.unwrap(id));
```
nategraf, line 524: We need to revert here instead of simply returning a payment error. As written, someone could mark a request as fulfilled without ever passing a signature check.

contracts/src/types/RequestLock.sol
willpote: > Since slashing will clear both slots anyway, why not leave this until slashing occurs?

I guess this is a question of who should receive the gas refund for clearing slot 2 in the "lock => lock expire => fulfill after lock => slash" case: the slasher or the fulfiller?

There is also the lock => lock expire => no fulfill case, which clearing both slots on slash was meant to cover.

I'm inclined to favor the fulfiller over the slasher

willpote: > Is this true? I think the Solidity layout rules mean that, because requestDigest does not fit in slot1 entirely, it is aligned into slot2. Assuming I'm not mistakenrequestLock.requestDigest = bytes32(0) should only touch slot2.

Ah yes you are right. This approach is still cheaper, so I'm inclined to keep it, as its probably something we'd end up doing during mainnet gas golfing anyway

contracts/src/BoundlessMarket.sol
willpote: Just want to confirm the case you are describing so I can add a test for it.

Is it:
1. Request priced, where request id indicates ecdsa (so client signature check is skipped during price).
2. Fulfill called for a unlocked request id that indicates smart contract signature, but same request digest (signature check skipped in assessor)
3. Request would be marked as fulfilled and we'd attempt to pay

willpote: Got it, one interesting thing here is that we're in the wasLocked path, and for the request to have been locked there must have existed a valid SC signature at lock time. Regardless I think it makes sense to confirm the signature is valid again at fulfillment time, as we may need to pull additional funds from the client (in the multiple request same id case).

contracts/src/types/ProofRequest.sol
willpote: I think this should be addressed. Unfortunately I had to move the logic out of the ProofRequest library as I can't access the OZ hashing functions which are intended to enable upgradability of the domain (unclear if we'd ever use this, but I thought best to leave support for it for now).

crates/guest/assessor/assessor-guest/src/main.rs
```diff
@@ -53,8 +54,17 @@ fn main() {
     // - record the selector if it is present
     // We additionally collect the request and claim digests.
     for (index, fill) in input.fills.iter().enumerate() {
-        let request_digest =
-            fill.verify_signature(&eip_domain_separator).expect("signature does not verify");
+        // ECDSA signatures are always checked here.
+        // Smart contract signatures (via EIP-1271) are checked on-chain either when a request is locked,
+        // or when an unlocked request is priced and fulfilled.
+        let smart_contract_signed = RequestId::from_lossy(fill.request.id).smart_contract_signed;
```
nategraf, line 60: In this case, I think we should panic on unknown flags, rather than ignoring them. It's a balance between forwards-compatibility and preventing accidental issues. In the guess, I'd favor the stricter approach.
```suggestion
        let smart_contract_signed = RequestId::try_from(fill.request.id).unwrap().smart_contract_signed;
```

nategraf (approved):
Looks good! Thanks for iterating on this!

crates/guest/assessor/assessor-guest/src/main.rs
willpote: 212b74d11184a83b76211c20aa1a06616c2f7d27

PR #347, opened 2025-02-27
BM-577: Support skipping addresses in slasher

willpote:
We don't have tests for the slasher, but I have confirmed the image builds and runs with a list of addresses.

The changes to Cargo.lock and BoundlessMarket.hex seem to be files that were missed in some previous PR on main (they pop up on a fresh main branch also).

edit: may have been wrong on the hex change


crates/slasher/src/lib.rs
```diff
@@ -216,7 +218,38 @@ where
...
+            // nor if the lockRequest calls were for example, made via a proxy contract.
+            // This is a temporary solution to avoid slashing requests from the team's broker.
+            let tx_hash = log_data.transaction_hash.unwrap();
+            let tx = self
+                .boundless_market
+                .instance()
+                .provider()
+                .get_transaction_by_hash(tx_hash)
+                .await?
+                .unwrap();
```
austinabell, line 233: Feels like it might be more ideal to just skip the check instead of panicking if the provider does not provide the transaction hash with the log? Not sure if there are other conditions a provider might not include the tx hash with the log

Doesn't really matter since it's just a slashing bot, though

austinabell (approved):

PR #349, opened 2025-02-28
BM-581: order-stream: fix pending connections and add timeout

austinabell:
The issue was that previously on a failed upgrade, the pending connection was never removed or logged, so users could get in a state that rejected all future requests. 

This change adds a timeout to the pending connections such that if a request is made outside of the timeout window, it retries again, so that this doesn't get rejected permanently. This also adds a log on failure such that there is more information about failed upgrades.

### DIFF
--- a/crates/order-stream/src/lib.rs
+++ b/crates/order-stream/src/lib.rs
@@ -506,6 +506,10 @@ mod tests {
         let pending_connection = app_state.set_pending_connection(addr).await;
         assert!(pending_connection, "Should return true for a timed out connection");
 
+        // Newly set connection should result in pending_connection == false
+        let pending_connection = app_state.set_pending_connection(addr).await;
+        assert!(!pending_connection, "Should return false for a replaced connection within timeout");
+
         // Test removing a pending connection
         app_state.remove_pending_connection(&addr).await;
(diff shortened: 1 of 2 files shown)


crates/order-stream/src/lib.rs
```diff
@@ -415,4 +456,59 @@ mod tests {
...
+            let mut pending_connections = app_state.pending_connections.lock().await;
+            let old_time =
+                Instant::now() - (AppState::PENDING_CONNECTION_TIMEOUT + Duration::from_secs(1));
+            pending_connections.insert(addr, old_time);
+        }
+
+        // Now it should allow a new connection since the old one timed out
+        let pending_connection = app_state.set_pending_connection(addr).await;
+        assert!(pending_connection, "Should return true for a timed out connection");
+
```
nategraf, line 508: It might be good to have one more case here, just to make sure the above action changed the state in the expected way.
```suggestion

        // Newly set connection should result in pending_connection == false
        let pending_connection = app_state.set_pending_connection(addr).await;
        assert!(!pending_connection, "Should return false for a replaced connection within timeout");

```

nategraf (approved):

PR #357, opened 2025-03-03
BM-510: Stake rollover between requests with same id

willpote:
If two requests have the same id, and one is locked and not fulfilled. The fulfiller of the second request is eligible for a portion of the slashed stake. Note a request can only ever be locked once- without this property this would be a lot more complex to implement.

Some edge cases here:
1/ If the first request fully expires, then someone calls slash before the second request is fulfilled, then the second request fulfiller will not receive a portion of slashed stake
2/ If the second request is cheaper than the first request, the client may be eligible a refund (they were charged a high amount at lock time, then the fulfillment was ultimately cheaper)


contracts/src/BoundlessMarket.sol
```diff
@@ -360,6 +353,13 @@ contract BoundlessMarket is
         // At this point the request has been fulfilled. The remaining logic determines whether
         // payment should be sent and to whom.
         if (lock.lockDeadline >= block.number) {
+            if (lock.fingerprint != bytes8(requestDigest)) {
+                revert RequestLockFingerprintDoesNotMatch({
+                    requestId: id,
+                    provided: bytes8(requestDigest),
+                    locked: lock.fingerprint
+                });
+            }
```
nategraf, line 362: Why is this check moved? Here is some more context on why this check is required. Unless I am missing something, it's needed even when the deadline is not exceeded. https://github.com/boundless-xyz/boundless/pull/163

nategraf (commented):
I didn't get to a point of really understanding the changes here, and I need to go for the night. I'll come back to this tomorrow.

contracts/src/BoundlessMarket.sol
willpote: I moved this because if a lock expires but the request as a whole has not expired, you were unable to fulfill a different request with the same ID. The (absurdly named) test case "testFulfillWasLockedRequestRepeatIndexStakeRolloverFirstRequestNotExpired" would fail

I think in the PR you linked since we used to fully zero out the request lock on fulfillment, we didn't have enough information to slash the original locker. Now that we keep information in the request lock, I think we are able to hold them accountable even if they fulfill a different request. 

We have a couple tests that cover something similar to the situation you described: "testFulfillWasLockedRequestByOriginalLocker" and "testSlashWasLockedRequestFulfilledByLocker" and "testFulfillWasLockedRequestRepeatIndexStakeRolloverFirstRequestNotExpired" but I'll add something that matches exactly the case you described

willpote: Test here: [a491928](https://github.com/boundless-xyz/boundless/pull/357/commits/a4919282ab7f725590cadbdf1df0b330399be278)

nategraf: I've looked at the logic again, in the most up to date commit, and it looks good to me now. I may have also misread this before. I'll look at those tests as well. https://github.com/boundless-xyz/boundless/blob/38adb9e1e2ef912b2de428dafbcfd502ee32452c/contracts/src/BoundlessMarket.sol#L419-L425

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1469,6 +1474,277 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
     }
 
+    // One request is locked, fully expires.
+    // A second request with the same id is then fulfilled.
+    // Slash should award stake to the fulfiller of the second request.
+    function testFulfillWasLockedRequestRepeatIndexStakeRollover() public {
+        Client client = getClient(1);
+
+        // Create two distinct requests with the same ID. It should be the case that only one can be
+        // filled, and if one is locked, the other cannot be filled.
```
nategraf, line 1484: This comment is not true anymore.
```suggestion
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1469,6 +1474,277 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+    }
+
+    // One request is locked, the lock expires, but the request is not yet expired.
+    // A second request with the same id is then fulfilled.
+    // Slash should award stake to the fulfiller of the second request.
+    function testFulfillWasLockedRequestRepeatIndexStakeRolloverFirstRequestNotExpired() public {
+        Client client = getClient(1);
+
+        // Create two distinct requests with the same ID. It should be the case that only one can be
+        // filled, and if one is locked, the other cannot be filled.
```
nategraf, line 1548:
```suggestion
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1469,6 +1474,277 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+            timeout: uint32(100),
+            lockStake: 1 ether
+        });
+        Offer memory offerB = Offer({
+            minPrice: 2 ether,
+            maxPrice: 2 ether,
+            biddingStart: uint64(block.number),
+            rampUpPeriod: uint32(0),
+            lockTimeout: uint32(block.number) + offerA.timeout + 101,
+            timeout: uint32(block.number) + offerA.timeout + 101,
```
nategraf, line 1564:
```suggestion
            lockTimeout: offerA.timeout + 101,
            timeout: offerA.timeout + 101,
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1469,6 +1474,277 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+            createFillAndSubmitRoot(requestB, APP_JOURNAL, address(fulfiller));
+
+        vm.prank(address(fulfiller));
+        boundlessMarket.priceAndFulfill(requestB, clientSignatureB, fill, assessorReceipt);
+
+        // Check that the request ID is marked as fulfilled.
+        expectRequestFulfilled(fill.id);
+
+        // Slash should revert as the original locked request has not yet fully expired.
+        vm.prank(address(locker));
```
nategraf, line 1595:
```suggestion
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1469,6 +1474,277 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+            abi.encodeWithSelector(
+                IBoundlessMarket.RequestIsNotExpired.selector, fill.id, uint64(block.number) + uint64(offerA.timeout)
+            )
+        );
+        boundlessMarket.slash(fill.id);
+
+        // Advance to where the original locked request has fully expired.
+        vm.roll(uint64(block.number) + uint64(offerA.timeout) + 1);
+
+        vm.prank(address(locker));
```
nategraf, line 1606:
```suggestion
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1469,6 +1474,277 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+    }
+
+    // One request is locked and the client is charged 2 ether. The request expires unfulfilled.
+    // A second request with the same id is then fulfilled for a cost of just 1 ether.
+    // The client should be refunded the difference.
+    function testFulfillWasLockedRequestRepeatIndexSecondRequestCheaper() public {
+        Client client = getClient(1);
+
+        // Create two distinct requests with the same ID. It should be the case that only one can be
+        // filled, and if one is locked, the other cannot be filled.
```
nategraf, line 1624:
```suggestion
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1469,6 +1474,277 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+            timeout: uint32(100),
+            lockStake: 1 ether
+        });
+        Offer memory offerB = Offer({
+            minPrice: 1 ether,
+            maxPrice: 1 ether,
+            biddingStart: uint64(block.number),
+            rampUpPeriod: uint32(0),
+            lockTimeout: uint32(100),
+            timeout: uint32(block.number) + offerA.timeout + 101,
```
nategraf, line 1640:
```suggestion
            timeout: offerA.timeout + 100,
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1469,6 +1474,277 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+            timeout: uint32(100),
+            lockStake: 1 ether
+        });
+        Offer memory offerB = Offer({
+            minPrice: 3 ether,
+            maxPrice: 3 ether,
+            biddingStart: uint64(block.number) + uint64(offerA.timeout) + 1,
+            rampUpPeriod: uint32(10),
+            lockTimeout: uint32(100),
+            timeout: uint32(block.number) + offerA.timeout + 101,
```
nategraf, line 1705:
```suggestion
            timeout: 100,
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1469,6 +1474,277 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+        boundlessMarket.lockRequest(requestA, clientSignatureA);
+
+        client.expectBalanceChange(-2 ether);
+
+        vm.roll(uint64(block.number) + uint64(offerA.lockTimeout) + 1);
+
+        // Attempt to fill request B, which costs just 1 ether at the time of fulfillment.
+        (Fulfillment memory fill, AssessorReceipt memory assessorReceipt) =
+            createFillAndSubmitRoot(requestB, APP_JOURNAL, address(fulfiller));
+        vm.prank(address(fulfiller));
```
nategraf, line 1666: Fulfillment transactions don't need to come from the prover address.
```suggestion
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1469,6 +1474,277 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+        // Lock-in request A.
+        vm.prank(address(locker));
+        boundlessMarket.lockRequest(requestA, clientSignatureA);
+
+        vm.roll(uint64(block.number) + uint64(offerA.lockTimeout) + 1);
+        // Attempt to fill request B.
+        (Fulfillment memory fill, AssessorReceipt memory assessorReceipt) =
+            createFillAndSubmitRoot(requestB, APP_JOURNAL, address(fulfiller));
+
+        vm.prank(address(fulfiller));
```
nategraf, line 1588:
```suggestion
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1469,6 +1474,277 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+        // Lock-in request A.
+        vm.prank(address(locker));
+        boundlessMarket.lockRequest(requestA, clientSignatureA);
+
+        vm.roll(uint64(block.number) + uint64(offerA.timeout) + 1);
+        // Attempt to fill request B.
+        (Fulfillment memory fill, AssessorReceipt memory assessorReceipt) =
+            createFillAndSubmitRoot(requestB, APP_JOURNAL, address(fulfiller));
+
+        vm.prank(address(fulfiller));
```
nategraf, line 1524:
```suggestion
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1857,6 +2133,61 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+            timeout: uint32(100),
+            lockStake: 1 ether
+        });
+        Offer memory offerB = Offer({
+            minPrice: 3 ether,
+            maxPrice: 3 ether,
+            biddingStart: uint64(block.number) + uint64(offerA.timeout) + 1,
+            rampUpPeriod: uint32(10),
+            lockTimeout: uint32(100),
+            timeout: uint32(block.number) + offerA.timeout + 101,
```
nategraf, line 2157:
```suggestion
            timeout: 100,
```

nategraf (approved):
Took me a bit to fully understand the different execution paths, but this looks good to me now.

willpote:
I had to bring back OZ's SafeCast as solmate's implementation doesn't handle int256 conversions, which this change requires.

PR #359, opened 2025-03-04
BM-572: Remove freezing - BREAKING

capossele:
This PR drops the frozen functionality and it is a breaking change:

```
contracts/src/BoundlessMarket.sol:52: Upgraded `accounts` to an incompatible type
        - In mapping(address => struct Account)
          - Bad upgrade to struct Account
        - In struct Account
          - Deleted `flags`
            > Keep the variable even if unused
          - Upgraded `requestFlagsInitial` to an incompatible type
            - Bad upgrade from uint56 to uint64
 ```

### CODE REVIEW

File: crates/boundless-market/src/contracts/artifacts/Account.sol [hunk 1]
```diff
@@ -16,8 +16,7 @@ struct Account {
     uint96 balance;
     /// @dev Balance of staked tokens.
     uint96 stakeBalance;
-    /// @dev flags is a bitfield for account-wide flags. Currently only the least significant
-    /// bit is used to indicate that the account is frozen.
+    /// @dev flags is a bitfield for account-wide flags. Currently unused.
```
willpote, line 19: Hmm shouldn't we roll it back into `requestFlagsInitial`?

willpote: Interesting, I can't think of another use case for it, and once the system is truly permissionless you could always get around a flag by creating a new account. So I'm in favor of removing

capossele: dropped with https://github.com/boundless-xyz/boundless/pull/359/commits/892d7dd7c22db67e78efdc2bf6b2e04067767014

willpote (approved):

PR #360, opened 2025-03-04
BM-592: Tests for when client withdraws funds before fulfill

willpote:
This was an issue caught by Veridise's static analysis tool that would result in us not returning an error when the client does not have funds at fulfillment time.


capossele (approved):

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1325,6 +1325,41 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+        vm.prank(address(client));
+        boundlessMarket.withdraw(balance);
+
+        // Advance the chain ahead to simulate the lock timeout.
+        vm.roll(uint64(block.number) + request.offer.lockTimeout + 1);
+
+        (Fulfillment memory fill, bytes memory assessorSeal) =
+            createFillAndSubmitRoot(request, APP_JOURNAL, address(testProver));
+
+        vm.expectRevert(abi.encodeWithSelector(IBoundlessMarket.InsufficientBalance.selector, client));
```
nategraf, line 1358: Interesting! It's somewhat odd that this is what happens, since there are funds associated with the request, escrowed for the lock, that could be reasonably used to make the payment. It could make sense that the implementation would not support this, but it is also not clear this is the behavior we want to be ensuring is the case.

nategraf: Oh, now I think I understand what's happening! Since the lock is expired, and time has passed, the price in transient storage is used instead of the price on the lock. The price in transient storage is higher as long as the lock was not already the max price, since the auction effectively continues while the lock lasts.

```suggestion
        // NOTE: While the lock does have enough funds escrowed to cover the original price, the auction further
        // raised the price in the time that passed between the lock transaction and this fulfill transaction. As a
        // result, the escrowed funds are no longer sufficient.
        vm.expectRevert(abi.encodeWithSelector(IBoundlessMarket.InsufficientBalance.selector, client));
```

willpote: Just to clarify- in this test there aren't enough funds in escrow to cover the full price (since the price has increased since lock time)

I think the case you are describing is where the lock expires, and the price is flat so we could re-use the locks funds to pay- which is a case we do actually support (though not sure we have a test covering it, I can add one)

nategraf (approved):

contracts/test/BoundlessMarket.t.sol
willpote: Looks like we were writing at the same time! But yes you are right

PR #362, opened 2025-03-04
BM-557: Enable to find orders submitted off chain in boundless

capossele:
- Adds `request_digest` field to Order; 
- Adds unique constraint for request_digest on the order_stream DB;
- Adds fetch_order as a new API entrypoint

Closes BM-435


crates/boundless-market/src/client.rs
```diff
@@ -470,6 +470,38 @@ where
...
+                    .context(
+                        "Order stream client not available. Please provide an order stream URL",
+                    )?
+                    .fetch_order(request_id)
+                    .await?;
+                if orders.is_empty() {
+                    return Err(ClientError::Error(anyhow!("Order not found")));
+                }
+                // TODO: Handle multiple orders
+                Ok(orders[0].clone())
```
capossele, line 501: Not sure what's best here: there might be the chance that multiple orders share the same request_id.
Ideally one would identify a request with its digest and use that to request the order. However we do not currently have enough tooling for make this easy.

willpote: Hmm for now having an optional `requestDigest` in the fulfill command probably makes the most sense. Then if  there are multiple orders, and they haven't provided `requestDigest` we error out?

willpote: Might also be worth discussing if we really need the multiple orders single request id property. Its definitely interesting, but its not simple or intuitive for users, and if its passing on this type of complexity to users I'm not 100% sure its worth it.

(it also makes the contracts quite a bit more complex)

capossele: Ok IO've refactored a bit with [fc8fde7](https://github.com/boundless-xyz/boundless/pull/362/commits/fc8fde7ce3c260a7b9bcdf33d851198707947c44)

crates/boundless-market/src/client.rs
```diff
@@ -470,6 +471,40 @@ where
...
+                    request,
+                    request_digest: digest,
+                    signature: PrimitiveSignature::try_from(signature_bytes.as_ref())
+                        .map_err(|_| ClientError::Error(anyhow!("Failed to parse signature")))?,
+                })
+            }
+            Err(_) => Ok(self
+                .offchain_client
+                .as_ref()
+                .context("Order stream client not available. Please provide an order stream URL")?
```
willpote, line 503:
```suggestion
                .context("Request not found on-chain and order stream client not available. Please provide an order stream URL")?
```

willpote (approved):

PR #364, opened 2025-03-05
Review of #335: simplify fulfillAndPay by extracting callback

nategraf:
- **simplify fulfillAndPay function by extracting callbacks**
- **update gas snapshots**


nategraf:
On my way home I realized two things:
1. Using reentrance, it is possible to violate the "exactly once" property of the callback under this implementation. The target would either need to be crafted to do so, or itself call out to a contract crafted to do so.
2. The "exactly once" calling is not a meaningful guarantee, because I can issue an arbitrary number (N) of identical callbacks by creating N requests with the same image ID and requirements and fulfilling each with the same journal and seal. The called contract would find this indistinguishable from the "exactly once" property being broken.

My inclination is to not attempt to provide this property, and instead simplify the code by removing this check (implemented here by checking the fulfillment status of the request).

willpote (approved):
Nice!

PR #366, opened 2025-03-05
BM-580: Drop request and signature from RequestSubmitted event

capossele:
With this change we can save about 10k of gas when submitting requests onchain, at the price of an additional RPC call to retrieve the request and signature from the transaction calldata.

It does not remove all the requestId from the events, nor add an additional event to the priceRequest as suggested in the BM-580 issue. The first can be useful for filtering, avoiding downloading unnecessary transactions and the second would add more gas cost to the contracts.


Wollac (approved):

willpote (approved):
Wow, bigger savings that expected.

PR #367, opened 2025-03-05
Bump risc0-ethereum to v1.4 and alloy to 0.11

capossele:


crates/boundless-market/src/contracts/mod.rs
```diff
@@ -699,236 +699,112 @@ pub fn eip712_domain(addr: Address, chain_id: u64) -> EIP721DomainSaltless {
...
-    use std::sync::Arc;
-
-    use crate::contracts::{
-        boundless_market::BoundlessMarketService,
-        hit_points::{default_allowance, HitPointsService},
-    };
 
     // Bytecode for the contracts is copied from the contract build output by the build script. It
     // is checked into git so that we can avoid issues with publishing to crates.io. We do not use
     // the full JSON build out because it is less stable.
```
nategraf, line 721: Comment is referring to the `.hex` files that are now removed.
```suggestion
```

nategraf (commented):
I'm having trouble understanding how we are able to drop the `.hex` files from `git`. We added those originally because we needed to make sure `boundless-market` (with the `test-utils` flag) could build for `crates.io`. What is changed here that allows us to avoid this issue? Is it the case that we now compile the contracts from the source included in the `artifacts` directory at build time? It doesn't seem like that could be the case since we need dependencies from submodules (e.g. OZ) to build the contracts.

capossele:
> I'm having trouble understanding how we are able to drop the `.hex` files from `git`. We added those originally because we needed to make sure `boundless-market` (with the `test-utils` flag) could build for `crates.io`. What is changed here that allows us to avoid this issue? Is it the case that we now compile the contracts from the source included in the `artifacts` directory at build time? It doesn't seem like that could be the case since we need dependencies from submodules (e.g. OZ) to build the contracts.

Instead of using the hex files we now include the bytecode as a string directly in the rust file.
https://github.com/boundless-xyz/boundless/blob/87e0437d8fe6c9ad8093c77ae2dd8f0a682166f1/crates/boundless-market/src/contracts/mod.rs#L722

capossele:
> > I'm having trouble understanding how we are able to drop the `.hex` files from `git`. We added those originally because we needed to make sure `boundless-market` (with the `test-utils` flag) could build for `crates.io`. What is changed here that allows us to avoid this issue? Is it the case that we now compile the contracts from the source included in the `artifacts` directory at build time? It doesn't seem like that could be the case since we need dependencies from submodules (e.g. OZ) to build the contracts.
> 
> Instead of using the hex files we now include the bytecode as a string directly in the rust file.
> 
> https://github.com/boundless-xyz/boundless/blob/87e0437d8fe6c9ad8093c77ae2dd8f0a682166f1/crates/boundless-market/src/contracts/mod.rs#L722

Now that I think about though, I'm not so sure that would work, you might be right

capossele:
@nategraf thanks for catching that, I think I've found a solution that works. We generate a file `bytecode.rs` only if the solidity compiled contracts are available, which is not the case when running cargo publish

crates/boundless-market/src/contracts/artifacts/bytecode.inc
Wollac, line 1: What is this file used for?
I couldn't find any reference to it?

crates/boundless-market/build.rs
```diff
@@ -222,11 +236,44 @@ fn copy_artifacts() {
...
-            file.write_all(bytecode.as_bytes()).unwrap();
+            // Append the contract definition with embedded bytecode
+            rust_content.push_str(&format!(
+                r#"alloy::sol! {{
+    #[sol(rpc, bytecode = "{}")]
+    contract {} {{
+        {}
+    }}
+}}"#,
+                bytecode,
```
Wollac, line 247: What if I have a slightly different solc version installed on my machine? Couldn't this lead to different bytecode and trigger a file change?

crates/boundless-market/build.rs
```diff
@@ -193,17 +193,31 @@ fn copy_interfaces_and_types() {
...
+
+    // Start with file header content
+    let mut rust_content = String::from("// Auto-generated file, do not edit manually\n\n");
+
     for contract in ARTIFACT_TARGET_CONTRACTS {
         let source_path = src_path.join(format!("{contract}.sol/{contract}.json"));
+        if !source_path.exists() {
+            println!("cargo:warning=Skipping contract bytecode generation");
+            return;
+        }
```
Wollac, line 220: When a single file doesn't exist, essentially nothing gets written no default is used?

crates/boundless-market/build.rs
```diff
@@ -222,11 +236,44 @@ fn copy_artifacts() {
...
+        "RiscZeroSetVerifier" => {
+            "constructor(address verifier, bytes32 imageId, string memory imageUrl) {}"
+        }
+        "BoundlessMarket" => {
+            r#"constructor(address verifier, bytes32 assessorId, address stakeTokenContract) {}
+            function initialize(address initialOwner, string calldata imageUrl) {}"#
         }
+        "ERC1967Proxy" => "constructor(address implementation, bytes memory data) payable {}",
+        "HitPoints" => "constructor(address initialOwner) payable {}",
+        _ => "",
```
Wollac, line 276: These need to match the actual constructors and must be kept consistent. As an alternative to the entire bytecode.rs file we could as this point also copy the entire compiled json file (and keep it checked in). Then we would have interface and bytecode and a default.

crates/boundless-market/build.rs
```diff
@@ -222,11 +236,44 @@ fn copy_artifacts() {
                 .unwrap()
                 .trim_start_matches("0x");
 
-            // Write to new file with .hex extension
-            let dest_file = dest_path.join(format!("{contract}.hex"));
```
Wollac, line 226: When https://github.com/alloy-rs/core/issues/659 gets resolved we could keep the .hex approach

crates/boundless-market/src/contracts/artifacts/bytecode.inc
capossele: Forgot to delete it

crates/boundless-market/build.rs
capossele: That is true, we do have some constraints in the forge.toml though to decrease this likelihood

crates/boundless-market/build.rs
capossele: Yeah although that I just a trick for cargo publish, as otherwise the one checked on GitHub will be used

crates/boundless-market/build.rs
capossele: Unfortunately checking in the json does not work as they are too susceptible to changes during different compilations/environments

crates/boundless-market/src/contracts/artifacts/bytecode.inc
capossele: Deleted by https://github.com/boundless-xyz/boundless/pull/367/commits/b3ece52000900214f38bf0f17e95e8188588dc4a

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -786,240 +786,115 @@ pub fn eip712_domain(addr: Address, chain_id: u64) -> EIP721DomainSaltless {
     }
 }
 
+#[cfg(feature = "test-utils")]
+#[allow(missing_docs)]
+pub mod bytecode;
```
nategraf, line 791: Why is this module public, versus just using it within the test utils module?

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -1030,86 +905,78 @@ pub mod test_utils {
...
-            Ok((verifier, set_verifier, hit_points, boundless_market))
-        }
+        // Mine forward some blocks using the provider
+        deployer_provider.anvil_mine(Some(10), Some(2)).await.unwrap();
+        deployer_provider.anvil_set_interval_mining(2).await.unwrap();
+
+        Ok((verifier, set_verifier, hit_points, boundless_market))
+    }
+
+    pub struct DefaultTestCtx;
```
nategraf, line 941: This struct appears to exist just to group together `DefaultTestCtx::create` and `DefaultTestCtx::create_with_rpc_url`, and IIUC these functions are not on TextCtx because of the way generics would force the type to be specified. I'd recommend that the these functions just be at the module level instead, so `create_test_ctx` and `create_text_ctx_with_rpc_url` instead. It's a bit awkward, but these isn't really an advantage to adding this struct here.

crates/boundless-market/src/contracts/hit_points.rs
```diff
@@ -50,7 +46,7 @@ where
     }
 
     /// Returns the underlying IHitPointsInstance.
-    pub fn instance(&self) -> &IHitPointsInstance<T, P, Ethereum> {
+    pub fn instance(&self) -> &IHitPointsInstance<(), P, Ethereum> {
```
nategraf, line 49: I was so confused by this until I finally found alloy's private-but-not-really traits. Very weird, but I think I see they dilemma they are trying to work around 🤷 
https://github.com/alloy-rs/alloy/blob/main/crates/contract/src/lib.rs#L39-L48

nategraf (approved):

crates/boundless-market/build.rs
capossele: Maybe it is not worth for now, as when https://github.com/alloy-rs/core/pull/898 gets into a release we could revert to using .hex files

crates/boundless-market/build.rs
capossele: yeah we might want to revert using hex files once that gets merged into a release

PR #368, opened 2025-03-06
BM-596: Drop journal and seal from ProofDelivered event

capossele:
Similarly to #366 this PR drops the journal and seal from he emitted ProofDelivered event. This saves about 4K of gas in test, but might be a bigger saving considering that in real deployment both journal and seal are bigger


crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -1274,16 +1287,54 @@ where
...
+    }
+    if let Ok(call) = IBoundlessMarket::priceAndFulfillAndWithdrawCall::abi_decode(data, true) {
+        return Ok(vec![call.fill]);
+    }
+    if let Ok(call) = IBoundlessMarket::priceAndFulfillBatchAndWithdrawCall::abi_decode(data, true)
+    {
+        return Ok(call.fills);
+    }
+
+    Err(anyhow!("Failed to decode calldata as any fulfillment call"))
```
nategraf, line 1325: We should include the hex-encoded selector (i.e. first 4 bytes of calldata) or the full calldata here to help with debugging if this error occurs. It would definitely be easy to add a new fulfill entrypoint without remembering to update this list.

nategraf (approved):

PR #370, opened 2025-03-07
BM-462: Drop requirePayment

capossele:
Here we drop the requirePayment bool. Valid fulfillments should never revert, but only return an error message (or an array of errors if batching). 
As for simulating whether a prover will get paid or not would be sufficient to just call the transaction and check if returns any error.


contracts/src/BoundlessMarket.sol
```diff
@@ -432,11 +432,7 @@ contract BoundlessMarket is
         }
 
         if (paymentError.length > 0) {
-            if (fill.requirePayment) {
-                revertWith(paymentError);
-            } else {
-                emit PaymentRequirementsFailed(paymentError);
-            }
+            revertWith(paymentError);
```
nategraf, line 435: Sorry, this was not explained well in the issue. We actually want to always apply `requirePayment = false` rather than `requirePayment = true` 🙈
```suggestion
            emit PaymentRequirementsFailed(paymentError);
```

contracts/src/BoundlessMarket.sol
```diff
@@ -473,7 +469,7 @@ contract BoundlessMarket is
         // payment should be sent and to whom.
         // While the request is locked, only the locker is eligible for payment.
         if (lock.prover != assessorProver) {
-            return abi.encodeWithSelector(RequestIsLocked.selector, RequestId.unwrap(id));
+            return "";
```
nategraf, line 472: Related to the comment above, we want to leave this line as is.

contracts/src/BoundlessMarket.sol
capossele: When you simulate a call you only have read permissions so any event would not be available. As such I dropped the event and instead returned an error

willpote: My bad (I wrote the issue) Sorry this wasn't clear!

nategraf (approved):

PR #375, opened 2025-03-07
BM-416: Use timestamps instead of block numbers for request information

nategraf:
Change the auction timing (e.g. `offer.biddingStart` and `offer.timeout`) to use timestamps instead of block numbers. This should simplify the experience for developers and make it more consistent across different blockchains.

As a drive-by fix, this fixes the logic for determining how much gas and stake is assigned to pending orders.


crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -1009,19 +1020,19 @@ where
...
         let min_price = U256::from(offer.minPrice);
 
-        if block_numb < offer.biddingStart {
-            return Err(MarketError::Error(anyhow!("Block number before bidding start")));
+        if timestamp <= offer.biddingStart {
+            return Ok(offer.minPrice);
         }
 
-        if block_numb < offer.biddingStart + offer.rampUpPeriod as u64 {
+        if timestamp < offer.biddingStart + offer.rampUpPeriod as u64 {
```
willpote, line 1032: Noting we'll need to update this in #378

crates/broker/src/order_monitor.rs
```diff
@@ -163,14 +174,17 @@ where
         // TODO: spawn a side thread to avoid missing new blocks while this is running:
         let order_count = if let Some(last_monitor_block) = opt_last_block {
             let current_block = self.chain_monitor.current_block_number().await?;
+            let current_block_timestamp = self.chain_monitor.current_block_timestamp().await?;
 
+            // TODO: It looks like this might now be true.
```
willpote, line 179: Is this just a note that this is a fix? Can remove

willpote (approved):
Overall this looks great. 

My main question is around using RPC provider timestamps vs system timestamps (see comments)

Also noting that the order-generators will need to be updated; issues there may not have been caught due to BM-555

crates/boundless-market/src/client.rs
```diff
@@ -237,7 +238,7 @@ where
             storage_provider: None,
             offchain_client: None,
             local_signer: None,
-            bidding_start_offset: BIDDING_START_OFFSET,
+            bidding_start_delay: BIDDING_START_OFFSET,
```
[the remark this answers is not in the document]
nategraf: https://github.com/boundless-xyz/boundless/pull/375/files/9f0bb4aedbcd368350b6f34a5edfec8fa3fd2d76..9d4faf9b63d1840132e23a9a4bb3d8f5619ea83e

crates/boundless-cli/src/bin/boundless-cli.rs
```diff
@@ -850,6 +836,11 @@ async fn fetch_file(url: &Url) -> Result<Vec<u8>> {
     Ok(data)
 }
 
+// TODO(#379): Avoid drift relative to the chain's timestamps.
+fn now_timestamp() -> u64 {
+    SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap().as_secs()
+}
```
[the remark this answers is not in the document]
nategraf: This is true! I had some issues in tests, and I know client clocks can be quite different, so I do want to make some kind of hybrid here that estimates the difference between the chain clock and the system clock to use as an offset. I'm sure we are not the first ones to do this, so there must be reasonable approaches already out there.

crates/broker/src/order_monitor.rs
nategraf: Ah, I forgot to come back to that. I now understand what it's doing here, and I've adjusted the message to be more accurate. I think at some point the logic must have changed here. There still may be value in actually trying to scan the chain for locked orders that are _not_ in the DB, as a way of recovering from error.

c23799f0685619bf10757ea02de9a23e95498472

PR #377, opened 2025-03-07
BM-547: Use SafeCastLib and ERC20 from solmate

capossele:
Partially related to [BM-547 (reduce-bytecode-increase-optimizer)](https://linear.app/risczero/issue/BM-547/reduce-bytecode-increase-optimizer) this PR switches with using the solmate alternatives


willpote (approved):
Decent savings on depositing stake, otherwise not quite as big savings as hoped on bytecode and gas in other areas :(

Future thing worth considering is our own cast library that removes all the functions we dont use (not sure if the compiler does that automatically or not

capossele:
It seems that the compiler already takes care of that, at least when it comes to libraries

PR #378, opened 2025-03-07
BM-606: Set price to zero after lock deadline

nategraf:
In recent work, we split the request timeout into two parts: a lock timeout and a timeout. In particular, the lock timeout is the moment where any prover that has locked the request is considered to have failed their obligation, loses exclusivity, and is slashable.

On `main` right now, the price of a request continues to increase during the locked period, and so if the lock expires it is likely that the price has increased to `maxPrice` (and more generally, it is no less than the original lockin price). Because of exclusivity, if a prover submits their bid, there is no opportunity for other provers to indicate their bid until the lock has expired, and the price may have risen past where they would have bid. This impacts price discovery, and can cause the client to end up paying more than the "true" market price defined as the minimum price that a prover capable of filling the request would accept. Note that this "true" price is not directly observable.

Adding to this, since (unburned portion of) the stake is added to the reward to any prover that fulfills after the lock deadline, in the case that the locked prover fails, there is some notion that if the stake is high enough, the stake alone is sufficient to motivate a prover to fill the request. It is likely that this is the case for a large class of requests.

After much discussion, we've decided to address the issue above by setting the price paid by the client for a request delivered after the lock deadline and before the final request deadline to zero. As a reward, the prover will still obtain any unburned stake taken from a prover that failed the request, if another prover previously locked it.

As with most market mechanisms, it is impossible to tell whether this is the correct choice until we have data from a live market with funds flowing through it. When we have that data, we may revisit this.


crates/boundless-market/src/contracts/artifacts/Offer.sol
```diff
@@ -20,13 +20,20 @@ struct Offer {
...
+    /// each block until it reaches `maxPrice`.
     uint32 rampUpPeriod;
     /// @notice Timeout for the lock, expressed as seconds from bidding start.
-    /// @dev This is the deadline for the lock to expire.
+    /// @dev Once locked, if a valid proof is not submitted before this deadline, the prover can
+    /// be "slashed", which refunds the price to the requester and takes the prover stake.
+    /// Additionally, the fee paid by the client is zero for proofs delivered after this time.
+    /// Note that after this time, and before `timeout` a proof can still be delivered to fulfill
+    /// the request.
     uint32 lockTimeout;
```
willpote, line 32: I think we should also note that if a never locked request is fulfilled before this deadline then they are eligible for payment, if after then not

willpote: paymentTimeout could be a better name? Though that may miss the nuance that you can still get stake if fulfilling a previously locked request.. naming is hard

nategraf: I've clarified it a bit further in 0e83b9d8fc0c05928c53eab57ac034a9fe9c8a66

crates/boundless-market/src/contracts/artifacts/Offer.sol
```diff
@@ -20,13 +20,20 @@ struct Offer {
...
-    /// @notice Timeout for delivering the proof, expressed as seconds from bidding start.
-    /// @dev Once locked-in, if a valid proof is not submitted before this deadline, the prover can be "slashed," which refunds the price to the requester.
+    /// @notice Timeout for the request, expressed as seconds from bidding start.
+    /// @dev After this time the request is considered completely expired and can no longer be
+    /// fulfilled. After this time, the `slash` action can be completed to finalize the transaction
```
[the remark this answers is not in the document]
nategraf: I've tried to make it more explicit in the comment on lock timeout, about no fee being transferred. It's hard to communicate it exactly right because its technical possible to deliver a proof still to have the request marked as fulfilled, and if you are either us (RISC Zero) or an application with strong interests in filling requests related to your protocol, there might be a reason to do so.

willpote:
One thought, when we validate the offer should be check that  that rampUpPeriod is <= lockTimeout? With the new design I don't see a reason to ever ramp up beyond that point

nategraf:
This PR is finally ready for another look and to merge

contracts/src/BoundlessMarket.sol
```diff
@@ -525,6 +528,14 @@ contract BoundlessMarket is
         }
         uint96 price = context.price;
 
+        // Deduct any additionally owned funds from client account. The client was already charged
+        // for the price at lock time once when the request was locked. We only need to charge any
+        // additional price for the difference between the price of the fulfilled request, at the
+        // current block, and the price of the locked request.
+        //
+        // Note that although they have they same ID, the locked request and the fulfilled request
```
willpote, line 536:
```suggestion
        // Note that although they have the same ID, the locked request and the fulfilled request
```

contracts/src/types/Offer.sol
```diff
@@ -20,13 +20,22 @@ struct Offer {
...
+    /// Note that after this time, and before `timeout` a proof can still be delivered to fulfill
+    /// the request. This applies both to locked and unlocked requests; if a proof is delivered
+    /// after this timeout, no fee will be paid from the client.
     uint32 lockTimeout;
-    /// @notice Timeout for delivering the proof, expressed as seconds from bidding start.
-    /// @dev Once locked-in, if a valid proof is not submitted before this deadline, the prover can be "slashed," which refunds the price to the requester.
+    /// @notice Timeout for the request, expressed as seconds from bidding start.
+    /// @dev After this time the request is considered completely expired and can no longer be
+    /// fulfilled. After this time, the `slash` action can be completed to finalize the transaction
+    /// if it was not fulfilled.
```
willpote, line 38:
```suggestion
    /// if it was locked but not fulfilled.
```

willpote (approved):

PR #383, opened 2025-03-09
BM-607: Implement order stream client ping

austinabell:
Adds a configurable ping time to order stream client. Avoiding creating a separate public API for this to avoid breaking changes to include polling time.

Refactors the order stream integration test to handle multiple websocket connections and disconnection and receiving orders through that websocket.

Also simplifies the ws handler to not spawn and manage tasks. I assume this might have been done to allow cancellable tasks, but this seems unnecessary.

### DIFF
--- a/crates/order-stream/src/ws.rs
+++ b/crates/order-stream/src/ws.rs
@@ -206,139 +206,141 @@ async fn broadcast_order(db_order: &DbOrder, state: Arc<AppState>) {
 }
 
 async fn websocket_connection(socket: WebSocket, address: Address, state: Arc<AppState>) {
-    let parent_state = state.clone();
-    parent_state.ws_tasks.spawn(async move {
-        let (mut sender_ws, mut recver_ws) = socket.split();
+    let (mut sender_ws, mut recver_ws) = socket.split();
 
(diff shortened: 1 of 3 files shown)


crates/order-stream/src/ws.rs
```diff
@@ -206,139 +206,141 @@ async fn broadcast_order(db_order: &DbOrder, state: Arc<AppState>) {
...
-                            tracing::warn!("Error receiving message from {address}: {err:?}");
-                            break;
-                        }
-                        None => {
-                            tracing::debug!("Empty recv from {address}, closing connections");
+                    }
+                    Some(Ok(Message::Close(_))) => {
+                        tracing::warn!("Client {address} sent close message, closing conn");
+                        break;
+                        // TODO: cleaner management of Some(Ok(Message::Close))
```
austinabell, line 307: @mothran What did you have in mind for this? Just printing out the status code and message, or something else?

mothran: Looks like nothing specific, just graceful shutdown if the client wants to close the connection. Not sure explicitly what I was thinking at the time.

nategraf (approved):

PR #385, opened 2025-03-10
BM-536: Add broker support for unaggregated proofs

capossele:
This PR enables the broker and the local prover to fulfill requests that specify a given selector, aka unaggregated proofs. It also revamp the test_utils mod so that both a mock and prod test deployment can be used in tests.


capossele:
Putting back to draft as when using the local prover we get:
```
Error: BonsaiProver does not support compression on existing receiptsSet receipt_kind on ProverOpts in initial prove request to get a Groth16Receipt.
```

contracts/e2e-test/E2E.sol
```diff
@@ -0,0 +1,300 @@
...
+import {BoundlessMarketLib} from "../src/libraries/BoundlessMarketLib.sol";
+import {HitPoints} from "../src/HitPoints.sol";
+import {IHitPoints} from "../src/HitPoints.sol";
+
+Vm constant VM = Vm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
+bytes32 constant APP_IMAGE_ID = 0x722705a82a1dab8369b17e16bac42c9c538057fc1d32933d21ea2b47f292efb4;
+uint256 constant DEFAULT_BALANCE = 1000 ether;
+
+/// Test designed to checks that a full e2e flow works.
+contract E2E is RiscZeroCheats, Test {
```
willpote, line 42: Cool that we can test things e2e in Solidity!

What I'm unclear on is when we should write e2e tests here vs in E2E tests in Boundless Market. Should we always do both? Or are there particular types of features we should cover here and not in the market

willpote (commented):
Only got 1/2 way through and only nits so far, will finish up review tomorrow

crates/boundless-market/src/contracts/test_utils.rs
```diff
@@ -0,0 +1,316 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Interface to interact with the `HitPoints` contract used for staking on the market during the Boundless testnet.
+//!
+//! NOTE: This module will be removed in later phases of the testnet and before mainnet.
```
nategraf, line 17: This description is from a different file

crates/boundless-market/src/contracts/test_utils.rs
```diff
@@ -0,0 +1,316 @@
...
+        <[u8; 32]>::from(set_builder_id).into(),
+        String::default(),
+    )
+    .await
+    .context("failed to deploy RiscZeroSetVerifier")?;
+    Ok(*instance.address())
+}
+
+pub async fn deploy_hit_points<P: Provider>(
+    deployer_signer: &PrivateKeySigner,
```
nategraf, line 103: Since you only need the address here, I would pass that in instead of the signer.
```suggestion
    owner_address: Address
```

crates/boundless-market/src/contracts/test_utils.rs
```diff
@@ -0,0 +1,316 @@
...
+) -> Result<Address> {
+    let deployer_address = deployer_signer.address();
+    let instance = HitPoints::deploy(deployer_provider, deployer_address)
+        .await
+        .context("failed to deploy HitPoints contract")?;
+    Ok(*instance.address())
+}
+
+pub async fn deploy_boundless_market<P: Provider>(
+    deployer_signer: &PrivateKeySigner,
```
nategraf, line 114: Same as above
```suggestion
    owner_address: Address,
```

crates/broker/src/order_picker.rs
```diff
@@ -90,13 +91,20 @@ where
...
-        if order.request.requirements.selector != FixedBytes::<4>([0; 4]) {
-            tracing::warn!("Removing order {order_id:x} because it has a selector requirement");
-            self.db.skip_order(order_id).await.context("Order has a selector requirement")?;
-            return Ok(());
-        }
+        match Selector::from_bytes(order.request.requirements.selector.into())
+            .context("Failed to parse selector")?
+        {
+            // Supported selectors
+            Selector::FakeReceipt | Selector::SetVerifierV0_2 | Selector::Groth16V1_2 => {}
```
nategraf, line 98: `Selector::FakeReceipt` is all zeroes, which is a bit confusing in this context being it will interpret requests with no particular selector as a request for a "fake receipt". Splitting off into its own case would be more clear.
```suggestion
            // Default: any selector is valid
            [0u8; 4] => {}
            Selector::SetVerifierV0_2 | Selector::Groth16V1_2 => {}
```
(I don't think that suggestion will actually compile)

crates/broker/src/order_picker.rs
```diff
@@ -90,13 +91,20 @@ where
...
         }
 
-        // TODO(#BM-536): Filter based on supported selectors
-        // Drop orders that specify a selector
-        if order.request.requirements.selector != FixedBytes::<4>([0; 4]) {
-            tracing::warn!("Removing order {order_id:x} because it has a selector requirement");
-            self.db.skip_order(order_id).await.context("Order has a selector requirement")?;
-            return Ok(());
-        }
+        match Selector::from_bytes(order.request.requirements.selector.into())
```
nategraf, line 94:
```suggestion
        // TODO(BM-40): When accounting for gas costs of orders, a groth16 selector has much higher cost.
        match Selector::from_bytes(order.request.requirements.selector.into())
```

crates/broker/src/aggregator.rs
nategraf, line 1: I don't see anything here that prevents the receipts orders that are compressed independently from being added to the set builder batch. Does that happen?

crates/broker/src/submitter.rs
```diff
@@ -167,31 +182,56 @@ where
...
-                    order_claim,
-                    order_path,
-                    inclusion_params.digest(),
-                );
-                let seal =
-                    set_inclusion_receipt.abi_encode_seal().context("Failed to encode seal")?;
+                let seal = match Selector::from_bytes(order_request.requirements.selector.into())
+                    .context("Failed to parse selector")?
+                {
+                    Selector::FakeReceipt | Selector::SetVerifierV0_2 => {
```
nategraf, line 188: Here as well, it's confusing the use "FakeReceipt" to mean "any selector". Would it be possible to represent the selector field with an `Option` in Rust, for instance?

contracts/e2e-test/E2E.sol
nategraf: Since we have the Rust e2e tests that include the Router, Groth16, etc, I'm personally inclined to drop this one. Foundry FFI is fragile and requires pre-installing the CLI (see GH actions config) and I think the Rust version of this test is likely to run faster / be a better foundation for adding more variations.

willpote (commented):
Finished review and no other comments, though also curious about Victors comment here: https://github.com/boundless-xyz/boundless/pull/385/files#r1996435818

crates/broker/src/aggregator.rs
willpote: Hmm, I suppose currently they are needed to recompute the root in the assessor, so I guess the proposal would be to also modify the assessor to skip fills with selectors in them when recomputing the root?

crates/broker/src/order_picker.rs
capossele: I've created a `UNSPECIFIED_SELECTOR` constant

crates/broker/src/aggregator.rs
capossele: that's right, no changes to that have been made. If we want to include that, as mentioned by @willpote we need to adjust the assessor as well as the contract side of things. Might be better to add to a separate PR maybe?

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -786,235 +786,17 @@ pub fn eip712_domain(addr: Address, chain_id: u64) -> EIP721DomainSaltless {
     }
 }
 
+/// Constant to specify when no selector is specified.
+pub const UNSPECIFIED_SELECTOR: FixedBytes<4> = FixedBytes::<4>([0xFF; 4]);
```
capossele, line 790: Now when no specific selector is specified, the value is 0xFFFFFFFF rather than all 0s. That is so that we can keep compatibility with the mock verifier and have more tools for our testing

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -523,6 +523,18 @@ impl Requirements {
...
+    ///
+    /// This will set the selector to the appropriate value based on the current environment.
+    /// In dev mode, the selector will be set to `FakeReceipt`, otherwise it will be set
+    /// to `Groth16V1_2`.
+    pub fn with_unaggregated_proof(self) -> Self {
+        match is_dev_mode() {
+            true => Self { selector: FixedBytes::from(Selector::FakeReceipt as u32), ..self },
+            false => Self { selector: FixedBytes::from(Selector::Groth16V1_2 as u32), ..self },
+        }
+    }
```
capossele: I've added this utility so that, since we can only support one zkvm version at the moment, it's easier for the requestor to use this rather than looking for the specific selector

capossele: It's also easier in tests as it automatically select the Fake receipt if in dev mode.

crates/boundless-market/src/contracts/test_utils.rs
```diff
@@ -0,0 +1,297 @@
...
+) -> Result<Address> {
+    let instance =
+        RiscZeroGroth16Verifier::deploy(deployer_provider, control_root, bn254_control_id)
+            .await
+            .context("failed to deploy RiscZeroGroth16Verifier")?;
+    Ok(*instance.address())
+}
+
+pub async fn deploy_mock_verifier<P: Provider>(deployer_provider: P) -> Result<Address> {
+    let instance = RiscZeroMockVerifier::deploy(deployer_provider, FixedBytes::ZERO)
```
willpote, line 77: It looks like when we create the MockVerifier you can specify which selector it expects

What would the changes look like to keep 0s to indicate "no selector specified", and instead change our mock verifier and dev mode proofs to use 1s instead? Reason I ask is that in Solidity 0s is much more natural for 'not specified' than 1s. Also if anyone ever constructs requests without our CLI/SDK, it feels like a footgun.

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -523,6 +523,19 @@ impl Requirements {
...
+    /// This will set the selector to the appropriate value based on the current environment.
+    /// In dev mode, the selector will be set to `FakeReceipt`, otherwise it will be set
+    /// to `Groth16V1_2`.
+    #[cfg(not(target_os = "zkvm"))]
+    pub fn with_unaggregated_proof(self) -> Self {
+        match risc0_zkvm::is_dev_mode() {
+            true => Self { selector: FixedBytes::from(Selector::FakeReceipt as u32), ..self },
+            false => Self { selector: FixedBytes::from(Selector::Groth16V1_2 as u32), ..self },
+        }
+    }
```
willpote, line 538: Related to my other comment, but since we can initialize a mock verifier that expects a specific selector, is this necessary? It feels not ideal that we have to introduce the concept of dev mode into the client

crates/boundless-market/src/contracts/test_utils.rs
capossele: Yeah it's not perfect, I know. The other way around though is also not idea. So If we were to specify a selector for the mock verifier, then decode seal would need to change as it assumes 0s for the mock verifier. If instead we would change the mock verifier to default to 1s, then that might work but we would need to update that behaviour on the risc0-ethereum and potentially its consumers (foundry-template maybe?).

willpote: Yeah thats true. I would lean towards updating it, but don't have a great sense of the impact it would have to other packages/consumers. Curious what @nategraf thinks

nategraf: Using a non-zero value makes sense to me. That would have been a good choice to start with. A challenge with that change is that it requires updating `risc0-ethereum` and possibly `risc0` since the selector is, by default, derived from the first 4 bytes of the verifier parameters. We could define a `MockVerifierParameters` and has it like we do for others to get a default selector. Using a chosen value like `0xFFFFFFFF` certainly works as well, with similar requirements to update it.

The intention for the mocker verifier is that it should be possible to chose a selector when deploying it, and when generating or encoding fake receipts. One goal is that we might have multiple mock verifiers behind a router, each with a different selector. So I'd say in this PR we can keep the default as zeroes, but override it (we may need to add some code to make that possible). It would be good to follow-up with a change to the default, but that can be a separate PR.

crates/broker/src/tests/e2e.rs
```diff
@@ -37,72 +120,152 @@ async fn simple_e2e() {
...
+    ctx.customer_market
+        .wait_for_request_fulfillment(
+            U256::from(request.id),
+            Duration::from_secs(1),
+            request.expires_at(),
+        )
+        .await
+        .unwrap();
+
+    // Check for a broker panic
```
willpote, line 212: It would be nice to check that the emitted seal uses the selector

willpote (approved):

PR #387, opened 2025-03-10
BM-612: Only restrict application verification gas if no selector required

willpote:
Since we have added a way for requestors to request "raw" proofs rather than merkle inclusion proofs, we need to remove the restriction on gas for verifying those raw proofs when requested.

I don't see an easy way to do this without introducing another array in fulfill batch, at least with the current structure. We may want to reconsider this when we get to mainnet gas golfing


nategraf (approved):
Looks good 👍

PR #393, opened 2025-03-12
BM-615 - Add caching for assessor and set builder images

### LINKED ISSUE #388: Cache guest binaries downloaded from IPFS
Right now, the broker downloads the application guest image on every job, and additionally downloads the set builder and assessor guests when starting the process. This is a source of extra bandwidth costs and time cost for request evaluation, and (more importantly) has been a source of downtime when IPFS gateway rate limits are hit.

One concrete step to improve on this is to cache the content downloaded from IPFS URLs. Because IPFS is content addressable, it is possible to use the CID int the URL as the cache key. Caching the set builder and the assessor images would be obvious and immediately helpful. Caching application guest images would also be quite helpful, but requires a heuristic for cache invalidation and a cap on the total cache size.

willemolding:
Closes #388 

Uses reqwest middleware to add a HTTP-aware cache. See https://docs.rs/http-cache-reqwest/latest/http_cache_reqwest/index.html. This is used only for caching requests for assessor and set-builder images at this time.

- Adds caching to requests to retrieve assessor and set builder images
- - Can trivially also add it to user images but this may need more discussion
- Add new command line flag `--cache-images` to broker
- Adds new command line arg `--cache-dir` to broker (default `/tmp/broker_cache/`)


crates/broker/src/lib.rs
```diff
@@ -118,6 +118,17 @@ pub struct Args {
     /// From the `RetryBackoffLayer` of Alloy
     #[clap(long, default_value_t = 100)]
     pub rpc_retry_cu: u64,
+
+    /// Flag indicating if images should be cached locally
+    ///
+    /// If set, images will be downloaded and stored for reuse.
+    /// Currently this only caches the assessor and set-builder images
+    #[arg(long, action = ArgAction::SetTrue)]
+    pub cache_images: bool,
```
austinabell, line 127: This seems unused? Perhaps should the cache_dir variable just be renamed to remove this? Or planned to use this flag in the future?

willemolding: Partly for enabling caching of different things (e.g images, inputs) but also to allow for a default cache dir location for when no `--cache-dir` is passed.

austinabell (approved):

willpote (commented):
Any reason we wouldn't default this to true? Otherwise it still feels like a footgun that provers will eventually run into.

willemolding:
> Any reason we wouldn't default this to true? Otherwise it still feels like a footgun that provers will eventually run into.

Ok yeah we can default to true. In this case maybe a `--nocache` flag is a nicer way to do it

willpote (approved):

PR #395, opened 2025-03-12
BM-625: docs: rename indexer with explorer

nahoc:
We now favor "Explorer" over "Indexer" in all comms


austinabell (approved):
Seems the risc0-ethereum submodule modified accidentally, other than that good!

PR #396, opened 2025-03-12
BM-626: Fix bento crash with zero keccak input request

austinabell:
The error would panic and crash the instance, which was indefinitely irrecoverable. This rejects that invalid case, such that the node will continue.

The part I am unsure about is the sanity check within the executor for keccak requests. This changes to error log and skip those requests, and didn't seem like a clean mechanism to fail to proof as a whole, but perhaps this should still go through to fail with the prove agent so that it errors and then perhaps the tasks get cancelled through taskdb? Haven't looked that in depth at the implications yet.

### DIFF
--- a/bento/crates/workflow/src/tasks/executor.rs
+++ b/bento/crates/workflow/src/tasks/executor.rs
@@ -219,6 +219,12 @@ impl Coprocessor {
 
 impl CoprocessorCallback for Coprocessor {
     fn prove_keccak(&mut self, request: ProveKeccakRequest) -> Result<()> {
+        if request.input.is_empty() {
+            anyhow::bail!(
+                "Received empty keccak input with claim_digest: {}",
+                request.claim_digest
+            );
+        }
         self.tx.send_blocking(request)?;
         Ok(())
     }
@@ -408,13 +414,6 @@ pub async fn executor(agent: &Agent, job_id: &Uuid, request: &ExecutorReq) -> Re
 
     writer_tasks.spawn(async move {
         while let Ok(mut keccak_req) = keccak_rx.recv().await {
-            if keccak_req.input.is_empty() {
-                tracing::error!(
-                    "Received empty keccak input with claim_digest: {}, skipping",
-                    keccak_req.claim_digest
-                );
-                continue;
-            }
             let keccak_count_tmp = keccak_count.load(Ordering::Relaxed);
             let redis_key = format!("{coproc_prefix}:{}", keccak_req.claim_digest);
             redis::set_key_with_expiry(
--- a/bento/crates/workflow/src/tasks/executor.rs
+++ b/bento/crates/workflow/src/tasks/executor.rs
@@ -408,6 +408,13 @@ pub async fn executor(agent: &Agent, job_id: &Uuid, request: &ExecutorReq) -> Re
 
     writer_tasks.spawn(async move {
         while let Ok(mut keccak_req) = keccak_rx.recv().await {
+            if keccak_req.input.is_empty() {
+                tracing::error!(
+                    "Received empty keccak input with claim_digest: {}, skipping",
+                    keccak_req.claim_digest
+                );
+                continue;
+            }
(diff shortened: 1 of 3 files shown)


bento/crates/workflow/src/tasks/executor.rs
```diff
@@ -408,6 +408,13 @@ pub async fn executor(agent: &Agent, job_id: &Uuid, request: &ExecutorReq) -> Re
 
     writer_tasks.spawn(async move {
         while let Ok(mut keccak_req) = keccak_rx.recv().await {
+            if keccak_req.input.is_empty() {
```
zeroecco, line 411: can you walk me through this? if keccak input is empty the whole job should fail. executor, and the consumers should never see work that is empty

austinabell: Is there a mechanism for me to cancel within this call? Is the suggestion just to remove this check and have it fail in the prover agent?

I assume this is never hit, but a request got put in for this that would consistently panic on startup, so any restarts would hit the same case.

austinabell: Or something like this do you mean:

```diff
diff --git a/bento/crates/workflow/src/tasks/executor.rs b/bento/crates/workflow/src/tasks/executor.rs
index db190d8..6249f3e 100644
--- a/bento/crates/workflow/src/tasks/executor.rs
+++ b/bento/crates/workflow/src/tasks/executor.rs
@@ -409,11 +409,15 @@ pub async fn executor(agent: &Agent, job_id: &Uuid, request: &ExecutorReq) -> Re
     writer_tasks.spawn(async move {
         while let Ok(mut keccak_req) = keccak_rx.recv().await {
             if keccak_req.input.is_empty() {
-                tracing::error!(
-                    "Received empty keccak input with claim_digest: {}, skipping",
-                    keccak_req.claim_digest
-                );
-                continue;
+                taskdb::update_task_failed(
+                    &coproc_pool,
+                    &job_id_ref,
+                    &format!("keccak_{}", keccak_count.load(Ordering::Relaxed)),
+                    "Received empty keccak input",
+                )
+                .await
+                .expect("Failed to update task failed");
+                break;
             }
             let keccak_count_tmp = keccak_count.load(Ordering::Relaxed);
             let redis_key = format!("{coproc_prefix}:{}", keccak_req.claim_digest);

```

mothran: Yeah something like that makes more sense to to prevent the executor from emitting something incorrect. You might be right about a redis level failure, but I would generally advise do what ever you can to bubble up task level errors to the agent's main (instead of inline update_task_failed) because then all error paths go through the same flow: https://github.com/boundless-xyz/boundless/blob/main/bento/crates/workflow/src/lib.rs#L240-L251

Because here we explicitly handle retries.

Where to fail really depends on what is the origin problem, if the executor is emitting a bad keccak, fail and bubble up the failure from within the exec task vs pushing a bad input along to keccak prove. If something happens between the exec and keccak-prove then I would recommend sanity checking inside the keccak task and failing it there.

austinabell: https://github.com/boundless-xyz/boundless/pull/396/commits/744cd2f0b275ca3bcc04f87896728307d695eaab seems to do what you are suggesting, as it fails within the exec task. To confirm, this is what you meant or did I misinterpret?

bento/crates/workflow/src/tasks/keccak.rs
```diff
@@ -40,6 +40,13 @@ pub async fn keccak(agent: &Agent, job_id: &Uuid, request: &KeccakReq) -> Result
         input: try_keccak_bytes_to_input(&keccak_input)?,
     };
 
+    if keccak_req.input.is_empty() {
+        anyhow::bail!(
+            "Received empty keccak input with claim_digest: {}, skipping",
```
austinabell, line 45:
```suggestion
            "Received empty keccak input with claim_digest: {}",
```

mothran (approved):

PR #397, opened 2025-03-12
BM-613: Use postgres by default in the broker

willpote:
Currently we use sqlite, but we've consistently been seeing issues of the form:
```
| 2025-03-10 22:06:09.373 | 0: SQL error |
| 2025-03-10 22:06:09.373 | 1: error returned from database: (code: 5) database is locked |
```
A change was made to reduce max connections to 1 to combat this. This likely helped, but the issue has actually still been occurring- just with not enough frequency to trigger alarms until earlier this week. 

We are unable to reproduce the issue which makes this hard to fix, even with custom built db fuzz testing. Additionally Sqlite is not recommended for use cases that perform concurrent writes as the entire database is locked on every write. Since we have multiple tasks trying to update rows for different orders and different tables, it might just be a case of us using the wrong DB for the job.

This PR changes the default DB to postgres. Sqlite support is left for ease of running tests.

Currently draft as the docker compose setup needs to be updated, and needs manual testing


nategraf:
Having two different DBs, and especially one we use in testing and one we use in prod, feels like a systematic risk for introducing bugs. One of the reasons we liked sqlite is that it is embedable within the process. We don't actually need a lot of fancy features from the DB, so we could consider non-SQL options as well (e.g. RocksDB or one or the many similar projects). I'm not super up to date with the options.

PR #403, opened 2025-03-14
BM-632: docs: remake homepage

nahoc:


nategraf (approved):
Dark mode does not seem to adjust the text color properly.

PR #404, opened 2025-03-14
BM-633: Add Keccak to Planner

rlukata:
Partially handles BM-616


bento/crates/workflow/src/tasks/executor.rs
```diff
@@ -326,18 +341,18 @@ pub async fn executor(agent: &Agent, job_id: &Uuid, request: &ExecutorReq) -> Re
     // set the segment prefix
     let segments_prefix = format!("{job_prefix}:{SEGMENTS_PATH}");
 
-    let mut writer_tasks = JoinSet::new();
     // queue segments into a spmc queue
-    let (segment_tx, segment_rx) = async_channel::bounded::<Segment>(CONCURRENT_SEGMENTS);
-    let (task_tx, task_rx) = async_channel::bounded(TASK_QUEUE_SIZE);
-    let (keccak_tx, keccak_rx) = async_channel::bounded(TASK_QUEUE_SIZE);
+    let (segment_tx, mut segment_rx) = tokio::sync::mpsc::channel::<Segment>(CONCURRENT_SEGMENTS);
```
zeroecco, line 345: correct me if I am wrong but couldn't this lead to OOM without a max queue depth?

rlukata: The tokio::sync::mpsc::channel implementation is bounded by default and requires a max queue depth set by (in this case) `CONCURRENT_SEGMENTS` and right below set by `TASK_QUEUE_SIZE`.

rlukata: Additionally, to call an unbounded channel we would use [this function](https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.unbounded_channel.html).

zeroecco (approved):

PR #408, opened 2025-03-17
chore: fixed Effective Hz calculation error

barajeel:
I noticed that if `output->'total_cycles'` is missing, `NULL` breaks the final Hz calculation. Added `COALESCE` to default to `0`, so everything calculates correctly even when the data isn’t there.


nategraf:
@rlukata or @zeroecco, does this make sense to you. Under what cases would `total_cycles` be `NULL`?

rlukata:
> @rlukata or @zeroecco, does this make sense to you. Under what cases would `total_cycles` be `NULL`?

Although `total_cycles` _can_ be NULL, the whole program would break when mapping to the rust type [ExecutorResp](https://github.com/boundless-xyz/boundless/blob/5a8104bad12fc7497be936006d56f9746a58f9c6/bento/crates/workflow-common/src/lib.rs#L72). I don't think this change helps us in a meaningful way but I appreciate the contribution.

PR #413, opened 2025-03-17
BM-598: Limit concurrent locked orders

austinabell:
Solves part of BM-598 in that we need multiple ways to limit the capacity of orders taken on by the broker, and limiting the amount of current locks is not a perfect solution but a reasonable mechanism that would be useful in the short term or to limit risk.

Will open PR for review once more thoroughly tested and will try to make sure no edge cases and minimimal overhead. Basically how this works is querying the database for orders that have been selected to be locked but not fulfilled, and then spawning tasks to handle pricing each. The capacity is calculated on an interval, every 10s, where the config and db will be checked, which will pick up when orders are fulfilled and more capacity available. Whenever pricing successful, it will reduce from the capacity to avoid extra tasks from being marked as locked.

May also refactor the design to do this, as this doesn't handle the case where the order is planned to be locked in the future very well. The issue with having the concurrent locks be monitored in the order monitor (which actually calls lock) is that it isn't clear when the broker needs to lock the order in by to make sure that it will complete it in time, because that is done in pricing.

### DIFF
--- a/crates/broker/src/order_picker.rs
+++ b/crates/broker/src/order_picker.rs
@@ -73,7 +73,7 @@ where
         Self { db, config, prover, provider, market }
     }
 
-    async fn price_order(&self, order_id: U256, order: &Order) -> Result<(), PriceOrderErr> {
+    async fn price_order(&self, order_id: U256, order: &Order) -> Result<bool, PriceOrderErr> {
         tracing::debug!("Processing order {order_id:x}: {order:?}");
 
         let (min_deadline, allowed_addresses_opt) = {
@@ -87,7 +87,7 @@ where
             if !allow_addresses.contains(&client_addr) {
                 tracing::warn!("Removing order {order_id:x} from {client_addr} because it is not in allowed addrs");
                 self.db.skip_order(order_id).await.context("Order not in allowed addr list")?;
-                return Ok(());
+                return Ok(false);
             }
         }
 
@@ -96,7 +96,7 @@ where
         if order.request.requirements.selector != FixedBytes::<4>([0; 4]) {
             tracing::warn!("Removing order {order_id:x} because it has a selector requirement");
             self.db.skip_order(order_id).await.context("Order has a selector requirement")?;
-            return Ok(());
+            return Ok(false);
         }
 
         // is the order expired already?
@@ -108,7 +108,7 @@ where
         if expiration <= now {
             tracing::warn!("Removing order {order_id:x} because it has expired");
             self.db.skip_order(order_id).await.context("Failed to delete expired order")?;
-            return Ok(());
+            return Ok(false);
         };
 
         // Does the order expire within the min deadline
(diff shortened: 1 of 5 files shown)


crates/broker/src/order_picker.rs
```diff
@@ -462,47 +503,93 @@ where
...
+
+                    _ = pricing_check_timer.tick() => {
+                        // Queue up orders that can be added to capacity.
+                        let order_size = if let Some(capacity) = capacity {
+                            capacity.saturating_sub(
+                                u32::try_from(pricing_tasks.len()).expect("tasks u32 overflow"),
+                            )
+                        } else {
+                            // If no maximum lock capacity, request a max of 10 orders at a time.
+                            10
```
nategraf, line 549: If capacity is much greater than 10, do we want to this only price 10 orders at a time?

crates/broker/src/order_picker.rs
```diff
@@ -462,47 +503,93 @@ where
...
+                        };
+
+                        picker_copy
+                            .spawn_pricing_tasks(&mut pricing_tasks, order_size)
+                            .await
+                            .map_err(SupervisorErr::Recover)?;
```
nategraf, line 555: If `order_size` is zero, this seems to avoid kicking off any pricing tasks, but it does still query the DB.

crates/broker/src/order_picker.rs
```diff
@@ -454,6 +455,46 @@ where
...
+        tasks: &mut JoinSet<Result<Option<U256>, (U256, PriceOrderErr)>>,
+        capacity: u32,
+    ) -> Result<()> {
+        if capacity == 0 {
+            return Ok(());
+        }
+
+        let order_res = self.db.update_orders_for_pricing(capacity).await?;
+
+        for (order_id, order) in order_res {
```
nategraf, line 485: It took me a minute to understand mechanically how this PR restricts the number of locks. It appears that is essentially done here since this loop will run 0 times if capacity is zero.

crates/broker/src/order_picker.rs
austinabell: There is the `capacity == 0` check in that method. I can move it to wrap this call, if preferred?

crates/broker/src/order_picker.rs
austinabell: yeah, tried to make it simpler, but couldn't think of ways to reduce more. Very open to suggestions, not ecstatic about this current design.

For this, yes, the high-level pattern is checking how many orders can be scheduled, and then spawning that many tasks. In this case though, if capacity is zero the code block isn't hit, but how it restricts further spawning is that the capacity is based on the `JoinSet` of tasks, and unless the pricing fails will count against the capacity, reducing the tasks spawned.

crates/broker/src/order_picker.rs
austinabell: Good point!

nategraf (approved):

Issue #410, opened 2025-03-17
Retries for the Bonsai prover


willpote:
We see 5xx like failures for some calls to Bonsai. 5XX failures are safe to retry, so we should do that to improve reliability.

```
2025-03-16T12:31:04.446-04:00
2025-03-16T16:31:04.443042Z ERROR broker::proving: FATAL: Order failed to prove: Monitoring proof failed
158
2025-03-16T12:31:04.446-04:00
Caused by:
159
2025-03-16T12:31:04.446-04:00
Prover failure: SessionId { uuid: "d9ffe6d9-936d-4e14-b534-afb96274c09e" } failed: INTERNAL_ERROR
```
and
```
| 2025-03-16 06:02:16.653 | 2025-03-16T06:02:16.652074Z ERROR broker::order_picker: Pricing order failed: 2546c553d857d20658ece248f7c7d0861a24068193fac809 Bonsai proving error |
| 2025-03-16 06:02:16.653 | Caused by: |
| 2025-03-16 06:02:16.653 | server error `{"message": "Internal server error"}` |
| 2025-03-16 06:02:16.653 | Stack backtrace: |
| 2025-03-16 06:02:16.653 | 0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from |
| 2025-03-16 06:02:16.653 | at ./usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.95/src/backtrace.rs:27:14 |
| 2025-03-16 06:02:16.653 | 1: <T as core::convert::Into<U>>::into |
| 2025-03-16 06:02:16.653 | at ./rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/core/src/convert/mod.rs:759:9 |
```

PR #415, opened 2025-03-18
BM-657: Add retries for calls to Bonsai

willpote:
We've seen transient errors when calling the Bonsai API. This adds retries to all API calls (previously we only retried calls to check job status).  This also adds retries to the entire proving workflow, to better capture the case where are able to kick off a proving job and it terminates, but it terminates with a retryable internal server error.

This should help with: https://github.com/boundless-xyz/boundless/issues/410

### DIFF
--- a/crates/broker/src/provers.rs
+++ b/crates/broker/src/provers.rs
@@ -11,6 +11,7 @@ use std::{
 use async_trait::async_trait;
 use bonsai_sdk::{
     non_blocking::{Client as BonsaiClient, SessionId, SnarkId},
+    responses::SnarkStatusRes,
     SdkErr,
 };
 use boundless_market::input::InputBuilder;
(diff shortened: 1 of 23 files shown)


rlukata (approved):
this looks great

willpote:
> this looks great

Thanks for the review!

I ended up adding retries to the entire proving workflow as well. This better captures the case we hit the other day where we successfully kick off a job, but the job ends up failing due to some transient AWS reason on the Bonsai side.

crates/broker/src/provers.rs
```diff
@@ -188,16 +209,72 @@ impl StatusPoller {
...
     }
+
+    async fn poll_with_retries_snark_id(
+        &self,
+        proof_id: &SnarkId,
+        client: &BonsaiClient,
+    ) -> Result<String, ProverError> {
+        loop {
+            let mut status = None;
+            for retry_count in 0..self.retry_counts {
```
rlukata, line 220: why not use `retry::` here?

willpote:
Had to rebase and force push to get signed commits working.

Issue #351, opened 2025-02-28
Seeing frequent `RequestLockFingerprintDoesNotMatch` errors


derrix060:
Hello all,

I'm using git commit `87fbcf7fa2dbe23908d219ea17189268b27544d7` (latest commit at the time of writing for `release-0.6.0`).

I'm running a prover [0x52b9d1fe1d6d1dbd2a3d39367beb86c3a201a4c1](https://indexer.beboundless.xyz/provers/0x52b9d1fe1d6d1dbd2a3d39367beb86c3a201a4c1) and it's getting frequently slashed because it can't submit the order due to the `RequestLockFingerprintDoesNotMatch` error.

```
Failed to submit proofs for batch 19: Error(BoundlessMarket Err: RequestLockFingerprintDoesNotMatch(RequestLockFingerprintDoesNotMatch { requestId: 914015857174204705534535675374558860140958506052279566612, provided: 0xfb199bc5cde89189, locked: 0x0000000000000000 }))
```

Super weird that it provided a reasonable fingerprint, but the locked is all zeroes.
I saw in the contract that there are some possible reasons for that:

1. Request already fulfilled: The most likely scenario is that the request was already fulfilled, but you're trying to fulfill it again. When a request is fulfilled, the contract zeroes out the lock data, including the fingerprint.

2. Request slashed: If the request was previously slashed, the fingerprint would also be zeroed out.

In the _fulfillVerifiedLocked function, the contract first checks:
```
// Check pre-conditions for transferring payment.
if (lock.prover == address(0)) {
    // NOTE: This check is not strictly needed, as the fact that the lock has already been
    // zeroed out means zero value can be transferred. It is provided for clarity.
    return abi.encodeWithSelector(RequestIsFulfilled.selector, RequestId.unwrap(id));
}
```

But it seems I'm proceeding past this check, which is odd if the fingerprint is zeroed. When a request is fulfilled, the lock is zeroed out here:
```
// Zero-out the lock to indicate that payment has been delivered and get a bit of a refund on gas.
requestLocks[id] = RequestLock(address(0), uint96(0), uint64(0), uint96(0), bytes8(0));
```

Similarly, when a request is slashed, the fingerprint is zeroed here:
```
// Zero out deadline, stake and fingerprint in storage to indicate that the request has been slashed.
RequestLock storage lockStorage = requestLocks[requestId];
lockStorage.deadline = 0;
lockStorage.stake = 0;
[...]

PR #417, opened 2025-03-18
BM-669: implementation of estimating capacity given prover KHz

austinabell:
Pointing at #413 and has a similar implementation, but can afford to be a bit more naive about how it estimates the proving time. This is not an ideal solution, but is helpful for the time being for issues like #351 where broker is overlocking orders.

### DIFF
--- a/.env.broker-template
+++ b/.env.broker-template
@@ -28,5 +28,7 @@ PRIVATE_KEY=
 BOUNDLESS_MARKET_ADDRESS=
 SET_VERIFIER_ADDRESS=
 
+# Note you must use an RPC URL that supports event filters. Many free RPC providers 
+# do not support them.
 RPC_URL=
 ORDER_STREAM_URL=
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,6 +45,13 @@ jobs:
 
   rust:
     runs-on: [self-hosted, Linux, X64, prod, cpu]
+    services:
+      postgres:
+        image: postgres:latest
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_PASSWORD: password
     steps:
       - name: checkout code
         uses: actions/checkout@v4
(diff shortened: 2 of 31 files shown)


nategraf (commented):
This approach seems like a good first pass to me. The idea of using an estimate of the queue depth is reasonable. Combined with limiting the max concurrency, it should apply back-pressure when things are proving slower than expected, although possibly not proactively enough.

Do you think it would be possible to update the `prover_available_at` timestamp if the prover is ahead of its estimated schedule? One way I could imagine this working would be set an estimated proving time on the order in the DB and using a query to calculate the `prover_available_at` time by adding all the estimates for pending orders.

Another, possibly simpler, option might be set `prover_available_at` to now when/if the queue becomes empty. Not sure if we have a good way to detect and update when the queue is empty.

Obviously there a million ways to slice this, with the right solution being some balance of simplicity and behaving well under various load scenarios.

austinabell:
> This approach seems like a good first pass to me. The idea of using an estimate of the queue depth is reasonable. Combined with limiting the max concurrency, it should apply back-pressure when things are proving slower than expected, although possibly not proactively enough.
> 
> Do you think it would be possible to update the `prover_available_at` timestamp if the prover is ahead of its estimated schedule? One way I could imagine this working would be set an estimated proving time on the order in the DB and using a query to calculate the `prover_available_at` time by adding all the estimates for pending orders.
> 
> Another, possibly simpler, option might be set `prover_available_at` to now when/if the queue becomes empty. Not sure if we have a good way to detect and update when the queue is empty.
> 
> Obviously there a million ways to slice this, with the right solution being some balance of simplicity and behaving well under various load scenarios.

Open to this, but I'll argue that this is would be adding more complexity and moving pieces with still having cases where this would lead to inaccurate estimates. Most notably, if an order completed early _because_ it used resources before previously locked orders (assuming no priority scheduling in taskdb), then it seems like it would have the case where that adjustment would lead to the broker potentially taking on more orders that would put the remaining orders at risk of completing in time.

Would prefer to make any more complex implementation when we have the infrastructure/instrumentation to measure the actual performance of these things. But, I don't feel strongly about this so if you use the phrase "I still think this would be useful" -- I'm happy to make one of these changes now :)

nategraf (approved):
Let's go with this implementation. We'll get to see how it behaves in practice.

PR #418, opened 2025-03-18
BM-661: Add license and description so that we can publish balance-alerts-layer

nategraf:


nategraf:
`check-publish` job is expected to fail on this PR. Once this PR is merged, I can publish the crate to fix this.

crates/boundless-market/src/contracts/bytecode.rs
capossele, line 1: Are we sure we need this change? We have not changed any contract here no?

nategraf: Ah, that's an accident!

capossele (approved):

crates/boundless-market/src/contracts/bytecode.rs
nategraf: fixed in c0a0f83678131118dbcae3c3e5a3a0c3de87b677

willemolding (approved):
👍

PR #419, opened 2025-03-19
BM-575: E2E Example for Composition

Wollac:
This PR adds an example similar to `examples/counter` to showcase how proof composition can be used using Boundless and unaggregated proofs.

It also switches all the example to the default prover and runs them using CUDA on the CI.

Fixes BM-575


Wollac:
Blocked by #385

crates/guest/util/identity/Cargo.toml
```diff
@@ -7,3 +7,4 @@ edition = "2021"
 
 [dependencies]
 risc0-zkvm = { version = "1.2", default-features = false, features = ["std"] }
+risc0-zkvm-platform = { version = "1.2", features = ["sys-getenv"] }
```
Wollac, line 10: This makes the `identity` guest insecure and vulnerable to dev-mode attacks. I don't think it's a problem, as this is just a test image, but it's definitely not great.

crates/guest/util/identity/src/main.rs
```diff
@@ -1,16 +1,25 @@
...
+    Receipt,
+};
 
 risc0_zkvm::guest::entry!(main);
 
 pub fn main() {
-    let claim: ReceiptClaim = env::read();
-    env::verify_integrity(&claim).unwrap();
+    let image_id: Digest = env::read();
+    let receipt: Receipt = env::read();
```
Wollac, line 19: This guest was not used anywhere, so I repurposed it for the composition example to accept receipts via input.
Alternatively, we could create a new guest just for this purpose.

crates/guest/util/identity/src/main.rs
```diff
@@ -1,16 +1,26 @@
...
+    sha::{Digest, Digestible},
+    Receipt,
+};
 
 risc0_zkvm::guest::entry!(main);
 
 pub fn main() {
-    let claim: ReceiptClaim = env::read();
-    env::verify_integrity(&claim).unwrap();
+    let bytes = env::read_frame();
```
Wollac, line 18: This guest was not used anywhere, so I repurposed it for the composition example to accept receipts via input.
Alternatively, we could create a new guest just for this purpose.

examples/composition/apps/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+            .wallet(EthereumWallet::from(deployer_signer))
+            .connect(&anvil.endpoint())
+            .await
+            .unwrap();
+        let counter = Counter::deploy(&deployer_provider, test_ctx.set_verifier_address).await?;
+        Ok(*counter.address())
+    }
+
+    #[test(tokio::test)]
+    #[ignore = "does not work with RISC0_DEV_MODE=1"]
```
Wollac, line 307: This means that this test does not currently run in the CI. However, the test works with the local prover (on x86) and Bonsai, so we could switch the CI to use it.

.github/workflows/main.yml
```diff
@@ -385,6 +385,7 @@ jobs:
       matrix:
         workspace:
           - examples/counter
+          - examples/composition
```
capossele, line 388: How about adding something like (**_untested_**):
```suggestion
        include:
          - workspace: examples/counter
             dev-mode: true
          - workspace: examples/composition
             dev-mode: false
```

Wollac: Fixed by running all examples using the local prover.

capossele (approved):

PR #428, opened 2025-03-21
BM-655: Update `risc0` and `risc0-ethereum` to v2.0.0

capossele:
## Notes:

- Reproducible build of the `assessor-guest` requires that `cargo-risczero` is installed with the `experimental` feature.
- With risc0 2.0 all the guest binaries (i.e.,ELFs) are the one with `.bin` file extension 

## TODO:
- [x] Upgrade Bento
- [x] Check reproducible builds
- [x] Fix Malformed ProgramBinary error


Cargo.toml
```diff
@@ -51,21 +51,21 @@ hex = { version = "0.4", default-features = false, features = ["alloc"] }
...
-risc0-binfmt = "~1.2"
-risc0-build = { version = "~1.2", features = ["docker"] }
-risc0-build-ethereum = { version = "~1.4" }
-risc0-circuit-recursion = { version = "~1.2" }
-risc0-ethereum-contracts = { version = "~1.4" }
-risc0-zkvm = { version = "~1.2", default-features = false }
-risc0-zkp = { version = "~1.2" }
+risc0-aggregation = { version = "0.4.0-rc.2" }
+risc0-binfmt = "2.0.0"
+risc0-build = { version = "2.0.1", features = ["docker", "unstable"] }
```
willpote, line 56: Why do we need unstable?

willpote (approved):
LGTM, though I'm not fully up to date on which exact versions with which features enabled for each package we should be using.

I also don't have context to review the Union changes, but I assume they were reviewed in a separate PR?

Thanks for driving this!

Cargo.toml
capossele: our assessor guest uses the bigint accelerator that requires some experimental features to be enabled and that's why we need the unstable feature there

PR #429, opened 2025-03-21
BM-672: simplify skip order logic in order picker

austinabell:
avoids case where a skip order set in DB is missed, also had misleading errors. Log levels were also set too high for some of the logs that would warn on expected behaviour. Kept the warnings for low gas/stake levels, though.


austinabell:
looped this into #433

PR #431, opened 2025-03-21
BM-674: Implement `Prover` trait for the `risc0_zkvm::default_prover`

Wollac:
This makes it possible to run the broker tests and proofs locally in CI.
It also allows transparent switching between mock, local, and bonsai by simply changing environment variables.


capossele (approved):

crates/broker/src/provers/default.rs
```diff
@@ -0,0 +1,490 @@
...
+#[derive(Debug, Default)]
+pub struct DefaultProver {
+    state: Arc<ProverState>,
+}
+
+#[derive(Debug, Default)]
+struct ProverState {
+    inputs: Mutex<HashMap<String, Vec<u8>>>,
+    images: Mutex<HashMap<String, Vec<u8>>>,
+    proofs: Mutex<HashMap<String, ProofData>>,
```
austinabell, line 27: Is `Mutex` required over a `RwLock`? Assume some tests will have some parallelism and just wondering if can avoid lock contention. No big deal, just curious.

austinabell (approved):

rlukata (commented):
would be nice to have an `execute only` mode

capossele:
> would be nice to have an `execute only` mode

That can be enabled with RISC0_DEV_MODE=1

crates/broker/src/provers/default.rs
Wollac: I am not sure if this is really critical, but I changed it to `RwLock`

PR #432, opened 2025-03-21
Update license headers to BSL and add copy of Apache license to repo root

nategraf:
This updates the header on files under the BSl license to point to the appropriate license. This also adds a copy of the Apache license to the repo, since many of the files are under that license. This also changes the header of documentation, examples, and the slasher to indicate Apache license.


lib/risc0-ethereum
nategraf, line 1: Is this an intentional change?

willpote (approved):

PR #433, opened 2025-03-22
BM-676: refactor order picker capacity logic

austinabell:
Current issue is that the order picker doesn't really have a mechanism to update this reserved capacity in the case of a failed lock operation. This refactors the logic such that if idle and no pricing tasks scheduled, this reserved capacity/time is reset to current time.

Also updates the logic of the pricing function to decrement the reservation if the order fails the order pricing checks, as there was this similar race condition there as well.


crates/broker/src/order_picker.rs
```diff
@@ -524,19 +549,45 @@ where
...
+    /// There is no concurrent lock limit.
+    Unlimited,
+}
+
+impl Capacity {
+    fn request_size(&self, pricing_tasks: usize) -> u32 {
+        match self {
+            Capacity::Idle(capacity) | Capacity::PartiallyLocked(capacity) => std::cmp::min(
+                capacity.saturating_sub(u32::try_from(pricing_tasks).expect("tasks u32 overflow")),
+                10,
```
nategraf, line 577: It's not clear from reading this why there is a cap of 10. My understanding is that this limits the number of concurrent pricing jobs are running at any time, such that if the max pending locks is more than 10 it will still only work on pricing 10 at a time. Is that accurate? If so, it would be good to add a comment to that effect and possible extract this constant to a named value.

nategraf (approved):

crates/broker/src/order_picker.rs
austinabell: [`d43c7e1` (#433)](https://github.com/boundless-xyz/boundless/pull/433/commits/d43c7e10ca821801266cb01e92ffcda14b3d3718)

Yes, that is accurate. Previously this was limited by the fact that orders were pulled with a sleep inbetween, and this just keeps the behaviour similar after switching to doing in batches.

Issue #437, opened 2025-03-24
Prevent re-ordering of fulfillments in batches


willpote:
Veridise Audit 2

When verifying a batch of fulfillments, the `BoundlessMarket` eventually calls `verifyBatchDelivery` that performs the following: 

1. Verifies all application receipts.
2. Creates the `batchRoot` from all `claimDigests`.
3. Verifies the `assessorReceipt`.

The set of claimDigests are connected to the `assessorReceipt` via the `batchRoot` which is part of the `AssessorJournal` (see snippet below). However, because the tree is constructed with a commutative hash function (see `MerkleProofish.sol`), its leaves can be permuted in multiple ways while resulting in the same `batchRoot`.  

> Snippet from `verifyBatchDelivery`
> 
> 
> ```solidity
> // Verify the application receipts.
> for (uint256 i = 0; i < fills.length; i++) {
>     Fulfillment calldata fill = fills[i];
> 
>     requestDigests[i] = fill.requestDigest;
>     claimDigests[i] = ReceiptClaimLib.ok(fill.imageId, sha256(fill.journal)).digest();
>     
>     // [VERIDISE]: verifying application receipts
>     // If the requestor did not specify a selector, we verify with DEFAULT_MAX_GAS_FOR_VERIFY gas limit.
>     // This ensures that by default, client receive proofs that can be verified cheaply as part of their applications.
>     if (!hasSelector[i]) {
>         VERIFIER.verifyIntegrity{gas: DEFAULT_MAX_GAS_FOR_VERIFY}(Receipt(fill.seal, claimDigests[i]));
>     } else {
>         VERIFIER.verifyIntegrity(Receipt(fill.seal, claimDigests[i]));
>     }
> }
> 
> // [VERIDISE]: creating batchRoot from claimDigests
> bytes32 batchRoot = MerkleProofish.processTree(claimDigests);
> 
> // Verify the assessor, which ensures the application proof fulfills a valid request with the given ID.
> // NOTE: Signature checks and recursive verification happen inside the assessor.
> bytes32 assessorJournalDigest = sha256(
>     abi.encode(
>         AssessorJournal({
>             requestDigests: requestDigests,
>             root: batchRoot,
>             callbacks: assessorReceipt.callbacks,
>             selectors: assessorReceipt.selectors,
>             prover: assessorReceipt.prover
>         })
>     )
> );
> // [VERIDISE]: verifying the assessorReceipt
[...]

Issue #436, opened 2025-03-24
Handle missing validation of RequestIds during fulfillment


willpote:
Veridise Audit Issue 1

In order to fulfill proof requests, a prover must provide a set of fulfillments as represented by the following struct:

> Snippet from `struct Fulfillment`
> 
> 
> ```solidity
> struct Fulfillment {
>     /// @notice ID of the request that is being fulfilled.
>     RequestId id;
>     /// @notice EIP-712 digest of request struct.
>     bytes32 requestDigest;
>     // [VERIDISE]: ommiting rest of fields
> ```
> 

The intended invariant that every object of this struct must satisfy is that the `requestDigest` corresponds to a proof request whose `RequestId` is `id`. However, none of the fulfillment functions (i.e., `fulfill` and `fulfillBatch` ) validate that the provided id corresponds to a proof request with digest requestDigest.

## Impact

This missing check gives attackers the liberty to choose arbitrary `RequestId`s when fulfilling requests. Since the provided `RequestId` is used to determine which account to be charged (see attached snippet below), an attacker can steal funds from users by manipulating the value of the fulfillment’s `id`.

> Snippet from `_fulfillAndPay`
> 
> 
> ```solidity
> // [VERIDISE]: Fetching the client account
> RequestId id = fill.id;
> (address client, uint32 idx) = id.clientAndIndex();
> Account storage clientAccount = accounts[client];
> (bool locked, bool fulfilled) = clientAccount.requestFlags(idx);
> // [VERIDISE]: Charging the account depending on the value of locked.
> if (locked) {
>     RequestLock memory lock = requestLocks[id];
>     if (lock.lockDeadline >= block.timestamp) {
>         paymentError = _fulfillAndPayLocked(lock, id, client, idx, fill.requestDigest, fulfilled, prover);
>     } else {
>         paymentError = _fulfillAndPayWasLocked(lock, id, client, idx, fill.requestDigest, fulfilled, prover);
>     }
> } else {
>     paymentError = _fulfillAndPayNeverLocked(id, client, idx, fill.requestDigest, fulfilled, prover);
> }
> ```
> 

## Proof of Concept

The following foundry test demonstrates the issue by shuffling the request ids of a request batch. It is recommended to add this test (or a variant of it) as a negative test in the test suite.

> Proof of concept.
> 
> 
> ```solidity
> function testFulfillBatchUnLockedRequestsPOC() public {
>     // Provide a batch definition as an array of clients and how many requests each submits.
[...]

PR #441, opened 2025-03-25
BM-679: Commit to requestId and requestDigest in Assessor

capossele:
This PR addresses both #436 and #437 issues by committing to the positional index, the `requestId` and `requestDigest` as part of the leaf body in the assessor tree. 

Closes BM-678
Closes BM-679


willpote (approved):

crates/assessor/src/lib.rs
```diff
@@ -66,6 +66,29 @@ impl Fulfillment {
...
 }
 
+/// Computes a commitment hash.
+///
+/// The commitment hash is computed as the keccak256 hash of the concatenation of the following:
+/// - The index of the request in the batch.
+/// - The request ID.
+/// - The request digest.
+/// - The claim digest.
+pub fn commit(
```
nategraf, line 76: We should use a struct here and domain separation (e.g. EIP-712) instead of just concatting these bytes.

contracts/test/TestUtils.sol
```diff
@@ -21,13 +21,14 @@ library TestUtils {
...
-            claimDigests[i] = ReceiptClaimLib.ok(fills[i].imageId, sha256(fills[i].journal)).digest();
+            bytes32 claimDigest = ReceiptClaimLib.ok(fills[i].imageId, sha256(fills[i].journal)).digest();
+            leaves[i] = keccak256(abi.encodePacked(i, fills[i].id, fills[i].requestDigest, claimDigest));
             requestDigests[i] = fills[i].requestDigest;
         }
-        bytes32 root = MerkleProofish.processTree(claimDigests);
+        bytes32 root = MerkleProofish.processTree(leaves);
 
         bytes memory journal = abi.encode(
             AssessorJournal({
```
nategraf, line 34: We should also drop the `requestDigests` field from `AssessorJournal` since is is now redundant.

crates/assessor/src/lib.rs
capossele: Gas released issue fixed in [cdaceae](https://github.com/boundless-xyz/boundless/pull/441/commits/cdaceae45cb35924c6036775741b3170cae6fa3a)

nategraf (approved):
Hashing looks good to me now. Thanks for taking care of this!

PR #442, opened 2025-03-25
BM-680: Prevent re-entrency during fulfillment

### LINKED ISSUE #438: Prevent re-entrency during fulfillment
Veridise Audit Issue 3

The `fulfill` and `fulfillBatch` functions are vulnerable to a **reentrancy attack** due to an unsafe external call made via `_executeCallback` before executing `_fulfillAndPay`. This order of execution allows a malicious client to exploit the callback mechanism to re-enter the contract and manipulate their funds, allowing them to withdraw their balance before the prover gets paid for currently unlocked requests.

The root cause of the vulnerability is that `_executeCallback` is executed before `_fulfillAndPay` in both functions. Additionally, `_fulfillAndPay` does not revert when the client lacks sufficient funds, which guarantees the feasibility of the exploit for unlocked requests.

> Snippet from `fulfill`
> 
> 
> ```solidity
> // Execute the callback with the associated fulfillment information.
> // Note that if any of the following fulfillment logic fails, the entire transaction will
> // revert including this callback.
> if (assessorReceipt.callbacks.length > 0) {
>     AssessorCallback memory callback = assessorReceipt.callbacks[0];
>     _executeCallback(fill.id, callback.addr, callback.gasLimit, fill.imageId, fill.journal, fill.seal);
> }
> 
> paymentError = _fulfillAndPay(fill, assessorReceipt.prover);
> emit ProofDelivered(fill.id);
> ```
> 

## Impact

A malicious client can use reentrancy to withdraw their funds before `_fulfillAndPay` executes, preventing the prover from being compensated.

## Recommendation

To mitigate this attack, `_fulfillAndPay` should be executed **before** `_executeCallback`. This ensures that the prover is compensated before the external call is made.

capossele:
Closes #438


contracts/src/BoundlessMarket.sol
```diff
@@ -393,6 +383,16 @@ contract BoundlessMarket is
             paymentError[i] = _fulfillAndPay(fills[i], assessorReceipt.prover);
             emit ProofDelivered(fills[i].id);
         }
+
+        // Execute the callback with the associated fulfillment information.
+        // Note that if any of the following fulfillment logic fails, the entire transaction will
```
willpote, line 388: Can remove

willpote (approved):

nategraf (approved):
Thanks for taking care of this!

PR #448, opened 2025-03-26
BM-644: Setup for aws deployments

willpote:
Initial setup for deploying from the Boundless repo. Goals here are to:
1. allow us to update our prover and other infra with just 1 PR, speeding development 
2. house business logic and the infra code in one repo, simplifying development of AWS apps, where business logic and infra are often tightly coupled 
3. move us closer to continuous deployments

Note there are three pulumi apps in this initial PR:
1. The `bootstrap` app deploys a role to each service account. This role is used by the `pipelines` app to deploy to each service account.
4. The `pipelines` app contains the Pulumi state bucket and keys, and AWS CodePipeline deployment pipelines that deploy to staging/prod when code is merged. Currently it contains one pipeline that deploys a `sample` app.
5. The `sample` app is intended as a sample/template app, showing local develop, and secret management.

The AWS account structure is that we have an ops account that contains the deployment pipelines, Pulumi keys, and Pulumi state bucket. This account deploys cross-account to the staging account and the prod account. We also have a development account, that each developer can use to deploy their own stacks for testing.


zeroecco (commented):
I suggest scoping down anything wild carded to the bare minimums. It is my opinion that we should approach this from an deny list default, only allowing things as they immediately present themselves as necessary on a per resource basis.

infra/pipelines/accountConstants.ts
rlukata, line 1: at some point it would be nice to automate getting these values instead of hardcoding but i think it's fine for the time being

infra/pipelines/components/pulumiSecrets.ts
```diff
@@ -0,0 +1,87 @@
...
+      accountId: string;
+      encryptKmsKeyArns: string[];
+      decryptKmsKeyArns: string[];
+    },
+    opts?: pulumi.ComponentResourceOptions
+  ) {
+    super('pipelines:PulumiState', name, args, opts);
+
+    const pulumiSecretsKey = new aws.kms.Key(
+      'pulumiSecretsKey',
```
rlukata, line 19: why not use `awskey`?

infra/pipelines/components/pulumiState.ts
```diff
@@ -0,0 +1,192 @@
...
+      accountId: string;
+      readOnlyStateBucketArns: string[];
+      readWriteStateBucketArns: string[];
+    },
+    opts?: pulumi.ComponentResourceOptions
+  ) {
+    super('pipelines:PulumiState', name, args, opts);
+
+    // Create key used for encrypting the Pulumi state bucket
+    const pulumiStateBucketKey = new aws.kms.Key(
```
rlukata, line 20: same here, why not use `awskeys`?

infra/pipelines/components/pulumiState.ts
```diff
@@ -0,0 +1,192 @@
...
+    // Boundless Pulumi backend state bucket. Used by all pipelines and staging/prod 
+    // for tracking state for their deployments. Not used for local development.
+    this.bucket = new aws.s3.BucketV2(
+      'boundlessPulumiStateBucket',
+      {
+        bucket: 'boundless-pulumi-state',
+      },
+      {
+        parent: this,
+        protect: true,
```
rlukata, line 100: might be worth adding `retainOnDelete`

infra/bootstrap/Pulumi.services-staging.yaml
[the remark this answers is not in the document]
willpote: I just redeployed to test, and Pulumi automatically populates it with the secret provider that is configured. So I updated to have that.

infra/bootstrap/index.ts
```diff
@@ -0,0 +1,27 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+
+const OPS_ACCOUNT_PIPELINE_ROLE_ARN = "arn:aws:iam::968153779208:role/pipeline-role-3b97f1a";
```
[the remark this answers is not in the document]
willpote: If it is an env variable then it wouldn't be checked into the repo, so the deployment wouldn't be reproducible

I think it would have to go into the stack config, which I was trying to avoid because i'd have to duplicate it across the staging and prod stack

infra/pipelines/accountConstants.ts
willpote: Agreed, it looks like it might be possible via "Stack References", but unclear if it works cross-account. Will investigate in a later PR.

infra/pipelines/components/pulumiState.ts
rlukata: resolving as we discussed offline

PR #451, opened 2025-03-28
use cpus instead of cpu_count

zeroecco:
ubuntu and linux use "cpus". "cpu_count" is for windows only

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -89,7 +89,7 @@ services:
         CUDA_OPT_LEVEL: 1
 
     mem_limit: 4G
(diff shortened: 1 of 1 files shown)


drouyang (approved):

rlukata (approved):

austinabell (approved):
works for my version of docker compose:
```
docker inspect fce | grep NanoCpus
            "NanoCpus": 2000000000,
```

but reading online some others mentioning that `cpu_count` at some point did replace `cpus`, perhaps was just windows diff 🤷

PR #453, opened 2025-03-28
BM-682: Add cli support for unaggregated proofs and callbacks

capossele:
This PR revamps the `boundless-cli` by grouping the commands in 4 categories:
- `account`
- `request`
- `proving`
- `ops`

It also adds a `config` command to check the configuration is working with the market.

It adds unit test for all the cli commands and fixes various issues we had with the CLI.

Closes BM-631
Closes BM-683
Closes BM-493
Closes BM-530


.env.localnet-template
```diff
@@ -1,15 +1,16 @@
 # Default anvil node and order stream endpoint
-RPC_URL="http://localhost:8545"
-ORDER_STREAM_URL="http://localhost:8585"
+export RPC_URL=http://localhost:8545
+export ORDER_STREAM_URL=http://localhost:8585
```
austinabell, line 3: why are these changes made? Noting this breaks in my shell so wondering the reasoning

capossele: Somehow that was the only way I was able to source the .env-* files. But if it doesn’t work with your shell maybe it’s my shell all weird

capossele: I'll revert this and find an alternative for my case

austinabell: And you were using `source <.env file>`? Noting I'm using fish, and I can just update a script to handle this case, just curious why the existing format wouldn't work. I thought the existing pattern was standard

capossele: basically I believe that sourcing only works within the same process, so for instance it works if you do it before a cargo run, but it does not work if you do it before a boundless-cli invocation (i.e., the installed binary)

capossele: an alternative could be to use set-a and set +b, maybe we can try with that

austinabell: Not sure I follow, what is the workflow that is not working, so I can try in bash to see what you're referring to?

capossele: sure, so start a local devnet, (or use the one on sepolia, it does not matter), then source the corresponding .env, then invoke the `boundless-cli` with whatever command, if you use this branch you can even use the `config` subcommand. That should not work, meaning the env variables don't get propagated to that process. If instead you do the same, but invoke the cli through cargo run --bin .. it works

austinabell: hmm yeah I see the issue. Well, if this is the cleanest way, and it _seems_ like it might be -- probably best to keep the export and I or whoever this is broken for can do a workaround.

Testing on `bash` and `zsh` I noticed that this did not work with `cargo run --bin` either (and assume it shouldn't, because it runs a built binary the same).

Another option, which I assume might have been the fallback before, is to have this config manipulation target `.env` instead of `.env.localnet` by default and use `dotenvy` to pull in env vars from that config (assuming running from within boundless directory). I opted to avoid this to not cause contention on the .env whenever starting localnet vs running commands against another network like the testnet. Perhaps that is a cleaner path? Curious why that behaviour was removed, not seeing `dotenvy` being used anymore (still in `Cargo.toml` of the cli)

Obviously your call, just sharing thoughts and options.

capossele: I've ended up with this: https://github.com/boundless-xyz/boundless/pull/453/commits/d47d2d2ff02644d5649c7a1c7a04934ef852f617

PR #454, opened 2025-03-28
BM-545: Fixes and example for decentralized clients

willpote:
* Adds an example for Smart Contract Clients.
* Adds additional methods to the SDK for submitting requests with a custom signature
* Updates the market monitor to support and validate smart contract signed requests (this adds an RPC call to the market monitor, but seems better than locking a request with an invalid signature)


capossele (approved):
beautiful

PR #456, opened 2025-03-31
BM-690: End to end example for callbacks

capossele:


examples/counter-with-callback/apps/src/main.rs
```diff
@@ -0,0 +1,304 @@
...
+        args.set_verifier_address,
+        args.counter_address,
+    )
+    .await?;
+
+    Ok(())
+}
+
+/// Main logic which creates the Boundless client, executes the proofs and submits the tx.
+async fn run<P: StorageProvider>(
```
willpote, line 91: Since a lot of the comments in this function are repeated in the original `counter` example, I think we could remove them and try and make the example as focused on just callbacks as possible (this is what i did in the smart contract requestor example)

willpote (approved):

PR #461, opened 2025-04-01
BM-707: Use read and write fram I/O functions in assessor

nategraf:
Currently on `main`, the assessor input is written/read using a custom length-value encoding. The `write_frame` and `read_frame` functions now exist to handle this parttern. This PR refactors to drop the manual version of this.

It also renames `AssessorInput::to_vec` to `encode` and adds a `decode` function to contain the usage of postcard to an opaque API, in case we need to change it later.


crates/assessor/src/lib.rs
```diff
@@ -82,14 +82,14 @@ pub struct AssessorInput {
...
-        result.extend_from_slice(&length.to_le_bytes());
-        result.extend_from_slice(&bytes);
-        result
+    /// Serialize the [AssessorInput] to a bytes vector.
+    pub fn encode(&self) -> Vec<u8> {
+        postcard::to_allocvec(&self).unwrap()
+    }
+
+    /// Deserialize the [AssessorInput] from a slice of bytes.
+    pub fn decode(bytes: &[u8]) -> Result<Self, impl core::error::Error> {
```
austinabell, line 91: Having this error type be `impl Error` might make this a bit harder to maintain over time and avoid breaking changes. No strong opinion here, but having it be a concrete type (and hide the inner error type if you want to avoid breaking changes in future) seems like a better option.

Curious the motivation, because if you instead intended for it to be the postcard error type, why not specify that directly?

austinabell (approved):

crates/assessor/src/lib.rs
nategraf: Good call, keeping me honest here. I've added errors created with `thiserror` in 796f4915b6b723d68b2ac7d7283cea8a39e59f9e and use this over anyhow everywhere in this library now.

PR #463, opened 2025-04-01
BM-708: Move `boundless alerts layer` to `boundless-market` and bump version to v0.7.0

capossele:
This PR moves the `boundless-alerts-layer` to be a module of the `boundless-market` crate so that we break the dependency generating pain when checking the publishable status of the `boundless-market` crate. 
Moreover, it bumps the workspace version to `v0.7.0` in preparation for the next release


willpote (approved):
Don't forget the Zeth order generator, we skip it in CI because it takes a long time to build.

Thanks for doing this!

Issue #457, opened 2025-03-31
Add domain separation between client and prover signatures


willpote:
Veridise Audit: V-RISC0-VUL-006

Note: We should explore the domain separation option first (assess impact on complexity and gas). Idea is to wrap the ProofRequest object that the _prover_ signs into a structure like `LockRequest` 

---

The function `lockRequestWithSignature` accepts as parameters a `request`, `clientSignature`, and `proverSignature`. The validity of a request is verified using the `clientSignature`. Whereas the `proverSignature` is used to retrieve the prover address and lock the request on behalf of the signing prover.

However, in both cases, the same `requestHash` is used to recover the address as there is no domain separation for a client and a prover: `ECDSA.recover(requestHash, "client/prover"Signature)` 

> Snippet from functions `_extractProverAddress` and `_verifyClientSignature`
> 
> 
> ```solidity
> function _extractProverAddress(bytes32 requestHash, bytes calldata proverSignature) ... returns (address)
> {
>    return **ECDSA.recover(requestHash, proverSignature)**;
> }
> 
> function _verifyClientSignature(ProofRequest calldata request, address addr, bytes calldata clientSignature) ... returns (bytes32)
> {
>    bytes32 requestHash = _hashTypedDataV4(request.eip712Digest());
>    if (request.id.isSmartContractSigned()) {
>       if (IERC1271(addr).isValidSignature(requestHash, clientSignature) != IERC1271.isValidSignature.selector) {
>          revert IBoundlessMarket.InvalidSignature();
>       }
>    } else {
>       if (**ECDSA.recover(requestHash, clientSignature)** != addr) {
>          revert IBoundlessMarket.InvalidSignature();
>       }
>    }
>    return requestHash;
> }
> ```
> 

## Impact

An attacker can call `lockRequestWithSingature` as follows: `lockRequestWithSignature(request, clientSignature, clientSignature)`, that is providing the `clientSignature` as a `provingSignature`. This will affect accounts that behave as both a client and a prover in the protocol, because they are likely to have positive `balance` and `stakeBalance`s. Consequently, the aforementioned call to `lockRequestWithSignature` would lead to a case where the requesting client would also lock their proof request. This could be abused to grieve the client through slashing and delays in fulfillment.

> Function `lockRequestWithSignature`
> 
> 
> ```solidity
> function lockRequestWithSignature(
>         ProofRequest calldata request,
>         bytes calldata clientSignature,
>         bytes calldata proverSignature
>     ) external 
[...]

PR #464, opened 2025-04-01
BM-711: BM-704: Domain separation between client/prover sigs

willpote:
Addresses #457 

This implementation has a slightly dodgy api, but saves 10k gas compared to a more naive implementation (see: 85b0c5b0201a61a1756e159bf224c8fdfd655258)


contracts/src/types/LockRequest.sol
```diff
@@ -0,0 +1,55 @@
...
+import {Input, InputType, InputLibrary} from "./Input.sol";
+import {Requirements, RequirementsLibrary} from "./Requirements.sol";
+import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import {IBoundlessMarket} from "../IBoundlessMarket.sol";
+import {IERC1271} from "@openzeppelin/contracts/interfaces/IERC1271.sol";
+
+using LockRequestLibrary for LockRequest global;
+
+/// @title Lock Request Struct and Library
+/// @notice Wraps a ProofRequest. Ensures domain separation between client signatures and prover signatures.
```
nategraf, line 20: This comment will be more like historical information as time passes. If someone ever asks "why does this exist?" we can provide this information.
```suggestion
/// @notice Message sent by a prover to indicate that they intend to lock the given request.
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -828,20 +829,43 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+        client.expectBalanceChange(0 ether);
+        testProver.expectBalanceChange(0 ether);
+        expectMarketBalanceUnchanged();
+    }
+
+    function testLockRequestWithSignatureProverSignatureIncorrectDomain() public {
+        Client client = getClient(1);
+        ProofRequest memory request = client.request(1);
+        bytes memory clientSignature = client.sign(request);
+        // Prover signs ProofRequest struct rather than LockRequest struct.
```
nategraf, line 859:
```suggestion
        // Prover signs ProofRequest struct rather than LockRequest struct.
        // NOTE: This was how the contract worked in a previous version. This is included as a regression test.
```

nategraf (commented):
The broker is certainly broken by these changes. If you prefer, I can take a stab at that. I started looking at it yesterday towards the end of the day.

nategraf (approved):

PR #468, opened 2025-04-02
BM-718: Use risc0-ethereum v2.0.1

capossele:


Wollac (approved):

crates/guest/assessor/assessor-guest/src/main.rs
```diff
@@ -86,8 +85,8 @@ fn main() {
         }
     }
 
-    // recompute the merkle root of the aggregation set
-    let root = merkle_root(&leaves);
+    // compute the merkle root of the commitments
+    let root = process_tree(leaves);
```
capossele, line 89: Here we need to decide if we want to go this way or we have to modify the solidity function MerkleProofish.process_tree to tag the leaves as it happens in the risc0-aggregation crate

nategraf: In this case we don't need to tag leaves in the same way since they are EIP712 tagged and we don't provide an interface that allows you to pass anything that could be confused for a node hash.

nategraf (approved):

PR #469, opened 2025-04-02
BM-511 Pick slashed unexpired orders

willemolding:
Adds the following to allow the broker to detect when orders have been slashed and use different pricing logic

- New tasks in MarketMonitor that monitors for orders being locked and filled by other provers. This allows the broker to get a better view of the protocol state and it required for checking when an order has had its lock expired but has not been filled

- Extends `update_orders_for_pricing` to also update orders that meet the slashed+unexpired requirement

- Adds new config fields 
- - `fill_slashed_orders_altruistically` which skips profitability checks for slashed orders if true
- - `mcycle_price_stake_token` which is the equivalent of `mcycle_price` but expressed in stake tokens. This should be removed once the broker has a view into the stake token price relative to the gas token

- Updates `price_order` logic to:
- - Correctly set the pricing estimate as a constant fraction of the stake for slashed orders
- - Detect that the expiry is the order expiry and not the lock expiry for slashed orders and the lockin stake is zero
- - Skip unprofitable slashed orders if configured to do so

Much of the churn in the diff is due to a refactor of the boundless_market crate test utils. This has been moved to its own crate allowing it to expose more code for testing using the market while also being used within the market crate itself. 

> If it is decided we don't want that refactor the branch up to commit [1cb2e14](https://github.com/boundless-xyz/boundless/pull/469/commits/1cb2e140ba2252da5e9e3080e40e40084b3f9de4) has all the order picking features prior to the refactor


willemolding:
One design decision up for discussion. Here I monitor for `ProverSlashed` events on-chain to detect when orders are in the slashed+unfilled state but another way to do it would be to monitor for `RequestFulfilled` events and observe a lack of fulfilled event once the lock expiry has passed.

I think both would work. One possible advantage of monitoring for `RequestFulfilled` is that it might give the prover a bit of a headstart to price the order and start proving if the slashing is slow to occur. It would still have to wait for the slashing event to submit though so we would have to monitor for both.

Anyone have any thoughts either way?

willpote:
> One design decision up for discussion. Here I monitor for `ProverSlashed` events on-chain to detect when orders are in the slashed+unfilled state but another way to do it would be to monitor for `RequestFulfilled` events and observe a lack of fulfilled event once the lock expiry has passed.
> 
> I think both would work. One possible advantage of monitoring for `RequestFulfilled` is that it might give the prover a bit of a headstart to price the order and start proving if the slashing is slow to occur. It would still have to wait for the slashing event to submit though so we would have to monitor for both.
> 
> Anyone have any thoughts either way?

Haven't had a chance to review yet (will look tomorrow morning), but just to answer these questions; currently its not possible for a request to be slashed until it is fully expired (https://github.com/boundless-xyz/boundless/blob/main/contracts/src/BoundlessMarket.sol#L691-L695). So I think monitoring for no RequestFulfilled + lockTimeout is the only approach that can enable requesters to get their proofs before we hit the final `timeout`.

> It would still have to wait for the slashing event to submit though so we would have to monitor for both.
Whys that? The intention with the design was that it would be the other way around- provers filling after lockTimeout have to make sure to submit _before_ the slashing event (and thus before `timeout`), to ensure they will get paid. This is because when `slash` is called that is when payment is settled (so if they submit after slash, they won't be compensated with their portion of the slashed stake)

crates/broker/src/order_picker.rs
```diff
@@ -117,12 +127,30 @@ where
...
+        let order_expiration =
+            order.request.offer.biddingStart + order.request.offer.timeout as u64;
 
         let now = now_timestamp();
+        // If order_expiration > lock_expiration the period in-between is when order can be filled
+        // by anyone without staking to partially claim the slashed stake
+        let slashed_unexpired = order.status == OrderStatus::Slashed
+            && lock_expiration <= now
+            && order_expiration > now;
+
```
willpote, line 143: Related to my other comment, but I don't think this works since requests are only slashed once they fully expire. I think we'll need another state that represents "LockExpiredUnfulfilled", and to track the `lockTimeout` for each value

crates/broker/src/order_picker.rs
```diff
@@ -332,35 +359,47 @@ where
...
 
         let one_mill = U256::from(1_000_000);
+        let total_cycles = U256::from(proof_res.stats.total_cycles);
+
+        let (min_price, max_price) = if slashed_unexpired {
+            let price = order.request.offer.lockStake / U256::from(FRACTION_STAKE_REWARD);
+            (price, price)
+        } else {
+            (U256::from(order.request.offer.minPrice), U256::from(order.request.offer.maxPrice))
+        };
```
willpote, line 369: Since the stake token and regular price are different currencies, I think its best if we branch here into two separate paths of code to avoid confusion. There is also an issue in that currently provers specify "min_mcycle_price" in terms of ETH, so we can't use it to decide whether to pick an order where the price is in stake token. I think we'll need to introduce a prover config like "min_mcycle_price_stake" to handle this.

crates/broker/src/order_picker.rs
```diff
@@ -332,35 +359,47 @@ where
...
         // Skip the order if it will never be worth it
-        if mcycle_price_max < config_min_mcycle_price {
+        if mcycle_price_max < config_min_mcycle_price && !ignore_profitability {
             tracing::warn!("Removing under priced order {order_id:x}");
             self.db.skip_order(order_id).await.context("Failed to delete order")?;
             return Ok(false);
         }
 
-        if mcycle_price_min >= config_min_mcycle_price {
+        if mcycle_price_min >= config_min_mcycle_price || ignore_profitability {
```
willpote, line 399: One issue is that we schedule the lock expired orders to be locked via the below `self.db.set_order_lock`. But its not actually possible to lock these orders anyone- they have to just be directly fulfilled without locking

crates/broker/src/order_picker.rs
```diff
@@ -520,9 +559,11 @@ where
             return Ok(());
         }
 
-        let order_res = self.db.update_orders_for_pricing(capacity).await?;
+        let new_orders = self.db.update_orders_for_pricing(capacity).await?;
+        let lock_expired_orders =
+            self.db.update_slashed_unexpired_for_pricing(capacity, now_timestamp()).await?;
```
willpote, line 564: Will re-using capacity here double the max capacity that the broker can handle? We probably want to let provers specify a separate configuration for fulfilling lockExpired orders.

willpote (commented):
Thank you for taking this on! This definitely seems quite tricky to get right.

Left some comments, happy to discuss more in the PR or hop on a call if that would be easier.

crates/broker/src/order_picker.rs
willemolding: Ok I switch the pricing strategy depending on the order type and added a new config field `mcycle_price_stake_token`. This is a fill-in until the broker can read the price of the stake token from somewhere

crates/broker/src/order_picker.rs
willemolding: Added a new return type from order pricing that allows specifying what to do with the order (Lock, Skip, ProveImmediate). The code calling the picker can then update the order state in the db accordingly

crates/broker/src/order_picker.rs
willemolding: I merged these to functions into 1 that does a more complex db query and respects the same capacity parameter

crates/broker/src/market_monitor.rs
```diff
@@ -56,8 +57,9 @@ where
         provider: Arc<P>,
         db: DbObj,
         chain_monitor: Arc<ChainMonitorService<P>>,
+        submitter_addr: Address,
```
willpote, line 60: `submitter_addr` makes me think this is the Proof Requestor. `prover_addr` feels more accurate

crates/broker/src/order_picker.rs
```diff
@@ -30,6 +30,14 @@ use boundless_market::{
 use thiserror::Error;
 use tokio::task::JoinSet;
 
+use OrderPricingOutcome::{Lock, ProveImmediate, Skip};
+
+// fraction the stake the protocol gives to the prover who fills an order that was locked by another prover but expired
+// e.g. a value of 1 means 1/4 of the original stake is given to the prover who fills the order.
+// This is determined by the constant SLASHING_BURN_BPS defined in the BoundlessMarket contract.
+// The value is 4 because the slashing burn is 75% of the stake, and we give the remaining 1/4 of that to the prover.
+const FRACTION_STAKE_REWARD: u64 = 4;
```
willpote, line 39: This is fine for now, but we should really fetch this from the contract. I created a task here: https://github.com/boundless-xyz/boundless/issues/517 Could you add a TODO?

crates/broker/src/order_picker.rs
```diff
@@ -218,25 +258,27 @@ where
...
             self.db.skip_order(order_id).await.context("Failed to delete order")?;
-            return Ok(None);
+            return Ok(Skip);
         }
 
         if order_gas_cost > available_gas {
             tracing::warn!("Estimated there will be insufficient gas to lock and fill order {order_id:x} after locking and fulfilling pending orders; available_gas {} ether", format_ether(available_gas));
             self.db.skip_order(order_id).await.context("Failed to delete order")?;
-            return Ok(None);
+            return Ok(Skip);
```
willpote, line 274: For lock expired orders, when we fulfill we don't actually need to lock, so the `order_gas_cost` variable on L252 ends up being an over-estimate, meaning we'll skip more orders than needed here. 

The logging here and above also mentions locking which could be confusing. Would be good to update it based on `lock_expired`

crates/broker/src/order_picker.rs
```diff
@@ -259,7 +301,7 @@ where
 
         if skip_preflight {
             // If we skip preflight we lockin the order asap
-            return Ok(Some(OrderLockTiming { target_timestamp_secs: 0, expiry_secs: expiration }));
+            return Ok(Lock { target_timestamp_secs: 0, expiry_secs: expiration });
```
willpote, line 304: We wouldn't be able to lock a lock expired order, so we need to check if its lock expired or not

willpote (commented):
Nice! This is looking very close!

crates/broker/src/market_monitor.rs
willemolding: changed

crates/broker/src/config.rs
```diff
@@ -126,12 +128,20 @@ pub struct MarketConf {
     ///
     /// Maximum number of concurrent lockin requests that can be processed at once
     pub max_concurrent_locks: Option<u32>,
+    /// Fill slashed orders altruistically
+    ///
+    /// If an order is detected to have been slashed but is still unexpired
+    /// setting this to true will have the broker attempt to fill the order regardless of profitability.
+    /// This is useful for provers who want to provide a liveness backstop for slashed orders
+    #[serde(default)]
+    pub fill_slashed_orders_altruistically: bool,
```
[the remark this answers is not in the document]
willemolding: Oh no turns out there isn't! I removed this

crates/broker/src/order_picker.rs
willemolding: done

crates/broker/src/order_picker.rs
willemolding: altered that gas computation to skip the locking gas for a lock expired order, also just changed the warning message a little bit so it applies in either case

crates/broker/src/order_picker.rs
willemolding: added extra conditional here

willpote (approved):
Nice! 

Could you also add the new config to:
./broker.toml
./infra/prover/broker-dev.toml
./infra/prover/broker-staging.toml
./infra/prover/broker-prod.toml

PR #470, opened 2025-04-03
BM-644: Deploy bonsai provers

willpote:
Deploy provers from the repo. Deploys a prover to staging environment, that points at a staging instance of the market. There is then a manual approval for deploying to prod, which points at our production instance. 

Provers can also be deployed by each developer to a development account.

* Prover logic is mostly a copy paste. Main changes are config related to get dev stacks working.
* VPC logic is a copy paste.
* The bootstrap stack was updated so that it can be deployed by each developer to the development account, so each dev has their own VPC.
* One issue I ran into was rate limiting from Github and Docker during builds running from AWS CodeBuild. To address this we use Github and Docker tokens to authenticate for higher rate limits.


dockerfiles/broker.dockerfile
```diff
@@ -11,8 +11,16 @@ RUN curl -L https://foundry.paradigm.xyz | bash && \
...
+# Github token can be provided as a secret with the name githubTokenSecret. Useful
+# for shared build environments where Github rate limiting is an issue.
+RUN --mount=type=secret,id=githubTokenSecret,target=/run/secrets/githubTokenSecret \
+    if [ -f /run/secrets/githubTokenSecret ]; then \
+        GITHUB_TOKEN=$(cat /run/secrets/githubTokenSecret) curl -L https://risczero.com/install | bash && \
+        GITHUB_TOKEN=$(cat /run/secrets/githubTokenSecret) PATH="$PATH:/root/.risc0/bin" rzup install rust 1.85.0; \
+    else \
+        curl -L https://risczero.com/install | bash && \
+        PATH="$PATH:/root/.risc0/bin" rzup install rust 1.85.0; \
+    fi
```
rlukata, line 23: im not 100% certain about this but i believe you don't need to pass the github token to the dockerfile itself, it just needs to be available in the env where the docker image is being built [[src](https://docs.docker.com/build/cache/backends/gha/#avoid-github-actions-cache-api-throttling)]

dockerfiles/broker.dockerfile
```diff
@@ -45,8 +53,9 @@ RUN \
     --mount=type=cache,target=/root/.cache/sccache/,id=bndlss_broker_sc \
     source ./sccache-config.sh ${S3_CACHE_PREFIX} && \
     cargo build --release --bin broker && \
-    cp /src/target/release/broker /src/broker && \
-    sccache --show-stats
+    cp /src/target/release/broker /src/broker
+
+RUN sccache --show-stats
```
rlukata, line 58: why? the less `run` the faster in theory no?

infra/README
```diff
@@ -0,0 +1,22 @@
...
+
+1. Vault in as developer admin
+1. `pulumi login "s3://boundless-pulumi-state?region=us-west-2&awssdk=v2"`
+1. `pulumi new aws-typescript --secrets-provider="awskms:///arn:aws:kms:us-west-2:968153779208:alias/pulumi-secrets-key" --name prover --stack staging --yes`
+1. `pulumi stack init --secrets-provider="awskms:///arn:aws:kms:us-west-2:968153779208:alias/pulumi-secrets-key" prod`
+1. `pulumi login --local`
+1. `pulumi stack init dev`
+1. `git add -u && git commit -m "<commit message>"`
+1. Add `Pulumi.dev.yaml` to `.gitignore`
+1. `git add -u && git commit -m ".gitignore"`
```
rlukata, line 22: Users that can't access `boundless-pulumi-state` or `pulumi-secrets-key` won't be able to access this. Who is this workflow aimed at?

dockerfiles/broker.dockerfile
willpote: I tried that initially but was still having issues until I added this. I also don't think docker builds would be repeatable if they could access env variables outside of the container (also seems like that might be a security risk?)

dockerfiles/broker.dockerfile
willpote: Forgot to remove after debugging, moving back

infra/pipelines/pipelines/prover.ts
```diff
@@ -0,0 +1,247 @@
...
+  role: aws.iam.Role;
+  githubToken: pulumi.Output<string>;
+  dockerUsername: string;
+  dockerToken: pulumi.Output<string>;
+}
+
+// The name of the app that we are deploying. Must match the name of the directory in the infra directory.
+const APP_NAME = "prover";
+// The branch that we should deploy from on push.
+const BRANCH_NAME = "willpote/deploy-prover";
```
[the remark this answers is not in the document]
willpote: Yes planning to update this to main before I push

infra/README
willpote: Thats a good point. I've split this into 2 parts- one for creating dev stacks which anyone can do, then a part for creating staging/prod which Ops admin permissions are needed for.

infra/pipelines/pipelines/prover.ts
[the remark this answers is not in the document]
willpote: I just updated it now.

dockerfiles/broker.dockerfile
rlukata: yeah looks like i was wrong about this, although the `--build-arg` is available, the docker image ignores it if it's not declared in the image. To test this I wrote this `Dockerfile`:
```
FROM alpine
RUN echo ${GITHUB_TOKEN}
ARG GITHUB_TOKEN
RUN echo ${GITHUB_TOKEN}
```
and this cli command to test it (you'll have to review the build logs):
```
docker build -t test-docker --build-arg GITHUB_TOKEN=test123 . --progress=plain --no-cache
```

rlukata (approved):

PR #475, opened 2025-04-03
BM-723: Upgrade deployment v0.7.0

capossele:


willpote (approved):

austinabell (approved):
Wondering if we update `.env.testnet` with this or separately. Also can get rid of the env file, but it can be convenient to load in those values

PR #476, opened 2025-04-03
wip switching from EFS/fargate to ec2

austinabell:


infra/prover/index.ts
```diff
@@ -0,0 +1,338 @@
...
+        {
+          Effect: 'Allow',
+          Action: ['secretsmanager:GetSecretValue', 'ssm:GetParameters'],
+          Resource: [privateKeySecret.arn, bonsaiSecret.arn],
+        },
+      ],
+    },
+  });
+
+  const brokerEcr = new awsx.ecr.Repository(serviceName, {
```
rlukata, line 124: One thing I would avoid using `awsx` and stick `aws`. There's many reasons why but the main one is that the `awsx` package doesn't have all the component resource functions that the `aws` sdk has.

rlukata:
What's the goal of this PR? Why install and run docker inside an EC2 when ECS/Fargate solve this problem very well?

infra/prover/index.ts
```diff
@@ -0,0 +1,338 @@
...
+        cidrBlocks: ['0.0.0.0/0'],
+        ipv6CidrBlocks: ['::/0'],
+      },
+    ],
+  });
+
+  // Set up CloudWatch logging for the EC2 instance
+  const brokerLogGroup = new aws.cloudwatch.LogGroup(serviceName, {
+    name: serviceName,
+    retentionInDays: 0,
```
rlukata, line 216: If you select 0, the events in the log group are always retained and never expire. Is this what we want?

austinabell:
> What's the goal of this PR? Why install and run docker inside an EC2 when ECS/Fargate solve this problem very well?

To switch to EBS given our issues running sqlite on EFS. This PR is in draft, and will change before coming in so no need to worry about it yet

rlukata:
> > What's the goal of this PR? Why install and run docker inside an EC2 when ECS/Fargate solve this problem very well?
> 
> To switch to EBS given our issues running sqlite on EFS. This PR is in draft, and will change before coming in so no need to worry about it yet

is there a limitation in using EBS where we need to switch to EC2? @willpote

PR #482, opened 2025-04-04
BM-729: Improve bento performance

zeroecco:
* parallel process union work
    * async process prove and keccak
    * breakout union stream


zeroecco:
in my testing this shaves roughly 100-200ms from the join workflow, ~50-100ms for prove and keccak workflows, and allows for breakout of union streams

bento/crates/workflow/src/tasks/executor.rs
```diff
@@ -161,25 +161,28 @@ async fn process_task(
...
-                .context("Invalid assumption count conversion")?;
-
-            let mut prereqs = vec![tree_task.depends_on[0].to_string()];
-            let mut union_max_idx: Option<usize> = None;
-
-            if !tree_task.keccak_depends_on.is_empty() {
-                prereqs.push(format!("{}", tree_task.keccak_depends_on[0]));
+            // Handle union work - if we have keccak/union work, treat it specially
+            let has_union_work = !tree_task.keccak_depends_on.is_empty();
+            let keccak_count = if has_union_work { 1 } else { 0 };
```
rlukata, line 166: this is the same as what it was, no?

bento/Cargo.toml
```diff
@@ -26,6 +26,7 @@ aws-sdk-s3 = "1.34" # used for minio for max compatibility
 bincode = "1.3"
 bytemuck = "1.16"
 deadpool-redis = "0.15"
+futures = "0.3"
```
rlukata, line 29: why futures and not tokio::spawn?

bento/crates/workflow/src/tasks/join.rs
```diff
@@ -8,46 +8,135 @@ use crate::{
...
-        .await?;
 
-    tracing::info!("Join Complete {job_id} - {}", request.left);
+    // Serialize and store in separate future to return quickly
+    let job_id_clone = *job_id;
+    let left_idx = request.left;
+    let ttl = agent.args.redis_ttl;
+
+    // Use spawn_local if available to reduce thread creation overhead
+    tokio::task::spawn(async move {
```
rlukata, line 111: why spawn a thread at all? seems like we could keep it in the same thread

bento/crates/api/src/helpers.rs
```diff
@@ -86,13 +86,26 @@ pub async fn get_or_create_streams(
             .context("Failed to create taskdb snark stream")?
     };
 
+    let gpu_union_stream = if let Some(res) = taskdb::get_stream(pool, user_id, UNION_WORK_TYPE)
```
rlukata, line 89: (minor nit) rename `gpu_union_stream` to `union_stream`

bento/crates/workflow/src/tasks/keccak.rs
```diff
@@ -4,13 +4,13 @@
 
 use crate::{
     redis::{self, AsyncCommands},
-    tasks::{serialize_obj, COPROC_CB_PATH},
+    tasks::{serialize_obj, COPROC_CB_PATH, RECEIPT_PATH},
```
rlukata, line 7: i suppose it works the same since receipt paths use `uuid`, but i would prefer to either keep `KECCAK_RECEIPT_PATH` or siwtch to `RECUR_RECEIPT_PATH`

bento/crates/workflow/src/tasks/keccak.rs
```diff
@@ -51,28 +51,58 @@ pub async fn keccak(
...
-    redis::set_key_with_expiry(
-        &mut conn,
-        &receipts_key,
-        keccak_receipt_bytes,
-        Some(agent.args.redis_ttl),
-    )
-    .await
-    .context("Failed to write keccak receipt to redis")?;
+    // Spawn background task for serialization and storage
+    tokio::spawn(async move {
```
rlukata, line 70: i don't think we benefit from using this here, imho spawn makes sense when we have multiple tasks

bento/crates/workflow/src/tasks/prove.rs
```diff
@@ -35,21 +35,36 @@ pub async fn prover(agent: &Agent, job_id: &Uuid, task_id: &str, request: &Prove
...
-        .with_context(|| format!("Failed to lift segment {index}"))?;
+    // Clone necessary data for the cleanup task
+    let pool = agent.redis_pool.clone();
+    let output_key = format!("{job_prefix}:{RECUR_RECEIPT_PATH}:{task_id}");
+    let ttl = agent.args.redis_ttl;
 
-    tracing::info!("lifting complete {job_id} - {index}");
+    // Spawn a separate task to handle serialization and storage
+    // This allows the main function to return immediately and pick up the next GPU task
+    tokio::spawn(async move {
```
rlukata, line 45: same as before, don't think we benefit

bento/crates/workflow/src/tasks/resolve.rs
```diff
@@ -8,104 +8,154 @@ use crate::{
...
                     let union_receipt: SuccinctReceipt<Unknown> =
-                        deserialize_obj(&union_receipt)
+                        deserialize_obj(&union_receipt_bytes)
                             .context("Failed to deserialize to SuccinctReceipt<Unknown> type")?;
-                    union_claim = union_receipt.claim.digest().to_string();
 
-                    // Resolve union receipt
-                    tracing::info!("Resolving union claim digest: {union_claim}");
+                    // Resolve union receipt first
+                    tracing::info!("Resolving union receipt");
```
rlukata, line 75: why remove the union claim? could be helpful

bento/crates/workflow/src/tasks/join.rs
austinabell: tasks != threads

but assumed that this is done such that other work doesn't wait on storing the result (this task doesn't block the function from returning)

bento/crates/workflow/src/tasks/union.rs
```diff
@@ -4,33 +4,48 @@
...
+    let right_receipt_key = format!("{receipts_prefix}:{}", request.right);
+
+    // Fetch left receipt first
+    let left_receipt_bytes: Vec<u8> = conn
+        .get(&left_receipt_key)
+        .await
+        .with_context(|| format!("Left receipt not found for key: {left_receipt_key}"))?;
+
+    // Start deserializing left receipt
+    let left_receipt_future = tokio::task::spawn_blocking(move || -> Result<_> {
```
rlukata, line 32: i seriously doubt this is gaining any peformance

bento/Cargo.toml
austinabell: it's used for `join`, but regardless I would think https://docs.rs/tokio/latest/tokio/macro.join.html / https://docs.rs/tokio/latest/tokio/macro.try_join.html should cover this use case (possibly missing something)

bento/crates/workflow/src/tasks/resolve.rs
```diff
@@ -8,104 +8,154 @@ use crate::{
...
-                    conditional_receipt = agent
-                        .prover
-                        .as_ref()
-                        .context("Missing prover from resolve task")?
-                        .resolve(&conditional_receipt, &assumption_receipt)
-                        .context("Failed to resolve the conditional receipt")?;
+                    // Wrap the pool in an Arc for thread-safe sharing
+                    let pool = Arc::new(agent.redis_pool.clone());
+
+                    let results = stream::iter(assumption_tasks)
```
rlukata, line 111: im clear on what's going on but im not positive that this is faster when payloads have a relatively small amount of assumptions, do you have a use case where we have a large amounts of assumptions?

bento/crates/workflow/src/tasks/join.rs
```diff
@@ -8,46 +8,135 @@ use crate::{
...
+                    .context("Failed to deserialize right receipt")
+            }
+        },
+    )
+    .await;
 
-    tracing::info!("Joining {job_id} - {} + {} -> {}", request.left, request.right, request.idx);
+    // Unwrap results
+    let left_receipt = left_receipt?;
+    let right_receipt = right_receipt?;
```
austinabell, line 90: If exiting on either failure, perhaps https://docs.rs/tokio/latest/tokio/macro.try_join.html is better, such that on first error it will short circuit and not wait for the other.

bento/crates/workflow/src/tasks/join.rs
rlukata: you're right, [tokio spawns a task that runs in a thread](https://docs.rs/tokio/latest/tokio/task/fn.spawn.html)
but i mean that we don't need to spawn anything here imho

bento/crates/workflow/src/tasks/resolve.rs
```diff
@@ -8,104 +8,154 @@ use crate::{
...
+
+                                Ok::<_, anyhow::Error>((receipt, claim))
+                            }
+                        })
+                        .buffer_unordered(concurrency_limit)
+                        .collect::<Vec<_>>()
+                        .await;
+
+                    // Process results sequentially (resolving must be sequential)
+                    for (i, result) in results.into_iter().enumerate() {
```
austinabell, line 136: Is this intentional that they are ordered based on the order each futures finishes? Assume you might want https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffered over the unordered variant?

bento/crates/workflow/src/tasks/union.rs
austinabell: How expensive is the deserialization? If expensive, then maybe worth running other deserializations in a blocking task, if no then perhaps this logic could be simplified to just be a join between the two?

bento/crates/workflow/src/tasks/join.rs
zeroecco: [0918e72](https://github.com/boundless-xyz/boundless/pull/482/commits/0918e72c4274985d24c76c9a26fed740a0cacb57)

bento/crates/workflow/src/tasks/resolve.rs
zeroecco: [0918e72](https://github.com/boundless-xyz/boundless/pull/482/commits/0918e72c4274985d24c76c9a26fed740a0cacb57)

bento/crates/api/src/helpers.rs
zeroecco: [0918e72](https://github.com/boundless-xyz/boundless/pull/482/commits/0918e72c4274985d24c76c9a26fed740a0cacb57)

bento/Cargo.toml
zeroecco: [0918e72](https://github.com/boundless-xyz/boundless/pull/482/commits/0918e72c4274985d24c76c9a26fed740a0cacb57)

bento/crates/workflow/src/tasks/resolve.rs
zeroecco: cannot defend my approach. [0918e72](https://github.com/boundless-xyz/boundless/pull/482/commits/0918e72c4274985d24c76c9a26fed740a0cacb57)

bento/crates/workflow/src/tasks/resolve.rs
zeroecco: [0918e72](https://github.com/boundless-xyz/boundless/pull/482/commits/0918e72c4274985d24c76c9a26fed740a0cacb57)

bento/crates/workflow/src/tasks/union.rs
rlukata: not expensive enough to spawn a new task, checkout bincode's perf [here](https://github.com/djkoloski/rust_serialization_benchmark?tab=readme-ov-file#raw-data), it's says deserializing 1mb takes about 2ms. it's not the same payload but i doubt it's, on average, going to make a signficant difference

zeroecco: when a jobs total wall time is ~800ms, if I can shave 8ms wall time that is a real 1% perf bump. I am looking for any and all laggy bits:
29401813

bento/crates/workflow/src/tasks/resolve.rs
austinabell: I think it was fine, it just needed to be ordered? Noting that this change removed the parallelism (which I assume was the goal) and can still be achieved. Are you saying that running concurrently didn't give perf here?

zeroecco: it did show performance improvements from testing; but not enough to warrant more work.

PR #483, opened 2025-04-07
BM-644: Slack alerts

willpote:
Sets up Slack alerts using Amazon Q Chatbot, and configures some basic alarms

Note I opted to have alarms defined within each service, and forward the alarms notifications to the ops account, rather than defining the alarms within the ops account as we do in Bonsai. I prefer this because:
1/ It lets developers test alarms when deploying to their dev accounts.
2/ It places alarms closer to their infrastructure, making it easier to see what is covered by alarms vs not.
3/ Ops account admin is locked down, so this design enables all devs to create alarms.


infra/pipelines/pipelines/prover.ts
```diff
@@ -1,20 +1,14 @@
...
-  dockerUsername: string;
-  dockerToken: pulumi.Output<string>;
-}
+interface ProverPipelineArgs extends BasePipelineArgs {}
 
 // The name of the app that we are deploying. Must match the name of the directory in the infra directory.
 const APP_NAME = "prover";
 // The branch that we should deploy from on push.
-const BRANCH_NAME = "main";
+const BRANCH_NAME = "willpote/slack";
```
rlukata, line 11: this shuold be main

infra/prover/index.ts
```diff
@@ -3,15 +3,15 @@ import * as aws from '@pulumi/aws';
...
 
 export = () => {
   // Read config
   const config = new pulumi.Config();
 
-  const isDev = pulumi.getStack() === "dev";
-  const prefix = isDev ? `${getEnvVar("DEV_NAME")}-` : "";
-  const serviceName = `${prefix}bonsai-prover`;
+  const stackName = pulumi.getStack();
+  const isDev = stackName === "dev";
```
rlukata, line 13: it doesn't look you're you're using `isDev`?

rlukata (approved):
minor nits

infra/prover/index.ts
willpote: Using it on L25+

rlukata: oh, should have looked on the left 🤦🏽

PR #486, opened 2025-04-08
BM-40: Account for estimated gas costs when pricing orders

nategraf:
On `main` , gas costs are not taken into account when pricing an order. This is a significant issue as gas costs often dwarf the computational costs of a proof when the proof is not sufficiently large.

This PR improves the basic gas estimation we currently use for gas balance management, and uses it to add the gas costs into the pricing function used for picking orders.

It also updates the `ChainMonitor` to refactor how it gets the block timestamp, and adding a call to get the gas price estimate (as reported by the `eth_gasPrice` RPC) along side the call to get the latest block up to every X seconds.

Note this is a significant change to how pricing is done. We also need to account for these changes on the client side when building a request. This PR covers the broker changes, and I will create a follow-up PR with the client changes to make it easier to set a reasonable price on a given request.


crates/broker/src/order_picker.rs
```diff
@@ -464,10 +510,37 @@ where
     /// Estimate of gas for locking a single order
     /// Currently just uses the config estimate but this may change in the future
     async fn estimate_gas_to_lock(&self, _order: &Order) -> Result<u64> {
+        // TODO: Add gas costs for orders with smart contract signatures.
```
willpote, line 513: Hmm, we should probably set a gas limit when we call isValidSignature in the contracts, and use that here

crates/broker/src/order_picker.rs
```diff
@@ -464,10 +510,37 @@ where
...
+            .context("unsupported selector")?
+        {
+            ProofType::Any | ProofType::Inclusion => Ok(base),
+            ProofType::Groth16 => Ok(base + groth16),
+            proof_type => {
+                tracing::warn!("Unknown proof type in gas cost estimation: {proof_type:?}");
+                Ok(base)
+            }
+        }
+    }
```
willpote, line 538: We also need to account for callbacks

crates/broker/src/config.rs
```diff
@@ -28,11 +28,20 @@ mod defaults {
...
+        // https://sepolia.etherscan.io/tx/0xe61b5cad4a45fc0913cc966f8e3ee72027c01a949a9deca916780e1245c15964
+        200_000
     }
 
     pub const fn fulfill_gas_estimate() -> u64 {
-        300_000_000
+        // Observed cost of a basic single fulfill transaction is ~350k gas.
+        // Additional padding is used to account for journals up to 10kB in size.
+        // https://sepolia.etherscan.io/tx/0x14e54fbaf0c1eda20dd0828ddd64e255ffecee4562492f8c1253b0c3f20af764
+        750_000
```
willpote, line 40: Wow.. this is high. Would it make sense to use the results of running the executor to get the journal size for each order?

willpote: Also is 10kb a hard limit we set somewhere?

crates/broker/src/order_picker.rs
nategraf: I've completed these changes now

crates/broker/src/order_picker.rs
nategraf: I've added gas equal to the callback limit

crates/broker/src/config.rs
nategraf: It's the default max size at which the prover will simply drop the request. We could try to estimate the effect of the journal size on gas costs. This number is large since its trying to be an overestimate. Maybe this is too conservative of an approach and drives up costs too much. Long-term, better gas estimation is an opportunity to compete as a prover and we expect folks to have their own way of doing it.

crates/broker/src/order_picker.rs
```diff
@@ -509,9 +510,17 @@ where
...
     /// Currently just uses the config estimate but this may change in the future
-    async fn estimate_gas_to_lock(&self, _order: &Order) -> Result<u64> {
-        // TODO: Add gas costs for orders with smart contract signatures.
-        Ok(self.config.lock_all().context("Failed to read config")?.market.lockin_gas_estimate)
+    async fn estimate_gas_to_lock(&self, order: &Order) -> Result<u64> {
+        let mut estimate =
+            self.config.lock_all().context("Failed to read config")?.market.lockin_gas_estimate;
+
+        // Gas allocated to verifying a smart contract signature. Copied from BoundlessMarket.sol.
+        const ERC1271_MAX_GAS_FOR_CHECK: u64 = 100000;
```
willpote, line 518: Lets move to top of file so its with MAX_PRICING_BATCH_SIZE

willpote (approved):
Looks good, could you add an issue to supporting skipping fills in the contracts?

crates/broker/src/order_picker.rs
nategraf: Done in 46eb009c

nategraf:
> Looks good, could you add an issue to supporting skipping fills in the contracts?

Its tracked as BM-436 (looks like we didn't create a GitHub issue for this one}

PR #487, opened 2025-04-08
BM-721: Deploy order stream

willpote:
Deploys the Order Stream
* Packages it as a Pulumi component so we can easily deploy new instances for new chains.
* Adds versioning to the endpoints.
* I changed order stream URL to be a secret in the prover and order-generator. This is so that we don't expose the staging URL of the order stream.
* Currently DNS is not fully set up. This first PR just creates the initial cert request.


infra/prover/index.ts
```diff
@@ -128,9 +137,9 @@ export = () => {
     }),
   });
 
-  brokerEcr.repository.arn.apply(arn => {
+  pulumi.all([brokerEcr.repository.arn, executionRole.name]).apply(([ecrRepoArn, executionRoleName]) => {
     new aws.iam.RolePolicy(`${serviceName}-ecs-execution-pol`, {
-      role: executionRole.id,
+      role: executionRoleName,
```
rlukata, line 142: you don't need the `.apply` function here, im pretty sure you can just:
```
role: executionRole.name
```
You're also not using `ecrRepoArn` that i can see.

rlukata (approved):
can't speak for the rust code changes but the infra code looks good ✅

nahoc (approved):
lgtm

PR #499, opened 2025-04-10
BM-751: Implement Secure S3 Input Download via IAM Role Assumption

Wollac:
### Description

This PR introduces support for securely downloading inputs (and images) stored in S3, leveraging AWS IAM Roles for granular access control and optionally AWS KMS (SSE-KMS) for enhanced data protection through server-side encryption. It addresses the requirement for partners to share sensitive input data with specific provers, replacing less secure or less flexible methods like pre-signed URLs.

### Approach Implemented (One Role per Prover):

This implementation follows an approach where each prover maintains its own dedicated IAM Role:

1. **Prover Configuration:**
   - Each prover is expected to create and manage its own unique IAM role within their AWS account. The ARN of this role is configured in the broker (using the new `aws_assume_role_arn` config option).
3. **Requestor Configuration:**
   - The requestor uploads input data to their S3 bucket. For maximum security, they can encrypt this data using SSE-KMS with their own KMS key.
   - They modify their S3 Bucket Policy to grant `s3:GetObject` permission only to the ARNs of the prover roles they trust (e.g., `ProverX-WorkerRole`). 
   - If using SSE-KMS: The requestor also modifies the KMS Key Policy of the key used for encryption to grant the `kms:Decrypt` permission specifically to the trusted prover role ARNs.
4. **Implementation:**
   - The requestor provides an `s3://<bucket>/<key>` URI in the proof request.
   - The new `S3Handler` is initialized. It uses `aws_config::from_env()` to load base AWS credentials. It explicitly checks if credentials can be resolved.
    - If `aws_assume_role_arn` is configured, the handler calls `sts:AssumeRole` targeting the prover's configured role ARN.
   - During `fetch()`, the S3 client calls `s3:GetObject`, this internally performs the following steps:
      - S3 first checks the Bucket Policy. If the assumed role is allowed `s3:GetObject`, the request proceeds. Otherwise, it's denied.
      - If the object is encrypted with SSE-KMS: S3 then automatically attempts decryption via KMS using the assumed role's credentials. This requires the requestor to have granted `kms:Decrypt` in the KMS Key Policy. If permission is denied here, the download fails.
5. **Security Layers:**
   - **Access Control:** The combination of the prover assuming a specific IAM Role and the requestor configuring the S3 Bucket Policy to allow `s3:GetObject` only for trusted role ARNs provides the fundamental access control, ensuring only authorized provers can attempt to download.
   - **Decryption Control:** Using SSE-KMS adds a vital second layer of security. Even if a prover has `s3:GetObject` permission, they cannot get the plaintext data unless the requestor also explicitly grants `kms:Decrypt` permission via the KMS Key Policy.


crates/broker/src/config.rs
```diff
@@ -126,6 +126,14 @@ pub struct MarketConf {
     ///
     /// Maximum number of concurrent lockin requests that can be processed at once
     pub max_concurrent_locks: Option<u32>,
+    /// Optional cache directory for storing downloaded images and inputs
+    /// if not set, files will be re-downloaded every time
+    pub cache_dir: Option<PathBuf>,
```
Wollac, line 131: This used to be part of `Args`, but after the refactoring access to the args is more difficult, also conceptually it is only used in the marked and thus should probably belong to the market config.

crates/broker/src/config.rs
```diff
@@ -126,6 +126,14 @@ pub struct MarketConf {
...
+    /// Optional S3 endpoint URL
+    pub s3_endpoint_url: Option<String>,
+    /// Optional IAM role for S3 access
+    /// The argument should take the form an Amazon Resource Name (ARN)
+    pub aws_assume_role_arn: Option<String>,
```
Wollac, line 136: Support for the IAM role is critical to take full advantage of the granular access control that AWS offers. This could also be an ENV variable (like all other AWS secrets), but since its name is not security relevant and having it in the config allows for better documentation, I went with the config, but would love to hear your thoughts on this.

crates/broker/src/config.rs
```diff
@@ -126,6 +126,14 @@ pub struct MarketConf {
     ///
     /// Maximum number of concurrent lockin requests that can be processed at once
     pub max_concurrent_locks: Option<u32>,
+    /// Optional cache directory for storing downloaded images and inputs
+    /// if not set, files will be re-downloaded every time
+    pub cache_dir: Option<PathBuf>,
+    /// Optional S3 endpoint URL
+    pub s3_endpoint_url: Option<String>,
```
Wollac, line 133: This is only needed for minio testing and unfortunately the AWS SDK config doesn't load it from ENV like all the other parameters.

crates/broker/src/storage.rs
```diff
@@ -2,296 +2,311 @@
...
-        self.max_size = Some(max_size);
-        self
-    }
+        let mut s3_conf = aws_config::from_env().retry_config(retry_config).load().await;
+        if match s3_conf.credentials_provider() {
+            None => true,
+            Some(credentials) => credentials.provide_credentials().await.is_err(),
+        } {
+            return Err(StorageErr::UnsupportedScheme("s3".to_string()));
+        }
```
Wollac, line 235: This disables S3 handling, when no AWS credentials have been set. Since we only introduced this feature to get access control and encryption this makes sense. However, in theory someone could also use a public bucket to upload data, then no credentials would be needed to fetch them.

crates/broker/src/storage.rs
```diff
@@ -2,296 +2,311 @@
...
+            RetryConfig::standard().with_max_attempts(max_retries as u32 + 1)
+        } else {
+            RetryConfig::disabled()
+        };
 
-    pub fn set_max_size(mut self, max_size: usize) -> Self {
-        self.max_size = Some(max_size);
-        self
-    }
+        let mut s3_conf = aws_config::from_env().retry_config(retry_config).load().await;
```
Wollac, line 229: This loads the AWS credentials from a file where we could document this somehow.

crates/broker/src/storage.rs
```diff
@@ -2,296 +2,311 @@
...
-            retries: retries.unwrap_or(DEFAULT_RETRY_NUMB),
-            client,
-        })
-    }
+            builder = builder.with(cache_middleware)
+        }
+        if let Some(max_retries) = max_retries {
+            let retry_policy =
+                ExponentialBackoff::builder().build_with_max_retries(max_retries as u32);
+            let retry_middleware = RetryTransientMiddleware::new_with_policy(retry_policy);
```
Wollac, line 161: Previously we had implemented manual retry logic that only retried the initial get state. While the `RetryTransientMiddleware` is much nicer, it only exits on transient errors, whereas before we didn't have that distinction and exited on everything. Conceptually, I think retiring only on transient errors should be the right choice.

crates/broker/src/config.rs
willpote: I didn't see this comment when I was doing my review, but in my opinion we should be consistent and stick with env variables. I think it is actually simpler to document, as with the current implementation we'll have to tell users "set your AWS_ACCESS_KEY_ID / SECRET in env, then set the role in config"

crates/broker/src/config.rs
willpote: Mentioned in my other comment, but I believe there actually is an env variable for this.

crates/broker/src/storage.rs
willpote: I don't follow, it loads it from env variables doesnt it?

crates/broker/src/config.rs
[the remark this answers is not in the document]
Wollac: Unfortunately, as I understand it, `AWS_ROLE_ARN` support is only available as part of the [WebIdentityTokenCredentialProvider](https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/aws-config/src/web_identity_token.rs), which is not what we want. But of course we could still reuse the same ENV name and configure it manually.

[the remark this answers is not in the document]
Wollac: According to this issue https://github.com/awslabs/aws-sdk-rust/issues/932, just using `AWS_ENDPOINT_URL` should work as expected. Nice find, so we can get rid of this config param.

crates/broker/src/storage.rs
Wollac: Oops, sorry, obviously ment loaded from env in the comment.

crates/broker/src/config.rs
Wollac: Addressed in d22f06bf29da5784c3e1284e26b4c3f7eb10bf9b

crates/broker/src/config.rs
Wollac: You are right, the `AWS_ENDPOINT_URL` env is loaded by default. Addressed in c467ce2c6c3bd0d85b10b50a14c862d940cdfa61

crates/broker/src/storage.rs
willpote: Agreed we do need to document this somewhere though. Perhaps @sashaaldrick has a suggestion

willpote (approved):

crates/broker/src/storage.rs
sashaaldrick: I have an issue for this here: https://linear.app/risczero/issue/BM-756/create-docs-for-kms-encrypted-partner-inputs

nategraf:
This failure is supremely weird. I don't see any changes that could cause checkout to fail, and testing it locally, everything works.

https://github.com/boundless-xyz/boundless/actions/runs/14606498954/job/40977249462?pr=499

Tested locally with
```sh
git clone git@github.com:boundless-xyz/boundless.git --branch wollac/feat/s3-prover
cd boundless
git submodule update --init --recursive
```

PR #505, opened 2025-04-10
BM-761: Switch most CI jobs to self-hosted runners

nategraf:
We try to use the self-hosted runners when possible for no-trivial jobs as they cost a lot less for us to run and are on higher-performance machines. This PR moves jobs in `main.yml` except the very small checks to self-hosted runners.


capossele (approved):
Made few changes, but otherwise good, thanks!

nategraf:
> Made few changes, but otherwise good, thanks!

Thanks!

PR #507, opened 2025-04-11
BM-736: Targeted benchmark proving

### LINKED ISSUE #485: Targeted proving tasks to specific provers for measuring capacity without trust
During testnet there is a need for Boundless (the company) to be able to assess the proving capabilities of specific provers on the network. Since this may be tied to future rewards this must be done in a way that doesn't rely on trusting the prover to accurately report their metrics.

Ideally this should reuse as much of the existing Boundless protocol and broker as possible and require only a config change for provers who wish to participate.

A possible workflow would be as follows:

- Provers whitelist a particular key in their broker config which will be provided to them offchain by Boundless. The holder of the private key is able to get provers to perform these assessment tasks
- For tasks originating from this key:
- - All pricing, preflight and locking should be skipped, the prover should accept and process this task asap
- - It should not be used in a batch
- - It should be submitted on-chain making a public record of the time between the order being submitted and filled. 

One important point is that there should be a way to ensure that the order cannot be filled by any prover other than the one that is being assessed. Ideally this should be done without requiring protocol changes. A simple way to do this would be to modify the guest programs used for assessment to accept one extra hidden input - a hash preimage generated by Boundless and shared exclusively with the prover. The hash of this preimage would be committed to in the journal and the verification of this hashing would be one extra step in the program execution. 

This allows only provers in possession of the preimage to build a valid proof for a given journal and the value in the journal is how the specific prover is selected. This will require a slight change to how inputs are built in the broker but this can be on the codepath selected by the key whitelisting.

willemolding:
Closes #485


crates/broker/src/config.rs
```diff
@@ -405,6 +420,36 @@ impl ConfigWatcher {
...
+
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<Option<Vec<u8>>, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        struct Helper(#[serde(with = "hex_serde")] Vec<u8>);
+
+        let helper = Option::deserialize(deserializer)?;
+        Ok(helper.map(|Helper(v)| v))
```
austinabell, line 449: This may work for toml/json, just noting that this will not work for some serialization protocols. Probably won't ever matter in our case, but wondering why handling `Option` case implicitly. `serde_with` crate would remove the need for all of this.

austinabell (commented):
Skimming this, it's unclear to me if trying to target specific provers with orders, why we would not just selectively send orders to specific addresses? _Feels_ like it should be handled at the order stream server level? If you really want to skip preflight, you just have the prover add the benchmark program ID to `skip_preflight_ids`?

Also perhaps I'm misreading the impl, but this means for brokers that are not being benchmarked will try to process the order and fail because they don't have the secret prefix? This seems unideal.

nategraf:
Moved this to draft to indicate we do not intend to merge as is, after discussion on the approach here.

PR #539, opened 2025-04-18
PER-13: BM-795: Remove bento

rlukata:
Closes PER-13


justfile
```diff
@@ -46,8 +42,8 @@ test-cargo-example:
 # Run database tests
 test-cargo-db: 
     just test-db setup
-    DATABASE_URL={{DATABASE_URL}} sqlx migrate run --source ./bento/crates/taskdb/migrations/
-    cd bento && DATABASE_URL={{DATABASE_URL}} RISC0_DEV_MODE=1 cargo test -p taskdb
+    git clone https://github.com/risc0/risc0.git && cd risc0/bento
+    DATABASE_URL={{DATABASE_URL}} sqlx migrate run --source ./crates/taskdb/migrations/
```
nategraf, line 46: Since Bento is gone, we don't want to do these migrations. We do still need to setup a DB for testing the order-stream, otherwise we'd remove this entirely.

dockerfiles/agent.dockerfile
nategraf, line 1: Why keep this file here instead of moving it to `risc0` as well.

rlukata: i was thinking that it made sense to still be able to run bento from boundless but tbh i think it's better to just remove

austinabell:
https://github.com/boundless-xyz/boundless/pull/564 has some changes to addr above and update docs. Feel free to merge it in here if you want

dockerfiles/agent.dockerfile
austinabell: until docker images published we can use, this workaround (hack) seems not too bad. Perhaps we want to have this locked to a certain commit to avoid breakages though? Definitely a bit of a pain in the short term, but seems cleaner to clone the repo inside of this docker build rather than cloning and using the other dockerfile outside of it, especially since it's just kept around for the build image and not kept in final.

willpote (approved):

PR #540, opened 2025-04-21
BM-798: remove NVCC flag from docs/compose

austinabell:
Experimenting with different NVCC flag configuration for older GPU support.

Draft until details resolved and tested it resolves compatibility.


compose.yml
```diff
@@ -85,7 +85,7 @@ services:
       # NOTE: we build a single image for all agents so specify the GPU flags here
       # TODO: can we clean this up / make it more logical?
       args:
-        NVCC_APPEND_FLAGS: "--gpu-architecture=compute_86 --gpu-code=compute_86,sm_86 --generate-code arch=compute_86,code=sm_86"
+        NVCC_APPEND_FLAGS: ${NVCC_APPEND_FLAGS:-"--gpu-architecture=compute_86 --gpu-code=compute_86,sm_86 --generate-code arch=compute_86,code=sm_86"}
```
zeroecco, line 88: this is wild; I didn't know that you can pass overrides! very cool find! maybe we should update the whole file to use this instead of passing around an env file?

austinabell: ah, I see what you're saying. Assumed this was intentional the way it is, but I can experiment with that to see if it makes things easier. There are a lot of env vars and might be more footguns of accidentally changing variables. Can do that in a follow up PR

infra/prover/components/bentoBroker.ts
```diff
@@ -457,12 +457,7 @@ export HOME=/home/ubuntu
 chmod +x /local/boundless/scripts/setup.sh
 /local/boundless/scripts/setup.sh
 
-# Install nvcc. After reboot we will set the NVCC flags in the compose.yml using this.
-sudo apt install nvidia-cuda-toolkit -y
```
austinabell, line 461: I'm about 4/5 confidence this isn't needed, but can test manually if we want to be absolutely sure it's not

willpote (approved):

Issue #162, opened 2024-11-24
Broker does not handle multiple requests with the same ID


nategraf:
Request IDs are not unique, although in most case a client will only want to submit one request per ID. However, there are some use cases for submitting multiple requests with the same ID, in particular this allows creating a set a "order variants" with a mix of e.g. pricing options or requirements.

Currently, the broker is not setup to handle this, and uses the order ID with the assumption that it unique.

PR #542, opened 2025-04-22
BM-800: Fulfill after lock expire fixes. Change order table to support multiple fulfillment methods.

willpote:
This PR firstly fixes some issues with the initial implementation for fulfilling requests where the lock has expired. The first version had two functional issues:

1. It attempted to fulfill the requests immediately when the `RequestLocked` event was detected, rather than waiting until the lock had expired. This has been resolved by storing a "target_timestamp" value that represents when the lock expires, and only attempting to fulfill after that point, similar to our logic for locking requests.

2. It attempted to call the contract to fulfill lock-expired requests without pricing them. This is resolved by using `priceAndFulfillBatch` when our batch contains lock expired orders.

These issues were missed due to us not having an E2E test. I've now added an E2E test that covers the full workflow.

---

Secondly this PR makes some changes to the orders DB table aimed at reducing complexity, race conditions, and setting us up for future features like "fulfilling orders without locking" and "multiple requests with same id". 

The initial implementation monitored for `RequestLocked` events, and if detected, updated the order's OrderStatus back to Pricing, to indicate that we should reconsider the request for fulfilling after the lock expired. This introduced a couple of issues:

1. Race Conditions: for example, if we updated the status back to Pricing while the order was in the middle of proving, when the proving workflow completed the status would be overwritten to PendingAggregation, meaning the request is never re-priced.

2. Breaks the "assembly line" mental model, adding complexity. The proving node originally processed orders like an assembly line; market_monitor -> order_picker -> order_monitor -> proving -> aggregation -> submission. Allowing an orders status to be set to a previous status, breaks the simple pipeline model and complicates the state transition graph, making things harder to debug and increasing risks of bugs.

To address this, I have changed the orders table primary key to be of type `{request_id}-{request_hash}-{fulfillment_type}`. Where fulfillment type indicates whether we are intending to "Lock and Fulfill" the request, or "Wait until the lock expires and fulfill" or "Fulfill without locking the request". So now each request will result in multiple rows in the orders table, where each row represents a proof request +and+ a specific way to fulfill it, and each row goes through the assembly line independently. (Note request hash is included in the id to enable us to support multiple requests with the same id in the future.)

---

Thirdly, this PR moves capacity tracking from the order-picker into the order-monitor. Previously we would assign capacity during order-picking, which in many cases in significantly before the time that we actually need the capacity- as we must wait until the order is profitable enough to lock, or wait until the lock expires to attempt to fulfill for slashed stake. Now we only assign capacity when we hit the `target_timestamp` and are about to either lock or prove the order (depending on the fulfillment type)

---

A couple follow ups for the task will be:

Capacity limits for executor runs (this PR removes any limitations at that stage)
Executor/proving caching - this design results in operations like executor being repeated across each order row. We can cache the results
Adding a submitMerkleAndPriceAndFulfill method to the contract. Currently fulfilling both locked and unlocked orders in one transaction is not possible.


crates/broker/src/lib.rs
```diff
@@ -203,8 +213,25 @@ struct Order {
...
+// by waiting for an existing lock to expire then fulfilling for slashed stake, or by fulfilling
+// without locking at all.
+//
+// For a given request, each type of fulfillment results in a separate Order being created, with different
+// FulfillmentType values.
+//
+// Additionally, there may be multiple requests with the same request_id, but different ProofRequest
+// details. Those also result in separate Order objects being created.
+//
+// See the id() method for more details on how Orders are identified.
```
austinabell, line 228: nit: This likely should be on the `Order` struct and be rustdoc comments (`///`)

crates/broker/src/lib.rs
```diff
@@ -217,14 +244,34 @@ impl Order {
...
             client_sig,
             lock_price: None,
+            fulfillment_type,
             error_msg: None,
         }
     }
+
+    // An Order is identified by the request_id, the fulfillment type, and the hash of the proof request.
+    // This structure supports multiple different ProofRequests with the same request_id, and different
+    // fulfillment types.
```
austinabell, line 254:
```suggestion
    /// An Order is identified by the request_id, the fulfillment type, and the hash of the proof request.
    /// This structure supports multiple different ProofRequests with the same request_id, and different
    /// fulfillment types.
```

crates/broker/src/lib.rs
```diff
@@ -217,14 +244,34 @@ impl Order {
...
+    pub fn id(&self) -> String {
+        format!(
+            "0x{:x}-{}-{:?}",
+            self.request.id,
+            self.request.signing_hash(Address::ZERO, 1).unwrap(),
```
austinabell, line 259: why is this hash necessary?

crates/broker/src/aggregator.rs
```diff
@@ -159,21 +159,21 @@ impl AggregatorService {
         })
     }
 
-    async fn prove_assessor(&self, order_ids: &[U256]) -> Result<String> {
+    async fn prove_assessor(&self, order_ids: Vec<String>) -> Result<String> {
```
austinabell, line 162: this doesn't seem to need to clone the vec of orders? It just uses a ref to the items?
```suggestion
    async fn prove_assessor(&self, order_ids: &[String]) -> Result<String> {
```

austinabell (commented):
Partial review because still personally unclear the motivation and if this is a good long-term option. Primarily:

> 1. Race Conditions: for example, if we updated the status back to Pricing while the order was in the middle of proving, when the proving workflow completed the status would be overwritten to PendingAggregation, meaning the request is never re-priced.
> 2. Breaks the "assembly line" mental model, adding complexity. The proving node originally processed orders like an assembly line; market_monitor -> order_picker -> order_monitor -> proving -> aggregation -> submission. Allowing an orders status to be set to a previous status, breaks the simple pipeline model and complicates the state transition graph, making things harder to debug and increasing risks of bugs.
> 
> To address this, I have changed the orders table primary key to be of type `{request_id}-{request_hash}-{fulfillment_type}`. Where fulfillment type indicates whether we are intending to "Lock and Fulfill" the request, or "Wait until the lock expires and fulfill" or "Fulfill without locking the request". So now each request will result in multiple rows in the orders table, where each row represents a proof request +and+ a specific way to fulfill it, and each row goes through the assembly line independently. (Note request hash is included in the id to enable us to support multiple requests with the same id in the future.)

The pattern that the broker tries to fulfill two orders at the same time feels more like a bug than something to handle with more complexity. Wondering if this could be solved with a compare and swap pattern on updates rather than handling in parallel.

willpote:
> Partial review because still personally unclear the motivation and if this is a good long-term option. Primarily:
> 
> > 1. Race Conditions: for example, if we updated the status back to Pricing while the order was in the middle of proving, when the proving workflow completed the status would be overwritten to PendingAggregation, meaning the request is never re-priced.
> > 2. Breaks the "assembly line" mental model, adding complexity. The proving node originally processed orders like an assembly line; market_monitor -> order_picker -> order_monitor -> proving -> aggregation -> submission. Allowing an orders status to be set to a previous status, breaks the simple pipeline model and complicates the state transition graph, making things harder to debug and increasing risks of bugs.
> > 
> > To address this, I have changed the orders table primary key to be of type `{request_id}-{request_hash}-{fulfillment_type}`. Where fulfillment type indicates whether we are intending to "Lock and Fulfill" the request, or "Wait until the lock expires and fulfill" or "Fulfill without locking the request". So now each request will result in multiple rows in the orders table, where each row represents a proof request +and+ a specific way to fulfill it, and each row goes through the assembly line independently. (Note request hash is included in the id to enable us to support multiple requests with the same id in the future.)
> 
> The pattern that the broker tries to fulfill two orders at the same time feels more like a bug than something to handle with more complexity. Wondering if this could be solved with a compare and swap pattern on updates rather than handling in parallel.

Yeah I considered something like that, but I actually felt like this approach would lead to less complexity. With more of a compare/swap every DB operation need to run a check, we have some optional fields that are populated later in the flow that we'd have to make sure to clear to maintain data integrity, and I think its harder to debug if we're moving backwards and forwards through Order Statuses. 

Also note, the orders don't fully end up being handled in parallel. More that we end up cancelling (setting to Skip) the existing order, and kick off a new order where we try to fulfill after the lock expires.

crates/broker/src/lib.rs
willpote: This is for future support of multiple requests with the same ID https://github.com/boundless-xyz/boundless/issues/162

crates/broker/src/lib.rs
```diff
@@ -140,8 +140,10 @@ enum OrderStatus {
     New,
     /// Order is in the process of being priced
     Pricing,
-    /// Order is ready to lock at target_timestamp
-    Locking,
+    /// Order is ready to lock at target_timestamp and then be fulfilled
+    FulfillAfterLocking,
+    /// Order is ready to be fulfilled when its lock expires at target_timestamp
+    FulfillAfterLockExpire,
```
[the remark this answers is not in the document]
willpote: Yes, definitely a breaking change, but I agree I think its ok for testnet

crates/broker/src/market_monitor.rs
```diff
@@ -282,14 +317,24 @@ where
             .into_stream()
             .for_each(|log_res| async {
                 match log_res {
-                    Ok((event, _)) => {
+                    Ok((event, log)) => {
                         tracing::info!("Detected request fulfilled {:x}", event.requestId);
-                        if let Err(e) = db.set_order_complete(U256::from(event.requestId)).await {
-                            tracing::error!("Failed to update order status to Done: {e:?}");
+                        if let Err(e) = db
+                            .set_request_fulfilled(
```
nategraf, line 323: Do we need to distinguish here on whether or not the prover was paid? Two cases where this distinction may matter if when someone fulfills an order we already locked (i.e. they don't get paid, and we are still eligible for payment), and if the order is not locked and someone else fills it but doesn't get paid due to client not having enough funds (https://github.com/boundless-xyz/boundless/issues/545). The latter of the two edge cases may be niche enough to ignore.

nategraf (commented):
Left some comments. I like the idea of having a table for events like lock and fulfill that can allow us to determine the onchain state of an order by ID, separate from the order indexed by request digest. Not sure about whether this should be one table with state or two like it is now. I think both work though.

crates/broker/src/lib.rs
willpote: Thats right, it has been useful to have it in the id for debugging. 

Good callout on the chain and contract- I meant to come back to this actually but forgot!

And correct on fulfillment type. I think having it operate separately keeps things a lot cleaner and easier to debug, and reduces risks of various race conditions.

Regarding proof work in progress, I think we should add a caching layer in the prover to the executor runs and proof runs (I think this is something we should do regardless of this PR anyway for DDOS/spam protection)

crates/broker/src/lib.rs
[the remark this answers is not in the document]
willpote: > It unclear to me why these are two different states vs being a case where the the state of the order may skip Locking and go straight to PendingProving or Proving when we don't intend to lock it.

`PendingProving` schedules it to be proved immediately, but in this case we want to wait until the lock expires (to see if the locker actually fulfills it or not) before proving.

I agree with your concerns on the state machine, though I actually think as part of Broker RPC separation we should be moving towards a different architecture for the DB- I've always found designs where multiple services are updating the same DB to be error prone and hard to scale, and I've seen many people consider it a bit of an anti-pattern ([e.g.](https://martinfowler.com/bliki/IntegrationDatabase.html)). 

My take is we should move more to a design where each service owns its own data store, and we strive to make the rows in each immutable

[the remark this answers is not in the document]
willpote: For the names open to ideas. Maybe: `WaitingToLock` and `WaitingForLockExpire`? I think its useful to align the naming to indicate that these are both states where we are waiting until some point in the future

crates/broker/src/market_monitor.rs
willpote: That would be useful to store

I'm inclined to leave this to a follow up as 1/ Currently the last point at which we consider aborting a fulfillment is right before we lock it, so handling these cases would be new code. and 2/ Its actually quite tricky to get this info at the moment since the payment failed event doesn't include the request id

crates/broker/src/order_monitor.rs
```diff
@@ -211,42 +299,411 @@ where
...
+                let expiration = match order.fulfillment_type {
+                    FulfillmentType::LockAndFulfill => order.request.lock_expires_at(),
+                    FulfillmentType::FulfillAfterLockExpire => order.request.expires_at(),
+                    _ => panic!("Unsupported fulfillment type: {:?}", order.fulfillment_type),
+                };
+
+                tracing::debug!("Order {} estimated to take {} seconds, and would be completed at {}. It expires at {}", order.id(), proof_time_seconds, completion_time, expiration);
+
+                if completion_time > expiration {
+                    tracing::info!("Order {:x} cannot be completed before its expiration at {}, proof estimated to complete at {}. Skipping", order.request.id, expiration, completion_time);
```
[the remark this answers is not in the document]
willpote: Added it back

crates/broker/src/order_monitor.rs
```diff
@@ -211,42 +299,411 @@ where
...
+            commited_orders.sort_by_key(|order| order.proving_started_at);
+            let started_proving_at = if !commited_orders.is_empty() {
+                commited_orders[0].proving_started_at.unwrap()
+            } else {
+                now_timestamp()
+            };
+
+            let proof_time_seconds = total_commited_cycles.div_ceil(1_000).div_ceil(peak_prove_khz);
+            let mut prover_available_at = started_proving_at + proof_time_seconds;
+            tracing::debug!("Already committed to {} orders, with a total cycle count of {}, a peak khz limit of {}, started working on them at {}, we estimate the prover will be available at {}", num_commited_orders, total_commited_cycles, peak_prove_khz, started_proving_at, prover_available_at);
```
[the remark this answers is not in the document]
willpote: Added a condition to skip when 0 orders

crates/broker/src/order_monitor.rs
```diff
@@ -211,42 +299,411 @@ where
...
+                final_orders.push(order);
+            }
+        }
+
+        tracing::info!(
+            "After filtering invalid orders, found total of {} valid orders to proceed to locking and/or proving", 
+            final_orders.len()
+        );
+        tracing::debug!(
+            "Final orders ready for locking and/or proving after filtering: {final_orders:?}"
```
austinabell, line 466: sorry I mixed up logs from https://github.com/boundless-xyz/boundless/pull/542/files/574d87e616bbb2be90aacc27194db23cc7fdabfa#r2066818924 this was the one I meant to point to. Perhaps if info can be useful it's made as a trace log for our debugging purposes? Debug is on by default with our default broker setup right now

crates/broker/src/order_monitor.rs
```diff
@@ -211,42 +299,411 @@ where
...
+
+                final_orders.push(order);
+                prover_available_at = completion_time;
+            }
+        } else {
+            final_orders = orders_truncated;
+        }
+
+        tracing::info!(
+            "Started with {} orders ready to be locked and/or proven. After applying capacity limits of {} max concurrent proofs and {} peak khz, filtered to {} orders: {:?}",
```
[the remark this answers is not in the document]
austinabell: https://github.com/boundless-xyz/boundless/pull/542/files/574d87e616bbb2be90aacc27194db23cc7fdabfa#r2066906879 mixed up the logs, sorry

crates/broker/src/order_monitor.rs
willpote: Oh, so I implemented a custom debug for this that keeps it short: https://github.com/boundless-xyz/boundless/pull/542/files#diff-99ebc229b742243f5d295e008af90162c013ee6455ac3d0c49f3f86ef9c2295eR84-R95

I've found it quite useful to have on debug mode, but happy to move to trace if you feel strongly

PR #549, opened 2025-04-24
Add yaml generator to `boundless-cli`

hmrtn:
Adds a simple YAML request generator for usage with the `boundless-cli`, attempting to fill the values in for you with sane(?) defaults. Automagically uploads image and input to Pinata (IPFS) with `--upload` flag and runs executor to determine journal digest, cycle count, price-ish, etc. 

If included, documentation should recommend careful review of generated requests prior to usage. 

usage: `boundless-cli request generate --elf-path <PATH_TO_ELF> --input-file/input-bytes <INPUT> --ouput-path <PATH_TO_OUTPUT_YAML>`


nategraf:
Instead of generating YAML, the approach we are taking is with the `submit-offer` command

PR #551, opened 2025-04-24
BM-889: Improve UX for benchmarking bento with orders

austinabell:
Benchmark command through boundless-cli pulls orders from on-chain or indexer and gives the benchmark results of that order.


nategraf (approved):

austinabell:
assuming the previous review as stale as there were a few changes since then

PR #556, opened 2025-04-24
BM-821: Bump prover builder pipeline memory

willpote:
Seeing `signal: 9, SIGKILL: kill` during compilation of the broker in the pipeline. This implies out of memory. Bumping the instance size.


nategraf (approved):

nategraf:
Looks like this give it 16 GB of RAM and 8 cores? That seems pretty reasonable

PR #560, opened 2025-04-25
BM-636: Add boundless-indexer

capossele:
Add indexer; supports MySql, Postgres and SQLite


crates/indexer/src/db.rs
```diff
@@ -0,0 +1,940 @@
...
+
+    async fn add_tx(&self, metadata: &TxMetadata) -> Result<(), DbError> {
+        sqlx::query(
+            "INSERT INTO transactions (
+                tx_hash, 
+                block_number, 
+                from_address, 
+                block_timestamp
+            ) VALUES ($1, $2, $3, $4)
+            ON CONFLICT (tx_hash) DO NOTHING",
```
willpote, line 221: I like that this makes the DB updates idempotent, which should make us more resilient to failures/crashes in the middle of processing a block. But with this approach the errors get swallowed- would prefer us to catch the conflict in the application code and log some sort of warning, so we can have alerts if we keep seeing this.

crates/indexer/src/lib.rs
```diff
@@ -0,0 +1,816 @@
...
+                            | ServiceError::EventQueryError(_)
+                            | ServiceError::RpcError(_) => {
+                                attempt += 1;
+                                tracing::warn!(
+                                    "Failed to process blocks from {} to {}: {:?}, attempt number {}",
+                                    from_block,
+                                    to_block,
+                                    e,
+                                    attempt
+                                );
```
willpote, line 162: Nit: should have some backoff here

crates/indexer/src/lib.rs
```diff
@@ -0,0 +1,816 @@
...
+    }
+
+    fn get(&self, tx_hash: B256) -> Option<&(TxMetadata, Bytes)> {
+        self.cache.get(&tx_hash)
+    }
+
+    fn insert(&mut self, metadata: TxMetadata, tx_input: Bytes) {
+        self.cache.insert(metadata.tx_hash, (metadata, tx_input));
+    }
+}
```
willpote, line 816: We should clear this cache periodically just to be safe

willpote (commented):
Nice!

crates/indexer/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+
+[dependencies]
+alloy = { workspace = true, features = ["network", "node-bindings", "rpc-types", "providers", "transports", "sol-types", "contract", "signers", "signer-local"] }
+anyhow = { workspace = true }
+async-trait = { workspace = true }
+boundless-market = { workspace = true }
+clap = { workspace = true }
+dotenvy = { workspace = true }
+hex = { workspace = true }
+risc0-zkvm = { workspace = true, features = ["std", "default"] }
```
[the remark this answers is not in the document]
capossele: Good call, moved to dev dependencies

crates/indexer/src/db.rs
capossele: We could refactor to catch those conflicts. The reason I've implemented this way was to keep the code as simple as possible, there may be cases where multiple events (e.g., proof delivered) are batched within the same tx. Surely, it can be refactor if you think getting out that warning is worth though

crates/indexer/src/lib.rs
capossele: Absolutely, that is cleared every time `process_blocks` is called

crates/indexer/src/lib.rs
capossele: Added a simple exponential backoff

crates/indexer/src/db.rs
capossele: Added with [2b0307a](https://github.com/boundless-xyz/boundless/pull/560/commits/2b0307a3c3ad61a430f820f6e157d8133d8fa049)

capossele: At the end that solution was not good enough, so I've reverted to just ignore all the potential conflicts

willpote (approved):

PR #562, opened 2025-04-25
BM-824: Create revision 2 EFS for prover

nategraf:
Rename EFS file system with `-rev2` tag to cause the Pulumi to recreate it in the code pipeline.


willpote (approved):

nategraf:
Looks like our examples test is having some flakiness :(

PR #580, opened 2025-04-29
BM-840: fix logic with estimation log and warn if broker behind

austinabell:
There was an underflow in the log if the estimation is before the current time. This seems like a reasonable place to warn, as it indicates that the broker may overlock orders. Possibly could be noisy if tuned tightly, but airing on the side of caution. Happy to only warn if above a margin of error also

### DIFF
--- a/crates/broker/src/order_monitor.rs
+++ b/crates/broker/src/order_monitor.rs
@@ -546,22 +546,25 @@ where
         // Apply peak khz limit if specified
         if peak_prove_khz.is_some() && !orders_truncated.is_empty() {
             let peak_prove_khz = peak_prove_khz.unwrap();
-            let mut commited_orders = self.db.get_committed_orders().await?;
-            let num_commited_orders = commited_orders.len();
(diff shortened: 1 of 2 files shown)


nategraf (approved):

nategraf:
This is a helpful change. It seems related as well to some underflow errors I noticed, one of which I mentioned in the Slack. I included db4a947ad58f0a0cd0e8cbb344398dcf81e9b522, which you can feel free to remove if you think it shouldn't be included here.

austinabell:
> This is a helpful change. It seems related as well to some underflow errors I noticed, one of which I mentioned in the Slack. I included [db4a947](https://github.com/boundless-xyz/boundless/commit/db4a947ad58f0a0cd0e8cbb344398dcf81e9b522), which you can feel free to remove if you think it shouldn't be included here.

Those all look good, I hadn't checked those fns, thanks!

willpote:
Thanks for fixing!

PR #584, opened 2025-04-30
Refactor usage of the word "ELF" and rename the "upload_image" function

nategraf:
We use the word "ELF" to refer to the guest program in quite a lot of places, including public documentation and APIs. I think this is something that makes sense to us, but is actually needless jargon when the work "program" is just as clear in most contexts. With `risc0-zkvm` v2, its also no longer accurate since the guest program files are actually a struct that contains two ELFs in a wrapper (the kernel and the user program).

We also use the word "image" and "upload_image" or "image URL" a decent amount. In many places, this would be more clear (and sometimes more accurate) to refer to this as the program as well. In this PR, I've renamed the `upload_image` function on the `StorageProvider` to `upload_program`. I am doing some major work now on the request builder, so I didn't rename the "image URL" but I intend to.


willpote (approved):
Fan of this!

Breaking so we'll have to remember to bump to a new minor when we release this

PR #589, opened 2025-04-30
Add .ignore file for ripgrep and silver-searcher (ag)

nategraf:


capossele (approved):

.ignore
```diff
@@ -0,0 +1 @@
+/lib/
```
capossele, line 1: Wouldn’t this affect the git sub modules though?

nategraf: Using `ag` and `rg` would not report results in the submodules, that's correct. I personally find myself always needing to add `--ignore lib` to my searches. If you think this is not the default people want though I'm happy to keep this change just local to my machine.

capossele: You’re right, should be fine

PR #593, opened 2025-05-01
BM-864: avoid uploading image if already exists

austinabell:
draft just because pointing at #592 to test both at same time. PR is ready to go now otherwise


crates/broker/src/provers/bonsai.rs
```diff
@@ -211,20 +227,24 @@ impl StatusPoller {
 
 #[async_trait]
 impl Prover for Bonsai {
-    async fn upload_input(&self, input: Vec<u8>) -> Result<String, ProverError> {
-        retry::<String, ProverError, _, _>(
+    async fn has_image(&self, image_id: &str) -> Result<bool, ProverError> {
+        let status = retry::<bool, ProverError, _, _>(
```
willpote, line 231: This could use `self.retry` also no?

willpote (approved):

PR #595, opened 2025-05-01
BM-843: Add submitRootAndPriceAndFulfill to contract

capossele:
- Add submitRootAndPriceAndFulfill to contract
- Drop all single fulfillments methods to decrease contract deployed size
- Adds a Fulfill structure that can be used to send various fulfillment types with a build pattern
- Add configurable withdrawal after fulfilling

Closes BM-846


willpote:
Haven't had a chance to review yet, but can we keep breaking changes on a separate branch? Want to keep the broker that deploys to staging/prod stable so we can release it

contracts/src/BoundlessMarket.sol
```diff
@@ -228,48 +228,6 @@ contract BoundlessMarket is
...
-        // NOTE: Assessor guest ensures that the number of selectors <= the number of request digests in the journal.
-        if (assessorReceipt.selectors.length > 0 && assessorReceipt.selectors[0].value != bytes4(fill.seal[0:4])) {
-            revert SelectorMismatch(assessorReceipt.selectors[0].value, bytes4(fill.seal[0:4]));
-        }
-        // Verification of the assessor seal does not need to comply with DEFAULT_MAX_GAS_FOR_VERIFY.
-        VERIFIER.verify(assessorReceipt.seal, ASSESSOR_ID, assessorJournalDigest);
-    }
-
     /// @inheritdoc IBoundlessMarket
     function verifyBatchDelivery(Fulfillment[] calldata fills, AssessorReceipt calldata assessorReceipt) public view {
```
willpote, line 232: We can drop "batch" from this also

crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -1381,6 +1392,132 @@ pub fn decode_calldata(data: &Bytes) -> Result<(Vec<Fulfillment>, AssessorReceip
...
+    pub price: Option<Price>,
+    /// The fulfillments to be submitted
+    pub fulfillments: Vec<Fulfillment>,
+    /// The assessor receipt
+    pub assessor_receipt: AssessorReceipt,
+    /// Whether to withdraw the fee
+    pub withdraw: bool,
+}
+
+impl<P: Provider> Fulfill<P> {
```
willpote, line 1432: I like this, but I think its important we have consistency across the SDK with the work @nategraf is doing, so definitely sync with him!

Also I think something like: FulfillmentClient, FulfillmentHelper, FulfillmentBuilder would be better

willpote (approved):
Approving, though pls merge into a `next` branch or similar, and sync with Victor on the SDK design!

crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -1381,6 +1392,133 @@ pub fn decode_calldata(data: &Bytes) -> Result<(Vec<Fulfillment>, AssessorReceip
...
+    /// The list of client signatures for the requests.
+    pub client_sigs: Vec<Bytes>,
+}
+
+#[derive(Clone)]
+#[non_exhaustive]
+/// Builder for creating a fulfillment request.
+pub struct FulfillmentBuilder<P> {
+    /// The Boundless market service instance.
+    pub boundless_market: BoundlessMarketService<P>,
```
nategraf, line 1420: I don't think the `FulfillmentBuilder` (i.e. `FulfillmentTx` if you like that name) should hold an instance of the market and have a provider. Instead, we can have a function on the market that accepts a `FulfillmentBuilder` and does the work to send it.

crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -1381,6 +1392,133 @@ pub fn decode_calldata(data: &Bytes) -> Result<(Vec<Fulfillment>, AssessorReceip
...
+    /// The list of requests to be priced.
+    pub requests: Vec<ProofRequest>,
+    /// The list of client signatures for the requests.
+    pub client_sigs: Vec<Bytes>,
+}
+
+#[derive(Clone)]
+#[non_exhaustive]
+/// Builder for creating a fulfillment request.
+pub struct FulfillmentBuilder<P> {
```
nategraf, line 1418: On the name, maybe we should call this something like `FulfillmentTx` or similar. I would call it `Fulfillment`, but that main is taken by the Solidity derived struct. It does follow a builder pattern, but it doesn't need to be called a "builder" because it does not help you create a `Fulfillment`, and has no `build` method.

crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -1381,6 +1392,133 @@ pub fn decode_calldata(data: &Bytes) -> Result<(Vec<Fulfillment>, AssessorReceip
...
+}
+
+#[derive(Debug, Clone)]
+/// Represents the parameters for pricing a request.
+pub struct Price {
+    /// The list of requests to be priced.
+    pub requests: Vec<ProofRequest>,
+    /// The list of client signatures for the requests.
+    pub client_sigs: Vec<Bytes>,
+}
```
nategraf, line 1413: I would avoid calling it price. I think the fact that this appears in the public API is a bit unfortunate and I would avoid it if we can. I would prefer to this as an "unlocked request" instead. I would also refactor this to be a single `UnlockedRequest { request: ProofRequest, client_signature: Bytes }` and then wrap that in `Vec<UnlockedRequest>` on the `FulfillmentBuilder`. We can easily unzip this struct when it comes to to build the calldata.

crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -1381,6 +1392,133 @@ pub fn decode_calldata(data: &Bytes) -> Result<(Vec<Fulfillment>, AssessorReceip
...
+
+    /// Sets the parameters for submitting a Merkle Root.
+    pub fn with_submit_root(self, verifier_address: Address, root: B256, seal: Bytes) -> Self {
+        Self { root: Some(Root { verifier_address, root, seal }), ..self }
+    }
+
+    /// Sets the parameters for pricing a request.
+    pub fn with_price(self, requests: Vec<ProofRequest>, client_sigs: Vec<Bytes>) -> Self {
+        Self { price: Some(Price { requests, client_sigs }), ..self }
+    }
```
nategraf, line 1458: Based on my comment above, I would recommend we call this `with_unlocked_request`.

nategraf (commented):
I looked at the `FulfillmentBuilder` and left some comments there. I think this is a very nice utility to make building the fulfillment transactions more clear and easier to maintain. You don't need to wait for me to review again to merge this.

willpote (approved):

PR #596, opened 2025-05-02
Fix `_fufillNeverLocked` edge case

capossele:
Fixes BM-843 by setting the `fulfilled` flag only after the payment.

It builds on top of #595 due to avoid conflicts after changes to the contract, as such it should be merged after that


contracts/src/BoundlessMarket.sol
```diff
@@ -525,6 +523,9 @@ contract BoundlessMarket is
             price = _applyMarketFee(price);
         }
         accounts[assessorProver].balance += price;
+
+        clientAccount.setRequestFulfilled(idx);
+        emit RequestFulfilled(id);
     }
```
willpote, line 529: This creates a bit of inconsistency between "NeverLocked" vs the "Locked" and "WasLocked" paths. In Locked and WasLocked it is still possible for a request to be marked as fulfilled but payment not yet sent. 

I think the desired behavior should be that for NeverLocked, the request still gets marked as fulfilled if there is InsufficientBalance, but someone can come later and claim payment even though its fulfilled. Unfortunately I can't see a way to do this without adding an SSTORE and some new data structure.. so I'm wondering if its even worth handling this edge case at all

nategraf: In both directions there are unfortunate edge cases. One that exists with the code on `main` is that someone can have a "fulfilled" request with an impossibly high fee, since no fee actually needs to be exchanged. This is not the case with the paths that involve a lock, because the fee must be transferred on lock otherwise the tx will revert. This would be a problem for an indexer that tries to track the price of fulfilled requests. That is in addition to the fact that the request can get into a state where it is marked as fulfilled and it is impossible for payment to ever be transferred with relation to that request.

In this implementation, the main edge case is that it may become impossible for the request to become marked as fulfilled, even if a valid proof can be delivered. Its also the case that we break an invariant that appears to otherwise hold that if the transaction does not revert, all requests with fulfillment in the transaction will be marked as fulfilled by the end. This could result in a `ProofDelivered` event firing twice for the same request ID and two distinct requests with distinct requirements, or a `ProofDelivered` event firing for a request ID without there ever being a `Fulfilled` event. It could also end up with a pathological case where the request becomes locked _after_ this point, which just gets quite confusing.

Overall, I think the second set of edge cases is worse (although the first gives its best shot), and we should maintain the behavior that is on `main` now (i.e. revert these changes). Sorry for not thinking this through in detail before.

Some things we should do:
1. Adjust the comment at the top of this function to reflect the existence of this kind of weird edge case around a request becoming fulfilled without payment being sent.
2. Document (and double check) the invariant that the first `ProofDelivered` event should coincide with a `Fulfilled` event in all cases.
3. Revert on `!context.valid` [1]. This is actually a separate bug I am just noticing now. The issue is that, in particular for smart contract signed requests, the signature check on a request happens in the `priceReqeust` function. The current implementation using `return` instead of `revert` creates an issue where a third party can cancel someone else request by crafting a request with the same ID and fulfilling it before any lock occurs. They can do so without the request ever being signed.

[1] https://github.com/boundless-xyz/boundless/pull/596/files#diff-c53abde009685fc5570c04d0869fa35ec1453a929da5350c95fc23ad8ee61d0dR507-R509

PR #598, opened 2025-05-02
BM-861: Log error codes when printing errors using Debug trait. Add missing market monitor alarms

willpote:
When logging an error using trait Debug's formatting using `{:?}` error codes were not being logged, since the Debug derive is not aware of them. 

This PR adds a custom Debug macro that includes error codes. 

It also changes the logging behavior so that backtraces are not printed unless the `RUST_BACKTRACE` environment variable is set to true. Imo the backtraces are noisy and not particularly useful for debugging.

It also adds a custom alarm for event polling RPC errors, and a missing alarm on unexpected errors in the market monitor


crates/broker/src/errors.rs
```diff
@@ -5,3 +5,31 @@
 pub trait CodedError: std::error::Error {
     fn code(&self) -> &str;
 }
+
+// Macro for implementing Debug for CodedError. Ensures the error code is included in the debug output.
+#[macro_export]
+macro_rules! impl_coded_debug {
+    ($name:ident) => {
```
nategraf, line 12: It's unclear to me that this needs to be a macro rather than an blanket impl of the trait. I can try our a change here.

nategraf (approved):

crates/broker/src/errors.rs
willpote: Still getting the hang of Rust's type system, but it didn't seem possible to do something like:

```
impl<T: CodedError> Debug for T {
    fn fmt(&self, f: &mut Formatter) -> fmt::Result { ... }
}
```

since a 'foreign' trait for generic type isn't allowed. And can't do `impl Debug for CodedError` since CodedError is a trait not a type

PR #605, opened 2025-05-06
BM-871: Always revert on invalid context

capossele:
Addressing @nategraf comments mentioned in #596 :

1. Adjust the comment at the top of this function to reflect the existence of this kind of weird edge case around a request becoming fulfilled without payment being sent.
2. Document (and double check) the invariant that the first ProofDelivered event should coincide with a Fulfilled event in all cases.
3. Revert on !context.valid. This is actually a separate bug I am just noticing now. The issue is that, in particular for smart contract signed requests, the signature check on a request happens in the priceReqeust function. The current implementation using return instead of revert creates an issue where a third party can cancel someone else request by crafting a request with the same ID and fulfilling it before any lock occurs. They can do so without the request ever being signed.


contracts/src/BoundlessMarket.sol
nategraf, line 1: L591 in this file is an inaccurate comment at this point.

contracts/src/IBoundlessMarket.sol
```diff
@@ -36,9 +36,9 @@ interface IBoundlessMarket {
     event RequestFulfilled(RequestId indexed requestId);
 
     /// @notice Event logged when a proof is delivered that satisfies the request's requirements.
-    /// @dev It is possible for this event to be logged multiple times for a single request. This
-    /// is usually logged as part of order fulfillment, however it can also be logged by a prover
-    /// sending the proof without payment.
+    /// @dev It is possible for this event to be logged multiple times for a single request. The
+    /// first event logged will always coincide with the `RequestFulfilled` event, independently of
+    /// the payment outcome.
```
nategraf, line 41: I would change the slightly.
```suggestion
    /// first event logged will always coincide with the `RequestFulfilled` event and the fulfilled flag on the request being set.
```
And added a comment to the `RequestFulfilled` event that reads.
```solidity
/// @dev It is possible to a request to be fulfilled, and this event to be emitted, without the payment being transferred. Payment ma still be transferred later if the conditions are met.
```

nategraf (approved):

nategraf:
I pushed fcd7ba37 with a couple of code comments, and verified that the invariant I describe currently holds.

capossele:
Closing in favour of #703

PR #623, opened 2025-05-10
BM-901: short circuit if prover has order image already

austinabell:
Similar to https://github.com/boundless-xyz/boundless/pull/593 but for orders


capossele (approved):

nategraf:
@capossele and @austinabell , I added some log lines in https://github.com/boundless-xyz/boundless/pull/623/commits/3c9ccfd05606323412572cb8d210dbea5b2f0416 to help us observe when a fetch occurs vs when the cache is used.

PR #628, opened 2025-05-13
BM-895: New order generator

capossele:
- Modify the order generator to use a loop based on desired cycles
- Modify the infra to run 2 order generators, offchain and onchain
- Drop zeth generator as is broken and we'll switch to Kailua


infra/order-generator/index.ts
```diff
@@ -29,80 +31,129 @@ export = () => {
...
   
-  const simpleConfig = new pulumi.Config("order-generator-simple");
-  const simplePrivateKey = isDev ? pulumi.output(getEnvVar("SIMPLE_PRIVATE_KEY")) : simpleConfig.requireSecret('PRIVATE_KEY');
-  const simpleInterval = simpleConfig.require('INTERVAL');
-  const simpleLockStake = simpleConfig.require('LOCK_STAKE');
-  const simpleRampUp = simpleConfig.require('RAMP_UP');
-  const simpleMinPricePerMCycle = simpleConfig.require('MIN_PRICE_PER_MCYCLE');
-  const simpleMaxPricePerMCycle = simpleConfig.require('MAX_PRICE_PER_MCYCLE');
+  const imageName = getServiceNameV1(stackName, `order-generator`);
+  const repo = new awsx.ecr.Repository(`${imageName}-repo`, {
```
capossele, line 42: I've moved this out of the component so that we build and maintain only one image instead of one per component

willpote (approved):
🔥

PR #642, opened 2025-05-15
BM-924: Cap block range size on the indexer

capossele:
should fix:

```
2025-05-15T16:16:16.323327Z  WARN boundless_indexer: Failed to process blocks from 8038641 to 8333060: EventQueryError(TransportError(ErrorResp(ErrorPayload { code: -32602, message: "Log response size exceeded. You can make eth_getLogs requests with up to a 500 block range and no limit on the response size, or you can request any block range with a cap of 10K logs in the response. Based on your parameters and the response size limit, this block range should work: [0x7aa8f1, 0x7d5efc]", data: None }))), attempt number 11, retrying in 120s
```

### DIFF
--- a/crates/indexer/src/lib.rs
+++ b/crates/indexer/src/lib.rs
@@ -38,6 +38,8 @@ use url::Url;
 mod db;
 pub mod test_utils;
 
+const MAX_BATCH_SIZE: u64 = 500;
+
 type ProviderWallet = FillProvider<
     JoinFill<
         JoinFill<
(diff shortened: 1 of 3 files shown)


crates/indexer/src/lib.rs
```diff
@@ -138,12 +140,15 @@ where
...
                     }
 
-                    tracing::info!("Processing blocks from {} to {}", from_block, to_block);
+                    // cap to at most 500 blocks per batch
+                    let batch_end = min(to_block, from_block.saturating_add(MAX_BATCH_SIZE));
+
+                    tracing::info!("Processing blocks from {} to {}", from_block, batch_end);
 
-                    match self.process_blocks(from_block, to_block).await {
+                    match self.process_blocks(from_block, batch_end).await {
```
austinabell, line 148: This seems like it doesn't change the range at which the checkpoint is updated in `update_last_processed_block`, so this would just miss blocks past the batch size? Shouldn't this logic exist in `process_blocks`?

austinabell (approved):

PR #660, opened 2025-05-20
Support partial fulfillment

capossele:
This PR adds support for partial fulfillment of a batch without the need to recompute the assessor receipt.
The idea is to filter the fills based on the seal length. The seal is the only information not included as part of the leaf in the assessor tree.
It's not the cleanest solution so opening as a draft with only the contract changes to gather initial feedback.

Note: it builds on #595 

Closes BM-436


capossele:
> This would definitely be a large improvement, but still leaves some risk:
> 
> 1/ Prover checks if request has expired, hasn't expired yet, includes seal 2/ TX takes a few blocks to confirm 3/ Request has now expired
> 
> Its really just `InvalidRequestFulfillment` and `RequestIsExpiredOrNotPriced` we have to worry about, so wondering how bad it would be to return/catch these errors instead of reverting on them

Yeah this approach is kind of an hack, true. However catching anything regarding an invalid context would be the opposite of what we want to do with https://github.com/boundless-xyz/boundless/pull/605

capossele:
@willpote how about like this? [c1559f6](https://github.com/boundless-xyz/boundless/pull/660/commits/c1559f6efef41917799ca38639b96e5ff5720ae1)

If this way feels better, we can drop the first solution, or keep them both

contracts/src/BoundlessMarket.sol
```diff
@@ -312,14 +322,37 @@ contract BoundlessMarket is
...
+                            RequestIsFulfilled.selector, RequestId.unwrap(fills[i].id), fills[i].requestDigest
+                        )
+                    )
+                    || keccak256(paymentError[i])
+                        == keccak256(
+                            abi.encodeWithSelector(RequestIsExpiredOrNotPriced.selector, RequestId.unwrap(fills[i].id))
+                        )
+            ) {
+                continue;
+            }
```
willpote, line 345: I think if the error is request is fulfilled, we would still want to emit `ProofDelivered` no? Thats the current behavior (multiple `ProofDelivered` events, only ever one `Fulfilled` event)

The gas increase is also quite unfortunate. We should also be able to save the gas of re-encoding and hashing by just looking at the selector in assembly, something like:

```
bytes4 requestIsExpiredOrNotPricedSelector = RequestIsExpiredOrNotPriced.selector;
bool isExpired = false;
assembly {
    // Skip the slot that stores the length of the bytes array
    let errorSelector := mload(add(curPaymentError, 32))
    // Mask to keep only the first 4 bytes (selector)
    errorSelector := and(errorSelector, SELECTOR_MASK)
    // Compare with RequestIsFulfilled selector
    isExpiredOrNotPriced := eq(errorSelector, requestIsExpiredOrNotPricedSelector)
}

if (!isExpiredOrNotPriced) {
    emit ProofDelivered(fills[i].id);
}
```

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1571,6 +1575,87 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+
+        // Advance the chain ahead to simulate the first request timeout.
+        vm.warp(request.offer.deadline() + 1);
+
+        (Fulfillment[] memory fills, AssessorReceipt memory assessorReceipt) =
+            createFillsAndSubmitRoot(requests, journals, testProverAddress);
+
+        bytes memory error = abi.encodeWithSelector(IBoundlessMarket.RequestIsExpiredOrNotPriced.selector, request.id);
+        vm.expectEmit(true, true, true, true);
+        emit IBoundlessMarket.PaymentRequirementsFailed(error);
```
willpote, line 1611: We should rename this error now it can also cover Fulfillment errors

contracts/src/BoundlessMarket.sol
```diff
@@ -312,14 +322,37 @@ contract BoundlessMarket is
         // batch update to storage. However, updating the same storage slot twice only costs 100 gas, so
         // this savings is marginal, and will be outweighed by complicated memory management if not careful.
         for (uint256 i = 0; i < fills.length; i++) {
+            // Skip fulfillment if the prover did not provide a seal.
+            // This is used for partial fulfillments.
+            if (fills[i].seal.length < 4) {
+                continue;
+            }
             paymentError[i] = _fulfillAndPay(fills[i], assessorReceipt.prover);
```
willpote, line 330: Rename here too

willpote (commented):
Overall I feel good with this approach, and we can always remove/modify before mainnet. Risk seems pretty small to me of adding this to the contract.

I'm ok to drop the seal part, but don't feel strongly. It seems the only benefit to it would be the prover would be able to skip sending the seal, which typically would be small anyway, but they still have to send the journal etc. so doesn't seem like a big gain.

Curious how @nategraf feels about this approach

contracts/src/BoundlessMarket.sol
capossele: That's because I'm dumb, I meant to use `InvalidRequestFulfillment`. 
Yeah the optimization is a good idea, I'll also clean up a bit

willpote (approved):
Approving, though think we should get one more pair of eyes on before merge cc @nategraf

capossele:
Note that is not targeting main though

contracts/src/BoundlessMarket.sol
```diff
@@ -292,27 +292,44 @@ contract BoundlessMarket is
...
+                            InvalidRequestFulfillment.selector, RequestId.unwrap(fills[i].id), fills[i].requestDigest
+                        )
+                    )
+                    || keccak256(fulfillmentError[i])
+                        == keccak256(
+                            abi.encodeWithSelector(RequestIsExpiredOrNotPriced.selector, RequestId.unwrap(fills[i].id))
+                        )
+            ) {
+                continue;
+            }
```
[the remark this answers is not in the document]
capossele: so another option would be to add an additional `deadline` field to the Fulfillment struct; that field is completely controlled by the broker, and the contract can use it to compare it with the current timestamp and skip if it is greater.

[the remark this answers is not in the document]
capossele: Here's the alternative approach: https://github.com/boundless-xyz/boundless/pull/680

capossele:
Closing in favor of #703

PR #661, opened 2025-05-20
Explore multi-verifier support

capossele:
Explores ideas for adding support for fulfilling requests with:
- with or without journal;
- with a custom selector and verifier address

Related to BM-858


contracts/src/types/Requirements.sol
```diff
@@ -12,12 +12,19 @@ struct Requirements {
     bytes32 imageId;
     Callback callback;
     Predicate predicate;
+    /// @notice The selector for the seal.
+    /// @dev If the selector is set to zero, the seal is not required.
     bytes4 selector;
+    /// @notice The address of the IRiscZeroVerifier contract.
+    /// @dev If the address is set to zero, the default IRiscZeroVerifier is used.
+    address verifier;
```
capossele, line 20: This is added if we want to give to the requestor full freedom to decide how its application seal should be verified.
Otherwise deploying via multi-router could also work

PR #669, opened 2025-05-21
BM-956: add rust check to docs CI changes and resolve issues

austinabell:
Not sure if this was what was intended or if it works, just trying to get CI to pass for #540


sashaaldrick:
This looks correct to me, thanks for catching that Steel 2.0 history example and fixing CI here so that this won't happen again.

capossele (approved):

PR #671, opened 2025-05-21
BM-958: add just installer to prover setup script

sashaaldrick:
I noticed when running a prover on a fresh machine, that we recommend you run the setup script, and then we ask to run a lot of different `just` commands. If you're running this on a fresh machine, the script handily installs Rust, and Docker and Nvidia etc. dependencies for you, but not the `just` command runner.

It felt like it was missing, so I added it to the `setup.sh` script and verified it worked on an Ubuntu 22.04 install.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -336,7 +336,7 @@ configure_docker_nvidia
 install_rust
 
 # Install Just
-install_rust
+install_just
 
 # Install CUDA Toolkit
 install_cuda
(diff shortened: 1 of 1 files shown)


scripts/setup.sh
```diff
@@ -318,6 +335,9 @@ configure_docker_nvidia
 # Install Rust
 install_rust
 
+# Install Just
+install_rust
```
austinabell, line 339:
```suggestion
install_just
```

sashaaldrick: god damn it, thank you @austinabell  🤦‍♂️

austinabell (approved):

PR #680, opened 2025-05-22
Support partial fulfillment (version 2)

capossele:
As an alternative to #660 here we add an additional field `deadline` to the `Fulfillment` structure, that can be used either with the lockTimeout or the Timeout

Closes BM-436


contracts/src/types/Fulfillment.sol
```diff
@@ -14,6 +14,8 @@ struct Fulfillment {
     RequestId id;
     /// @notice EIP-712 digest of request struct.
     bytes32 requestDigest;
+    /// @notice The timestamp when the request is considered expired.
```
willpote, line 17:
```
/// @notice The timestamp when the request should be considered expired, and should be skipped during fulfillment.
/// @dev Generally should be set to either lockDeadline or deadline. Enables a batch to be partially fulfilled in the case where one request in the batch has expired
```

willpote (commented):
This does seem like a better approach!

PR #684, opened 2025-05-23
BM-979: Refactor nonce management

capossele:
Previous PRs have changed how txs, especially locking txs were concurrently sent to the network. This created a series of issues as when sending multiple Ethereum transactions concurrently, a race condition can arise leading to incorrect nonce assignments and failed transactions (nonce too low errors). This was mitigated by the use of a cached nonce manager. However, that also created a different issue: transactions were getting stuck in the mempool or experienced slow confirmation; Alloy’s default gas estimation doesn't dynamically react to the number of transactions already pending.

This PR drops (for the time being) the concurrent submission of lock requests, but introduces a `MutexLayer` to guarantee a consistent nonce managament by broadcasting only one tx at a time while keeping the access to the Provider send_transaction_row locked. Additionally, a dynamic gas filler adapts the gas price based on the number of pending txs.


crates/boundless-market/src/nonce_layer.rs
```diff
@@ -0,0 +1,136 @@
...
+            );
+        }
+
+        let tx = self.inner.fill(request).await.unwrap();
+
+        let builder = match tx {
+            SendableTx::Builder(builder) => builder,
+            _ => {
+                panic!("should not be called test");
+                // return Ok(tx);
```
austinabell, line 108: Accidentally left this in, feel free to keep the panic or just use the raw transaction as is -- but the former should only happen

crates/boundless-market/src/dynamic_gas_filler.rs
```diff
@@ -0,0 +1,146 @@
...
+                        "DynamicGasFiller: Adjusted gas limit: {}, gas price: {}",
+                        adjusted_gas_limit,
+                        adjusted_gas_price
+                    );
+                }
+                GasFillable::Eip1559 { gas_limit, estimate } => {
+                    let adjusted_gas_limit =
+                        (gas_limit as f64 * (1.0 + self.gas_limit_factor)).ceil() as u64;
+                    let adjusted_max_fee =
+                        (estimate.max_fee_per_gas as f64 * params.multiplier) as u128;
```
nategraf, line 127: I think we should not adjust the max fee, or adjust it just enough to accommodate the increase to the priority fee. As this is increase here will be quite large compared to the increase to the priority fee.
```suggestion
```

nategraf (approved):

crates/boundless-market/src/nonce_layer.rs
```diff
@@ -0,0 +1,136 @@
...
+            <EthereumWallet as NetworkWallet<Ethereum>>::default_signer_address(&self.wallet)
+        };
+        request.set_from(from_address);
+
+        // Get semaphore for this account and acquire permit
+        let semaphore = self.get_account_semaphore(from_address).await;
+        let _permit = semaphore.acquire().await.unwrap();
+
+        // Fetch the pending nonce if not already set
+        if request.nonce.is_none() {
```
capossele, line 92: @austinabell are we sure that the default filler are not going to fill this nonce up and that happen without semaphore control?

capossele: nvm, the fill happens later, all good!

austinabell: yeah this is only set if the caller explicitly sets a nonce, but we don't do that here. 

Thinking about this though, we should probably disable recommended fillers to at least simplify the flow I think?

PR #695, opened 2025-05-26
BM-991: Adjust top levels for staging, prover errors to known, lower order generator timeouts, tx timeouts in og

willpote:
Assorted fixes:
* Provers errors, which can be caused by transient Bonsai 500s, moved to known errors and with a higher threshold
* Lowering order generator timeouts. Some orders are send with timeouts of 12 hours+. Also makes the timeout after lock timeout higher
* Fix bug to propagate tx timeout settings to the OG code correctly
* Separated broker.toml for the bento/bonsai prover so we can config them differently


infra/prover/tomls/bento/broker-prod-11155111.toml
```diff
@@ -0,0 +1,37 @@
+[market]
+mcycle_price = "0.00001"
+mcycle_price_stake_token = "0" # will fill any lock-expired orders regardless of cost
+peak_prove_khz = 1000
```
willpote, line 4: Dropped these from 5k to 1k for now, pending better analysis

austinabell (approved):

PR #699, opened 2025-05-27
BM-994: Update docs to cater for new request building SDK functionality

sasha-computer:
after @nategraf's work on improving the request building functionality, this is a refactor of the docs specifically around "Request a Proof"

### DIFF
--- a/infra/prover/tomls/bonsai/broker-prod-84532.toml
+++ b/infra/prover/tomls/bonsai/broker-prod-84532.toml
@@ -37,4 +37,3 @@ single_txn_fulfill = true
 withdraw = true
 max_submission_attempts = 2
 # batch_poll_time_ms = 500
-withdraw = true


documentation/site/pages/developers/tutorials/request.mdx
```diff
@@ -75,17 +194,45 @@ let program_url = client.upload_program(program).await?;
...
-### 3. Upload the Input of Your Program to IPFS
+After which, you'd create a request with:
 
-:::tip[*Rather not have inputs publicly available?**]
-When working with trusted provers, you can store inputs in *Amazon S3* and restrict access with a bucket policy (plus an optional *KMS key policy* if you enable server-side encryption).
-Only provers whose IAM role appears in those policies can download the inputs.
+```rust
+let request = client.new_request()
+  .with_program_url(&program_url)
+  .with_input_url(&input_url);
```
nategraf, line 202: So, `with_input_url` does not exist right now. This could be added, or it could be accomplished with `.with_request_input(Url::parse("https://myguest.null/program")?)`

documentation/site/pages/developers/tutorials/request.mdx
```diff
@@ -101,45 +248,86 @@ let input_url = client.upload_input(&input_bytes).await?;
...
+
+// Submit the request directly
+let (request_id, expires_at) = client.submit_offchain(request).await?; // [!code hl] // [!code focus]
 # Ok(())
 # }
 ```
 </StripRustCodeComments>
 
-> Relevant links: [ExecutorEnvBuilder](https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.ExecutorEnvBuilder.html)
+### Offer Parameters
```
nategraf, line 326:
```suggestion
### Offer
```

nategraf, line 326: So, in addition to `OfferParams` that can be set when building the client, there are also config options when building the `Client`. In particular, if someone wants to use the cycle count to decide the price, then they can do so by calling `Client::builder().config_offer_layer(|config| /* set options here */)`. Could you also show this? Sorry that it's a bit confusing that there are two ways to do it. The main difference is that you use the methods on the client builder when you want to adjust the logic the builder uses, you use `OfferParams` when you want to override the builder.

https://github.com/boundless-xyz/boundless/blob/1f2019496da6745363dee9f59d178a8dc151eede/crates/boundless-market/src/client.rs#L286-L305

sashaaldrick: @nategraf could you please take a look here and let me know if this resolves your comment here:

https://boundless-documentation-g273n87e4-risczero.vercel.app/developers/tutorials/request#offer

documentation/site/pages/developers/tutorials/request.mdx
```diff
@@ -45,21 +82,128 @@ let client = Client::builder()
...
-This is done by uploading the program to a public HTTP server.
-The `boundless-market` provides `StorageProvider` implementations to help upload these binaries.
-Built-in providers are listed in the [`BuiltinStorageProvider`](https://docs.rs/boundless-market/latest/boundless_market/storage/enum.BuiltinStorageProvider.html).
+```bash
+export S3_ACCESS_KEY="abcdef..."
+export S3_SECRET_KEY="abcdef..."
+export S3_BUCKET="bucket-name..."
+export S3_URL="https://bucket-url..."
+export AWS_REGION="us-east-1"
+Once these are set, this will automatically use the specified [AWS S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/creating-buckets-s3.html) for storage of programs and inputs.
```
willpote, line 195: This is being rendered in the code block

documentation/site/pages/developers/tutorials/request.mdx
```diff
@@ -75,17 +219,57 @@ let program_url = client.upload_program(program).await?;
...
+```
+</StripRustCodeComments>
+
+If you already have the `program_url`, you do not need to upload the program again; you can simply use `with_program_url` with a hard-coded URL.
 
-See [Sensitive Inputs](/developers/tutorials/sensitive-inputs) for more information.
+#### Automagically
+
+If you are working in a monorepo, you can take advantage of [risc0-build](https://docs.rs/risc0-build/latest/risc0_build/) which automatically builds and exposes the ELF for the guest. The counter example uses this method:
+
```
willpote, line 244: I'm a bit confused on this; 1/ are we saying `with_program` will only work if you have manually made risc0-build available in your project? 2/ if so, are you sure it has to be a monorepo? and not just adding it as a dependency in your Cargo.toml?

sasha-computer: I've made this clarification: 

```
If you are working in a monorepo **(i.e. your zkVM host/guest is in the same repo)**, you can take advantage of [risc0-build] ...:
```

Yes, `with_program` requires the program bytes directly, i.e. the ELF:

https://github.com/boundless-xyz/boundless/blob/1f2019496da6745363dee9f59d178a8dc151eede/crates/boundless-market/src/request_builder/mod.rs#L324-L327

`risc0-build` uses `methods/build.rs` to expose the ELF as `<GUEST_NAME>_ELF` automatically so that you can use the ELF bytes within a rust file within the same repo directly; the counter example does this:

https://github.com/boundless-xyz/boundless/blob/1f2019496da6745363dee9f59d178a8dc151eede/examples/counter/apps/src/main.rs#L19

Otherwise, `with_program_url` is available, which would require you to run `upload_program()` beforehand or hardcode a URL directly which is explained right above in the `#### Manually` section:

https://github.com/boundless-xyz/boundless/blob/d7f73291ee346e1e18afe735c51fc54831659269/documentation/site/pages/developers/tutorials/request.mdx?plain=1#L207-L240

willpote: Thanks for the clarification!

documentation/site/pages/developers/tutorials/request.mdx
```diff
@@ -1,42 +1,79 @@
...
 
-1. [Create a boundless client.](#1-create-a-boundless-client)
-2. [Upload the program executable to IPFS.](#2-upload-program-executable-to-ipfs)
-3. [Upload the input of your program to IPFS.](#3-upload-the-input-of-your-program-to-ipfs)
-4. [Execute the program to estimate the total number of cycles, and to get the journal.](#4-execute-the-program-to-estimate-cycles-and-to-get-the-journal)
-5. [Create a proving request with URLs to the executable and inputs, the journal hash, the imageID, and the offer details (such as price range and timeout length).](#5-create-a-proof-request)
-6. [Submit the request to the market.](#6-submit-a-request)
-7. [Wait for the request to be fulfilled, returning the journal and the seal.](#7-wait-for-the-request-to-be-fulfilled)
+:::note[Source Code]
+The *Sending a Request* section uses the counter example as a template, its source code can be found at: [apps/src/main.rs](https://github.com/boundless-xyz/boundless/blob/main/examples/counter/apps/src/main.rs)
```
nategraf: This link will get stale

documentation/site/pages/developers/tutorials/request.mdx
```diff
@@ -1,42 +1,79 @@
...
+```bash
+export RPC_URL="https://..."
+export PRIVATE_KEY="abcdef..."
+```
+
+:::tip[Alchemy]
+We recommend using Alchemy for your RPC URL during testing; their free tier is more than enough to test requesting a proof.
+:::
+
+#### Storage Provider
```
nategraf, line 42: It would be good to make clear in this section the option of pre-uploading your program, to avoid the need to use a Pinata API key to do it at runtime. If they do, they will still need to post it to some public URL, and that could be Pinata, but it also could be any service or server of their choosing.

The text from "No Storage Provider" is good. Maybe it can be duplicated or referenced

nategraf (approved):
Looks good! I left some comments, and also made some suggested changes while reviewing it. That PR also includes some TODOs for things to address, not necessarily in this PR https://github.com/boundless-xyz/boundless/pull/786

Feel free to merge this PR (#699) when you are happy with it

documentation/site/pages/developers/tutorials/request.mdx
sasha-computer: I've pointed it now to the general examples/counter folder instead to make it a bit more fullproof

documentation/site/pages/developers/tutorials/request.mdx
sasha-computer: I have added a tip about this to the top of the **Storage Provider** section; thanks!

PR #702, opened 2025-05-28
BM-999: Reaper actor

austinabell:
To avoid edge cases where some committed orders may not be cleared from the DB.

Can remove https://github.com/boundless-xyz/boundless/commit/2bc6ffe71dae44a1051a8d5a7b35deaecb10c06e from the changes, but with the reaper failing expired tasks, this was the most glaring as what would cause a race condition because no expiry checks are being made.

This is by no means perfect, but a stop-gap until we have a better way to handle expiries without having issues with race conditions from the modifications from each actor.

### DIFF
--- a/crates/broker/src/reaper.rs
+++ b/crates/broker/src/reaper.rs
@@ -57,7 +57,7 @@ impl ReaperTask {
         let expired_orders = self.db.get_expired_committed_orders(grace_period.into()).await?;
 
         if !expired_orders.is_empty() {
-            info!("Found {} expired committed orders", expired_orders.len());
+            info!("[B-REAP-100] Found {} expired committed orders", expired_orders.len());
 
             for order in expired_orders {
                 let order_id = order.id();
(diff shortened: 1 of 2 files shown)


crates/broker/src/reaper.rs
```diff
@@ -0,0 +1,290 @@
...
+    async fn check_expired_orders(&self) -> Result<(), ReaperError> {
+        let grace_period = {
+            let config = self.config.lock_all()?;
+            config.prover.reaper_grace_period_secs
+        };
+
+        let expired_orders = self.db.get_expired_committed_orders(grace_period.into()).await?;
+
+        if !expired_orders.is_empty() {
+            info!("Found {} expired committed orders", expired_orders.len());
```
willpote: Can you add an error code + SEV2 alarm for this? Although we recover, would want us to be alerted so we can investigate why things got stuck

willpote (approved):

crates/broker/src/reaper.rs
austinabell: I added a code and an alert for it, but didn't add an error variant and into the code implementation since that seemed cosmetic. Are you partial to keeping the same pattern as others?

willpote: Thats fine!

willpote (approved):

PR #706, opened 2025-05-28
BM-1003: added the with_input_url method to RequestParams in the request builder API

sasha-computer:
added `with_input_url` to match the schema of `with_program_url`, otherwise we'd have to use something like:

`.with_request_input(Url::parse("https://myguest.null/program")?)` which is a little boring imo

now you can use:

`client.new_request().with_program_url(program_url)?.with_input_url(input_url)?`

### DIFF
--- a/examples/counter/apps/src/main.rs
+++ b/examples/counter/apps/src/main.rs
@@ -47,9 +47,6 @@ struct Args {
     /// URL where provers can download the program to be proven.
     #[clap(long, env)]
     program_url: Option<Url>,
-    /// URL where provers can download the input data.
-    #[clap(long, env)]
-    input_url: Option<Url>,
     /// Configuration for the StorageProvider to use for uploading programs and inputs.
     #[clap(flatten, next_help_heading = "Storage Provider")]
     storage_config: StorageProviderConfig,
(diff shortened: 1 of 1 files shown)


examples/counter/apps/src/main.rs
```diff
@@ -81,10 +87,21 @@ async fn run(args: Args) -> Result<()> {
...
-    // We use a timestamp as input to the ECHO guest code as the Counter contract
-    // accepts only unique proofs. Using the same input twice would result in the same proof.
-    let echo_message = format!("{:?}", SystemTime::now());
-    let request = client.new_request().with_program(ECHO_ELF).with_stdin(echo_message.as_bytes());
+    // Build the request based on whether URLs are provided
+    let request = match (args.program_url.clone(), args.input_url.clone()) {
+        (Some(program_url), Some(input_url)) => {
+            // Use the provided URLs
+            client.new_request().with_program_url(program_url)?.with_input_url(input_url)?
+        }
```
sashaaldrick, line 95: This part is not super necessary, I was using it to test the new `with_input_url`. I'm happy to revert the example back here.

nategraf (approved):

PR #716, opened 2025-05-29
BM-1027: add safe upgrade steps

austinabell:
replaces broker.toml with a template that will create the `broker.toml` on `just broker` if not exists.

Avoids git operations dropping the bind mount of the file on docker compose and avoids managing git diffs through updates on config.


austinabell:
https://boundless-documentation-4riowkifh-risczero.vercel.app/provers/broker#safe-upgrade-steps

willpote (approved):

documentation/site/pages/provers/broker.mdx
```diff
@@ -162,6 +162,77 @@ cargo build --bin broker --release
 just broker down
```
nategraf, line 152: Should this step still list `just broker down` as the way to turn down your broker. It seems like this should warn provers and tell them to use the procedure just below when shutting down if actively proving on the network, to avoid losing stake.

documentation/site/pages/provers/broker.mdx
```diff
@@ -162,6 +162,77 @@ cargo build --bin broker --release
...
+##### Reset the max_concurrent_proofs configuration
+
+Once the broker is running successfully with the new version, edit your `broker.toml` to restore normal operation:
+
+```toml [broker.toml]
+max_concurrent_proofs = <your_previous_value>
+# Example: max_concurrent_proofs = 4
+```
+
+::::
```
nategraf, line 228: Something is a little off here. When looking at the rendered version, everything below this point is nested under the "steps" indent

austinabell: That was on an old version. https://boundless-documentation-cx2gv3ne5-risczero.vercel.app/provers/broker#safe-upgrade-steps is the most recent link. Do you still see that there? If so, I don't see it

documentation/site/pages/provers/broker.mdx
austinabell: https://github.com/boundless-xyz/boundless/pull/716/commits/85206452eed82be0633e29a759044056a3ddc31d

Good suggestion!

PR #736, opened 2025-06-02
Avoid revert on partially lock-expired batch

nategraf:


nategraf:
As currently written, this is not quite right. It doesn't handle the case that the request is locked, but is fully expired.

contracts/snapshots/BoundlessMarketBasicTest.json
```diff
@@ -1,40 +1,40 @@
 {
   "ERC20 approve: required for depositStake": "45966",
-  "bytecode size implementation": "23783",
+  "bytecode size implementation": "24113",
```
nategraf: Getting dangerously close again 😅 . We'll have to do an optimization pass soon.

nategraf:
Note that with this PR, both `_fulfillAndPayLocked` and `_fulfillAndPayWasLocked` start with the same two checks. This could be moved out to the call site and may reduce bytecode size a tad. I didn't make this change because it is not required to fix the issue

contracts/snapshots/BoundlessMarketBasicTest.json
capossele: yeah I know, this limitation is getting very annoying!

capossele (approved):
It looks good overall!
I've suggested just a change in https://github.com/boundless-xyz/boundless/pull/743 but that is totally optional as explained in the PR description

contracts/src/BoundlessMarket.sol
```diff
@@ -359,42 +367,67 @@ contract BoundlessMarket is
...
+            }
+        } else if (locked && lock.requestDigest == fill.requestDigest) {
+            // Request was validated in lockRequest, check the expiration on the lock.
+            if (lock.deadline() < block.timestamp) {
+                paymentError = abi.encodeWithSelector(RequestIsExpired.selector, RequestId.unwrap(id));
+                emit PaymentRequirementsFailed(paymentError);
+                return (paymentError, true);
+            }
+        } else {
+            // Request is not validated by either price or lock step. We cannot determine the that
```
willpote:
```suggestion
            // Request is not validated by either price or lock step. We cannot determine that the
```

contracts/src/BoundlessMarket.sol
```diff
@@ -433,8 +460,8 @@ contract BoundlessMarket is
 
         // At this point the request has been fulfilled. The remaining logic determines whether
         // payment should be sent and to whom.
-        // While the request is locked, only the locker is eligible for payment.
-        if (lock.prover != assessorProver) {
+        // While the request is locked, only the locker is eligible for payment, and only for the request that was locked.
+        if (lock.prover != assessorProver || lock.requestDigest != fill.requestDigest) {
```
willpote, line 464: I believe we load already everything we need to check this on L400-406, so might be cheaper to check it there.

contracts/src/BoundlessMarket.sol
```diff
@@ -359,42 +367,67 @@ contract BoundlessMarket is
...
+        if (context.valid) {
+            // Request has been validated in priceRequest, check the reported expiration.
+            if (context.expired) {
+                paymentError = abi.encodeWithSelector(RequestIsExpired.selector, RequestId.unwrap(id));
+                emit PaymentRequirementsFailed(paymentError);
+                return (paymentError, true);
+            }
+        } else if (locked && lock.requestDigest == fill.requestDigest) {
+            // Request was validated in lockRequest, check the expiration on the lock.
+            if (lock.deadline() < block.timestamp) {
```
willpote: `lock.deadline()` is actually the deadline of the entire request, not the expiration of the lock. Might just be a comment issue, as I do think the intention is to check that the request has fully expired here (?)

contracts/src/BoundlessMarket.sol
```diff
@@ -359,42 +367,67 @@ contract BoundlessMarket is
...
+                paymentError = abi.encodeWithSelector(RequestIsExpired.selector, RequestId.unwrap(id));
+                emit PaymentRequirementsFailed(paymentError);
+                return (paymentError, true);
+            }
+        } else {
```
willpote: One thing I'm thinking through is that it looks like its possible to price a request while its still locked, which would cause it to skip L400-406. It seems we are covered, but just wanted to flag, and if true might be worth adding a comment

contracts/src/BoundlessMarket.sol
nategraf: Yes, but at that point we have not yet checked that the lock is not expired. Additionally, returning early from line 400 would prevent `ProofDelivered` from firing.

contracts/src/BoundlessMarket.sol
nategraf: Ah yes, My intention is to check for full expiration be0ab59c4baf05de62b6d2462cadaef1c063bee5.

contracts/src/BoundlessMarket.sol
nategraf: That's right. Payment will be declined later if the request does not match the one currently locked. I tried to think of a clarifying comment here, but I'm not sure what to say that would not potentially confuse someone with fresh eyes (i.e. who does not have a mental model based on how it works on `main` right now). Ultimately I think it's cleaner to divide the logic like this, where we check that the request is authentic here (i.e. it was signed by the client) and then later check that it meets the other requirements for payment.

Here is where the case that the given request does not match the locked one is handled.

https://github.com/boundless-xyz/boundless/blob/80fab42ab1d19a053efa2afafa2f8eecb77213e1/contracts/src/BoundlessMarket.sol#L461-L466

PR #738, opened 2025-06-03
Increase stake balance warnings for provers, improve logs in provers, support slasher for log debugging

willpote:
* Order stream requires 5 USDC to connect, but we get warned of low stake balance too late with the current config. Increased the threshold
* Ensure we log proof id whenever we start a proof
* Tidy up some noisy logs in order_monitor
* Fix application logs in Lambdas
* Added support for slasher to the log debugger


infra/prover/tomls/bonsai/broker-prod-11155111.toml
```diff
@@ -10,8 +10,8 @@ max_mcycle_limit = 8_000
 max_concurrent_locks = 3
 balance_warn_threshold = "5"
 balance_error_threshold = "1"
-stake_balance_warn_threshold = "5"
-stake_balance_error_threshold = "1"
+stake_balance_warn_threshold = "15"
+stake_balance_error_threshold = "10"
```
capossele, line 14: these might be too high though, as the USDC faucet can only issues 10 per hour, meaning it might be annoying for the brokers at the beginning at least

PR #739, opened 2025-06-03
Add allowlist for large proofs

willpote:
Temporary way for us to support huge proofs without DDOS risk


austinabell (approved):

crates/broker/src/order_picker.rs
```diff
@@ -1472,4 +1489,29 @@ mod tests {
...
+        {
+            config.load_write().unwrap().market.mcycle_price = "0.0000001".into();
+            config.load_write().unwrap().market.max_mcycle_limit = Some(1000);
+        }
+        let ctx = TestCtxBuilder::default().with_config(config).build().await;
+        
+        ctx.picker.config.load_write().as_mut().unwrap().market.allow_skip_mcycle_limit_addresses = Some(vec![ctx.provider.default_signer_address()]);
+        
+        let order = ctx.generate_next_order(Default::default()).await;
+        let request_id = order.request.id;
```
nategraf, line 1509: Can we add a case here to make sure that other addresses _do_ have the limit imposed, even when the list is non-empty?

nategraf (approved):

crates/broker/src/order_picker.rs
willpote: Added in e2c2c507095db102ac1169c9de9c52518b7927cb

PR #740, opened 2025-06-03
Re-add codes on failed to lock order

willpote:
Currently lock failed errors are treated as unexpected errors since the code is not printed. Re-adds the code

### DIFF
--- a/crates/broker/src/order_monitor.rs
+++ b/crates/broker/src/order_monitor.rs
@@ -518,12 +518,14 @@ where
                             match err {
                                 OrderMonitorErr::UnexpectedError(inner) => {
                                     tracing::error!(
-                                        "Failed to lock order: {order_id} - {inner:?}",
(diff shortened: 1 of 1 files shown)


austinabell (approved):

crates/broker/src/order_monitor.rs
```diff
@@ -518,12 +518,14 @@ where
...
-                                        "Failed to lock order: {order_id} - {inner:?}",
+                                        "Failed to lock order: {order_id} - {} - {inner:?}",
+                                        err.code()
                                     );
                                 }
                                 _ => {
                                     tracing::warn!(
-                                        "Soft failed to lock request: {order_id} - {err:?}",
+                                        "Soft failed to lock request: {order_id} - {} - {err:?}",
+                                        err.code()
```
austinabell, line 528: This one _seems_ unnecessary because the code should be in `err` debug though, no?

PR #743, opened 2025-06-04
use InvalidRequestFulfillment

capossele:
This suggestions reintroduce the use of `InvalidRequestFulfillment` to revert in case the `requestDisgest` does not match with what was locked. It's mostly for debugging purposes rather than functional, it does add an extra ~180 bytes to the deployment code size so it totally optional and just a suggestion.

I've also renamed `payLocked`, `payWasLocked` and moved their checks so that there is less code repetition, that saved ~300 bytes

### DIFF
--- a/contracts/src/BoundlessMarket.sol
+++ b/contracts/src/BoundlessMarket.sol
@@ -399,7 +399,7 @@ contract BoundlessMarket is
             }
         } else if (locked) {
             if (lock.requestDigest == fill.requestDigest) {
-                // Request was validated in lockRequest, check the expiration on the lock.
+                // Request was validated in lockRequest, check whether the request is fully expired.
                 if (lock.deadline() < block.timestamp) {
                     paymentError = abi.encodeWithSelector(RequestIsExpired.selector, RequestId.unwrap(id));
                     emit PaymentRequirementsFailed(paymentError);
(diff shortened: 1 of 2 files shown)


contracts/src/BoundlessMarket.sol
```diff
@@ -397,12 +397,20 @@ contract BoundlessMarket is
...
+                    paymentError = abi.encodeWithSelector(RequestIsExpired.selector, RequestId.unwrap(id));
+                    emit PaymentRequirementsFailed(paymentError);
+                    return (paymentError, true);
+                }
+            } else {
+                revert InvalidRequestFulfillment({
+                    requestId: id,
+                    provided: fill.requestDigest,
+                    locked: lock.requestDigest
+                });
```
nategraf, line 413: So I actually avoid reverting in this case intentionally. If there exists requests A and B with the same ID, and A is locked, it's ok for a prover (same or other) to `priceAndFulfilll` with B during the lock period, although they should not be paid for doing so.

capossele: You actually revert in that case, you just revert with `RequestIsNotLockedOrPriced ` rather than with what I'm suggesting. No?

capossele: To be more specific, this revert does not apply in case of pricing

nategraf: Oh, you are correct! My earlier reading was wrong. But I don't love adding an extra case for this. We can get the same value from transaction simulation if we every need to debug a request.

willpote (dismissed):

capossele:
Closing for now

PR #744, opened 2025-06-04
Add deployment address check to main CI workflow

sashaaldrick:


capossele:
In https://github.com/boundless-xyz/boundless/pull/745 I've suggested to use a python script instead, since this bash script relies a little bit on the file structure rather than its content. Also with a script we can also invoke it locally

deployments-check.py
```diff
@@ -0,0 +1,97 @@
...
+    todos = check_todos(docs_content)
+    if todos:
+        print("❌ Found TODO placeholders in documentation:")
+        for todo in todos:
+            print("  ", todo)
+        errors += len(todos)
+
+    networks = {
+        'ethereum-sepolia-prod': '### Ethereum Sepolia',
+        'base-mainnet': '### Base Mainnet',
```
capossele, line 52:
```suggestion
        'base-mainnet': '### Base Mainnet',
        'base-sepolia-prod': '### Base Sepolia'
```

deployments-check.py
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    rs_network_keys = {
+        'ethereum-sepolia-prod': 'SEPOLIA',
+        'base-mainnet': 'BASE'
```
capossele:
```suggestion
        'base-mainnet': 'BASE',
        'base-sepolia-prod': 'BASE'_SEPOLIA
```

capossele (commented):
Forgot one network

PR #752, opened 2025-06-04
BM-1108: change reaper grace period to 4 mins, add expiry check in aggregator

austinabell:
- Changes reaper grace period to 4 mins to avoid marking orders as expired while aggregator/submitter are working
  - The proving time for assessor and set builder can be variable based on proving setup so this doesn't guarantee, but no fatal consequences if an order marked as failed early
- Adds expiry check pre aggregation
  - Didn't seem clean to apply to submitter, but will double check before opening this PR


crates/broker/src/aggregator.rs
```diff
@@ -406,6 +406,56 @@ impl AggregatorService {
...
+    ) -> Result<Vec<AggregationOrder>, AggregatorErr> {
+        let mut valid_orders = Vec::with_capacity(orders.len());
+
+        for order in orders {
+            if order.expiration < current_time {
+                tracing::warn!(
+                    "Order {} has expired during aggregation, marking as failed",
+                    order.order_id
+                );
+                if let Err(err) =
```
willpote: Definitely hesitant to add more alerts, but I do worry that now when this happens the problem is essentially swallowed, and we won't know. And its pretty bad if this happens since you get slashed.

Thoughts on something like an alert where if it happens a few times an hour we get a SEV2?

willpote (approved):

crates/broker/src/aggregator.rs
austinabell: very reasonable, I set it to a low threshold (2) given this expiry timing is basically an edge case, so if it happens more than once, likely something wrong

PR #757, opened 2025-06-05
BM-1143: Add shell completions to boundless-cli

nategraf:
Ok, ok I know this isn't the highest priority work... but it only took 30 minutes to get working and think about how many seconds per day of fruitlessly pressing tab will be saved!

### DIFF
--- a/crates/boundless-cli/src/bin/boundless-cli.rs
+++ b/crates/boundless-cli/src/bin/boundless-cli.rs
@@ -57,7 +57,8 @@ use alloy::{
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use bonsai_sdk::non_blocking::Client as BonsaiClient;
 use boundless_cli::{convert_timestamp, DefaultProver, OrderFulfilled};
-use clap::{Args, Parser, Subcommand};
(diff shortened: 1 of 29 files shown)


crates/boundless-cli/src/bin/boundless-cli.rs
```diff
@@ -109,6 +110,17 @@ enum Command {
 
     /// Display configuration and environment variables
     Config {},
+
+    /// Print shell completions (e.g. for bash or zsh) to stdout.
+    ///
+    /// Bash:
+    ///
+    /// boundless-cli completions bash >> ~/.local/share/bash-completion/completions/boundless-cli
```
capossele: Would be better to overwrite as we do with zsh?
```suggestion
    /// boundless-cli completions bash > ~/.local/share/bash-completion/completions/boundless-cli
```

nategraf: Honestly not sure. Let me actually try that because I _may_ have copied these examples from rustup  and only tested the `zsh` one 😅

nategraf: I decided to just remove these instructions, instead of giving bad advice

PR #764, opened 2025-06-06
BM-1151: Add indexes to order table

capossele:


crates/slasher/migrations/1_slasher.sql
```diff
@@ -1,10 +1,13 @@
...
-CREATE TABLE last_block (
+CREATE INDEX IF NOT EXISTS idx_orders_expires_at ON orders(expires_at);
+CREATE INDEX IF NOT EXISTS idx_orders_lock_expires_at ON orders(lock_expires_at);
+
+CREATE TABLE IF NOT EXISTS last_block (
     id INTEGER PRIMARY KEY,
     block TEXT
-)
\ No newline at end of file
+)
```
austinabell, line 13: I think the migration should be as a separate migration file. Not certain, but wouldn't this fail to apply because the checksum wouldn't match but this migration has already been run?

capossele:
Closing for now

PR #765, opened 2025-06-06
BM-1152: Support multiple proof deliveries for the same request. Support partial fulfillment after a request is fulfilled/slashed

willpote:
Since on fulfillment & slash we clear `requestDigest` from the lock, it is:
1/ not possible to deliver a proof more than once for a request that has been locked
2/ not possible to partially fulfill a batch where one of the requests in the batch has been slashed 

We want to support a single request having multiple proofs delivered, since different proofs may have different proof types and thus verification costs/tradeoffs. 

This is also a problem for partial fulfillment, as if someone fulfills before you try to fulfill you will currently revert, and if you try to fulfill after expiry + slash you will currently revert

### DIFF
--- a/crates/boundless-market/src/contracts/artifacts/RequestLock.sol
+++ b/crates/boundless-market/src/contracts/artifacts/RequestLock.sol
@@ -118,20 +118,4 @@ library RequestLockLibrary {
             sstore(num, 0)
         }
     }
-
-    function clearSlot2(RequestLock storage requestLock) private {
-        assembly {
-            let num := add(requestLock.slot, 2)
-            sstore(num, 0)
-        }
-    }
-
-    function clearSlot1And2(RequestLock storage requestLock) private {
-        assembly {
-            let num := add(requestLock.slot, 1)
-            sstore(num, 0)
-            num := add(num, 1)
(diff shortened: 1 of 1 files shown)


contracts/src/types/RequestLock.sol
```diff
@@ -112,6 +112,13 @@ library RequestLockLibrary {
...
     }
 
+    function clearSlot1(RequestLock storage requestLock) private {
+        assembly {
+            let num := add(requestLock.slot, 1)
+            sstore(num, 0)
+        }
+    }
+
     function clearSlot2(RequestLock storage requestLock) private {
```
nategraf: We can remove `clearSlot2` and `clearSlot1And2` since we now never clear slot 2.

contracts/snapshots/BoundlessMarketBasicTest.json
```diff
@@ -1,40 +1,40 @@
 {
   "ERC20 approve: required for depositStake": "45966",
-  "bytecode size implementation": "24048",
+  "bytecode size implementation": "24381",
```
nategraf, line 3: :panic_intensifies:

PR #767, opened 2025-06-06
Skip submit batch if all requests expired, log improvements, increase reaper grace period

willpote:
* If all the requests in the batch have expired we don't need to submit it on-chain. Mostly an optimization once partial fulfillment is redeployed, but saves time/gas.
* Log improvements to allow a single log query for a request id to track the request through the entire life cycle (including aggregation and submission) and see proof times
* Increases reaper grace period


crates/broker/src/submitter.rs
```diff
@@ -180,6 +182,31 @@ where
...
+            .get_orders(&order_ids)
+            .await.context("Failed to get orders")?;
+        let expired_orders = orders
+            .iter()
+            .filter(|order| -> bool {
+                match order.fulfillment_type {
+                    FulfillmentType::LockAndFulfill => order.request.lock_expires_at() < now,
+                    FulfillmentType::FulfillAfterLockExpire => order.request.expires_at() < now,
+                    FulfillmentType::FulfillWithoutLocking => order.request.expires_at() < now,
+                }
```
austinabell: the `expire_timestamp` field on the order has this logic, but fine to do either way (handling expiry in multiple ways we can clean up later)

crates/broker/src/order_monitor.rs
```diff
@@ -490,18 +481,8 @@ where
...
-                    "{} [Lock expires at: {} ({} seconds from now), Expires at: {} ({} seconds from now)]",
-                    order.id(),
-                    order.request.lock_expires_at(),
-                    order.request.lock_expires_at().saturating_sub(now_timestamp()),
-                    order.request.expires_at(),
-                    order.request.expires_at().saturating_sub(now_timestamp())
-                ))
-                .collect::<Vec<_>>()
+            "Orders ready for proving, prioritized. Before applying capacity limits: {}",
+            orders.iter().map(|order| format!("{order}")).collect::<Vec<_>>().join(", ")
```
austinabell:
```suggestion
            orders.iter().map(ToString::to_string).collect::<Vec<_>>().join(", ")
```

Optional: could do something like this I think

austinabell (approved):

crates/broker/src/submitter.rs
willpote: Updated

PR #768, opened 2025-06-06
BM-1156: allow configurable number of pricing tasks

austinabell:


crates/boundless-market/src/contracts/bytecode.rs
```diff
@@ -1,7 +1,7 @@
...
...0011b9080820460021481151715613926576130d89260ff9160029161524b9160071c6001600160f81b031690600101614a07565b91906fa2a8918ca85bafe22016d0b997e4df60600160ff1b038411615b31579160209360809260ff5f9560405194855216868401526040830152606082015282805260015afa156122e8575f516001600160a01b03811615615b2757905f905f90565b505f906001905f90565b5050505f9160039190565b90615b605750805115615b5157805190602001fd5b630a12f52160e11b5f5260045ffd5b81511580615b91575b615b71575090565b639996b31560e01b5f9081526001600160a01b0391909116600452602490fd5b50803b15615b6956fe405787fa12a823e0f2b7631cc41b3ba8828b3321ca811111fa75cd3aa3bb5ace2682b2738d8666734f483aa3e081360183bbd5428f40ad104639f5961025142ba16a46d94261c7517cc8ff89f61c0ce93598e3c849801011dee649a6a557d1029016d09d72d40fdae2fd8ceac6b6234c7706214fd39c1cd1e609a0528c199300a16a46d94261c7517cc8ff89f61c0ce93598e3c849801011dee649a6a557d103a16a46d94261c7517cc8ff89f61c0ce93598e3c849801011dee649a6a557d100360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbcf0c57e16840df040f15088dc2f81fe391c3923bec73e23a9662efc9c229c6a00a16a46d94261c7517cc8ff89f61c0ce93598e3c849801011dee649a6a557d101237e158222e3e6968b72b9db0d8043aacf074ad9f650f0d1606b4d82ee432c00a164736f6c634300081a000a")]
```
capossele, line 4: was this intentional?

crates/broker/src/order_picker.rs
```diff
@@ -704,42 +706,64 @@ where
...
+                tokio::select! {
+                    // This channel is cancellation safe, so it's fine to use in the select!
+                    order = rx.recv() => {
+                        let order = order.ok_or_else(|| {
+                            SupervisorErr::Fault(OrderPickerErr::UnexpectedErr(anyhow::anyhow!(
+                                "Order channel closed unexpectedly"
+                            )))
+                        })?;
+
+                        pending_orders.push_back(order);
```
willpote, line 738: I'd like to be able to understand e2e whats happening with each order with 1 log query, so lets log the order id when its enqueued and dequeued

crates/boundless-market/src/contracts/bytecode.rs
austinabell: no, not sure what is non-deterministic about this build, didn't touch any of the contracts, just this from cargo build

crates/broker/src/order_picker.rs
austinabell: there is a log on price, so added this debug log for this here https://github.com/boundless-xyz/boundless/pull/768/commits/4d29bc96d52dad167f543d81483d7376b96ab7c5

crates/boundless-market/src/contracts/bytecode.rs
nategraf: I've been noticing this. It became non-deterministic, possibly between machines, at some point.

austinabell:
decided to rename to `max_concurrent_order_pricing` to be consistent with `max_concurrent_proofs` and batch size is a bit misleading. Will leave this PR open until a 👍 or some time passes for anyone to suggest a better name or raise concerns

willpote:
> decided to rename to `max_concurrent_order_pricing` to be consistent with `max_concurrent_proofs` and batch size is a bit misleading. Will leave this PR open until a 👍 or some time passes for anyone to suggest a better name or raise concerns

I'd prefer `max_concurrent_preflights` to match the terminology we use in our docs here: https://docs.beboundless.xyz/provers/terminology#preflight

Obv pricing does a bit more than run the preflight, but preflight does feel like the important part

crates/boundless-market/src/contracts/bytecode.rs
capossele: Maybe make sure you are running the same foundry version (1.2.2)

capossele: Actually never mind, as I updated just recently as well, so version should not been the problem here. Maybe it just got desynced during a merge or something

PR #771, opened 2025-06-08
Safe shutdown [wip]

willpote:
- **Safe shutdown part 1**
- **Wait for commited on shutdown**
- **Debug**


austinabell:
going to just make one more change to make pricing tasks cancellation safe and test it

Edit: done

PR #774, opened 2025-06-09
Make submitter transaction timeouts known error. Enable codes for submitter

willpote:
Transaction timeouts seems to just happen on Sepolia sometimes. Making the alarm more lenient. Also enables the Debug macro in the submitter which was missed.

### DIFF
--- a/crates/broker/src/submitter.rs
+++ b/crates/broker/src/submitter.rs
@@ -28,7 +28,7 @@ use risc0_zkvm::{
 use crate::{
     config::ConfigLock,
     db::DbObj,
-    now_timestamp,
+    impl_coded_debug, now_timestamp,
     provers::ProverObj,
     task::{RetryRes, RetryTask, SupervisorErr},
     Batch, FulfillmentType, Order,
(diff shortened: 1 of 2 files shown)


crates/broker/src/submitter.rs
```diff
@@ -58,15 +58,25 @@ pub enum SubmitterErr {
...
-            SubmitterErr::TxnConfirmationError(_) => "[B-SUB-004]",
+            SubmitterErr::BatchSubmissionFailed(submitter_errs) => {
+                // If all errors are txn confirmation errors, then we use a separate code to enable custom alerting.
+                if submitter_errs.iter().all(|e| matches!(e, SubmitterErr::TxnConfirmationError(_)))
+                {
+                    "[B-SUB-003]"
+                } else {
+                    "[B-SUB-004]"
+                }
+            }
```
nategraf: How hard would it be to move this to when the error is created? It seems a bit off to do this in the `Debug` impl

nategraf (approved):

crates/broker/src/submitter.rs
willpote: Its possible, will make that change

willpote: Handled https://github.com/boundless-xyz/boundless/pull/774/commits/864e05c5c02483fc44a7a1967332c36ffbcd20c1

PR #775, opened 2025-06-09
Add disable-dev-mode in the identity util guest to show best practice

nategraf:
When using composition via receipt verification in the guest, it is best to set `disable-dev-mode`, which is recommended for any production build of a Rust program that is verifying receipts. This adds an extra protection against accidentally turning on dev mode.


capossele:
Thanks!, I've included this commit (modified with the new zkvm version) into #755 to not create conflicts

PR #778, opened 2025-06-10
setup localnet to enable full proving easier

austinabell:
Backporting these changes I've been using to test full proving easier.
- Doesn't start broker by default with localnet
- Pass through `RISC0_DEV_MODE` env variable
- Change port of test and localnet order stream postgres instance to not conflict with bento
- Fix localnet env variable modification

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,5 +1,5 @@
 # Variables
-DEFAULT_DATABASE_URL := "postgres://postgres:password@localhost:5432/postgres"
+DEFAULT_DATABASE_URL := "postgres://postgres:password@localhost:5433/postgres"
 DATABASE_URL := env_var_or_default("DATABASE_URL", DEFAULT_DATABASE_URL)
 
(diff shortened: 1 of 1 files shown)


capossele (approved):

austinabell:
put into draft because I missed that it doesn't handle the file based URI in non-dev mode but just works if run after running in dev mode to upload the images. Perhaps that is a fine tradeoff, but just going to check if there is a cleaner way than having to start the broker twice on clean run

austinabell:
https://github.com/boundless-xyz/boundless/pull/778/commits/3431d40cf5514f8b7d7020338bab7f63b1467722 to make it easier to just uncomment to use these binaries through regular config, which means on a fresh build either:
- Modify broker toml config to uncomment these lines and run outside docker compose first to upload
- Modify broker toml and compose to uncomment those lines

Maybe a cleaner way will come to mind, so I'll leave this overnight

PR #780, opened 2025-06-10
Refactor cancellation to use tokens for graceful shutdown

austinabell:
Uses cancellation tokens instead, and cancels all non-critical tasks before waiting for other tasks to complete.

Other:
- Fixes order stream localnet to always use right RPC
- Reset config changes from #771 
- Reset logging changes also from that PR (all logs seem fine to me, although in detached mode unfortunate that logs stop on cancel -- maybe there is a workaround)
- Updates upper bound to be 2h for the cancellation in broker, 3h for compose shutdown before sending kill signal

### DIFF
--- a/crates/broker/src/order_picker.rs
+++ b/crates/broker/src/order_picker.rs
@@ -752,7 +752,7 @@ where
                         }
                     }
                     _ = cancel_token.cancelled() => {
-                        tracing::info!("Order picker received cancellation, shutting down gracefully");
+                        tracing::debug!("Order picker received cancellation, shutting down gracefully");
                         // Abort remaining tasks
                         tasks.abort_all();
                         break;
--- a/crates/broker/src/chain_monitor.rs
+++ b/crates/broker/src/chain_monitor.rs
@@ -162,7 +162,7 @@ where
                     }
                     // Handle cancellation
                     _ = cancel_token.cancelled() => {
-                        tracing::info!("Chain monitor received cancellation, shutting down gracefully");
+                        tracing::debug!("Chain monitor received cancellation, shutting down gracefully");
                         break;
                     }
                 }
--- a/crates/broker/src/market_monitor.rs
+++ b/crates/broker/src/market_monitor.rs
@@ -283,7 +283,6 @@ where
                     }
                 }
                 _ = cancel_token.cancelled() => {
-                    tracing::info!("Monitor orders received cancellation, shutting down gracefully");
                     return Ok(());
                 }
             }
@@ -388,7 +387,6 @@ where
                     }
                 }
                 _ = cancel_token.cancelled() => {
-                    tracing::info!("Monitor order locks received cancellation, shutting down gracefully");
                     return Ok(());
                 }
             }
@@ -450,7 +448,6 @@ where
                     }
                 }
                 _ = cancel_token.cancelled() => {
-                    tracing::info!("Monitor order fulfillments received cancellation, shutting down gracefully");
                     return Ok(());
                 }
             }


crates/broker/src/order_picker.rs
```diff
@@ -750,6 +751,12 @@ where
                             current_capacity = new_capacity;
                         }
                     }
+                    _ = cancel_token.cancelled() => {
+                        tracing::info!("Order picker received cancellation, shutting down gracefully");
+                        // Abort remaining tasks
+                        tasks.abort_all();
```
willpote, line 757: Related to the above comment, here is where I'd expect us to cancel any ongoing executor runs (assuming thats possible) if we went with the "task owns its own cleanup" approach

crates/broker/src/lib.rs
```diff
@@ -779,41 +796,49 @@ where
...
+        self.shutdown_and_cancel_critical_tasks(critical_cancel_token).await?;
 
         Ok(())
     }
 
-    async fn shutdown(&self) -> Result<(), anyhow::Error> {
+    async fn shutdown_and_cancel_critical_tasks(
+        &self,
+        critical_cancel_token: CancellationToken,
+    ) -> Result<(), anyhow::Error> {
```
willpote, line 826: With using cancellation tokens I was actually imagining we would end up removing this functionality, and on receiving a cancellation signal each task would be responsible for waiting for the work which it currently owns to complete (i.e. move to the next status) before shutting down. Just in terms of encapsulation, it makes more sense to me  for each task to own the logic of what it means for itself to safely shut down.

That being said, it is more complicated and a bigger task (which is why I was going with the original approach!)- but it does kind of feels like we are in a bit of a halfway state where we use cancellation tokens, but they all end up being no-ops with the current implementation.

Not blocking as this is still a big improvement, but may be something we want to move towards longer term

willpote (approved):
On the infra side a few things to update: 

https://github.com/boundless-xyz/boundless/blob/main/infra/pipelines/pipelines/prover.ts#L285
This is the max time for a deployment (includes updates/restarts, as well as any other infra changes being made e.g. if the Bonsai broker needs its docker image rebuilt, that happens here)

https://github.com/boundless-xyz/boundless/blob/main/infra/prover/components/bentoBroker.ts#L271
Max time for an update/restart

Maybe 3h for first one 2h for second?

crates/broker/src/lib.rs
austinabell: I suppose so, just the issue is that submitter would have to track the status of previous tasks shutdown and be sure of any race conditions between them. Seemed messy to me.

What do you mean by they end up being no-ops? The change from before is that the tasks before proving are actually shut down, and on shutdown of each they are more graceful as they only shut down on specific cycles rather than cancelled, which limits the possible edge cases. Perhaps I'm misunderstanding what you mean by this.

Agree that long term moving away from this is ideal, and perhaps safe to track the shutdown of previous tasks because we can assert that db writes are flushed before the following task reads, but rather avoid the race conditions to be safe first.

But if referring to the config modification here, that can be removed. I was keeping to be safe, but might just be more of a footgun in the future.

crates/broker/src/order_picker.rs
austinabell: I was considering this, but was under the assumption this wouldn't affect in-progress exec jobs. I thought those wouldn't be cancelled by our current session cancellation logic. Let me double check

crates/broker/src/lib.rs
willpote: > just the issue is that submitter would have to track the status of previous tasks shutdown

Hmm yeah, thats a fair point.. definitely tricky

> What do you mean by they end up being no-ops? 

Oh just that when each task receives the token, it just breaks out of the loop and returns immediately- it doesn't actually wait for any work to complete and run any logic to safely tear down resources or anything like that

austinabell: The intention was to handle cancellation safety also, which I blanked on and forgot to refactor after, so here is that change https://github.com/boundless-xyz/boundless/pull/780/commits/5c8bb1d92a3f35bb1a980594693f4d8571b6e88f

PR #786, opened 2025-06-11
Suggestions for #699

nategraf:
- **updates to cli.mdx from review**
- **move module doc to long_about on CLI**
- **remove suggestion to use publicnode, as it will not always work**
- **suggested edits to requests page**


sashaaldrick (approved):
ty victor! will merge now

PR #804, opened 2025-06-16
BM-1202: Adjust CPU util alarm. Minor log changes

willpote:
CPU util alarm uses maximum, which means it triggers fairly frequently during preflights. Switching to average, which is also recommended in the AWS docs. Some minor log changes for clarity based on prover feedback in TG.


capossele (approved):

austinabell:
Preflight should use usually 2 threads and at most 4 of this machine's 8. Could be the snark generation also that is triggering this -- I'll have a look when I have some time

PR #809, opened 2025-06-17
BM-1208: fix stake denom, avoid precision loss, clean up logs in picker

austinabell:
Right now most fulfill after expiry orders are just being ignored and hitting that previously confusing log:

```
INFO broker::order_picker: Removing order with request id 2546c553d857d20658ece248f7c7d0861a2406818b2830f9 because it's exec limit is below 0 cycles
```

This fixes that as it was using eth denom for stake and also losing precision on the calculation.

### DIFF
--- a/crates/broker/src/order_picker.rs
+++ b/crates/broker/src/order_picker.rs
@@ -1136,7 +1136,6 @@ mod tests {
             })
             .await;
         let order_id = order.id();
-        let request_id = order.request.id;
 
         let _request_id =
             ctx.boundless_market.submit_request(&order.request, &ctx.signer(0)).await.unwrap();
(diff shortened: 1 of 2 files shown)


crates/broker/src/order_picker.rs
```diff
@@ -385,30 +391,36 @@ where
...
-                let min_cycle_price_stake_token =
-                    min_mcycle_price_stake_token.div_ceil(U256::from(1_000_000));
-                (price / min_cycle_price_stake_token)
-                    .try_into()
-                    .context("Failed to convert U256 exec limit to u64")?
+                // (stake price * 1_000_000) / stake mcycle price = max cycles
+                (dbg!(price)
+                    .saturating_mul(ONE_MILLION)
+                    .div_ceil(dbg!(min_mcycle_price_stake_token)))
+                .try_into()
```
willpote: Debug

willpote (approved):
Yikes, good catch. We should backport this one

crates/broker/src/order_picker.rs
austinabell: Oops

PR #815, opened 2025-06-18
BM-1212: handle proving orders that were false positive skips

austinabell:
There are a few edge cases where the RPC will either return that the order was not locked when it was, a timeout that marks as skipped, or some other error like inconsistent state in the RPC calls. This is a catch all to avoid this case.

This is safe to do because when the order is skipped in the order picker, the prover does not modify the state of this after. This allows an operation to shift from skipped to pending proving. 

This does not cover the case where an order is locked that has not been skipped yet (either because it hasn't been priced or some race condition where the lock event comes before the db skip). This change can come in a future change, as is more opinionated and lower urgency

### DIFF
--- a/crates/broker/src/db/mod.rs
+++ b/crates/broker/src/db/mod.rs
@@ -129,6 +129,7 @@ pub trait BrokerDb {
     async fn get_order_compressed_proof_id(&self, id: &str) -> Result<String, DbError>;
     async fn set_order_failure(&self, id: &str, failure_str: &'static str) -> Result<(), DbError>;
     async fn set_order_complete(&self, id: &str) -> Result<(), DbError>;
+    async fn commit_skipped_order(&self, id: &str) -> Result<(), DbError>;
     /// Get all orders that are committed to be prove and be fulfilled.
     async fn get_committed_orders(&self) -> Result<Vec<Order>, DbError>;
     /// Get all orders that are committed to be proved but have expired based on their expire_timestamp.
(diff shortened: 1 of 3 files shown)


willpote (commented):
I know we've discussed before but the pattern of having different processes potentially modifying the status of the order, I still worry creates a big risk for bugs and introduces various race conditions. 

My preference would still be that we stick to an actor-style model, where each service owns the data it is processing and no one else can modify it until it passes the data on (one exception to this rule being the reaper, which is the last resort override).

With this implementation a couple things that come to mind that are hard to reason about:

1/ If I set an order from Skipped directly to PendingProving, it seems to mean that it skips the capacity logic. So while the order was marked as "Skipped" i could have accepted another order, making it now possible to overcommit to orders beyond `max_conc_proofs` and `peak_khz`

2/ The order of the different processes can lead to race conditions where the behavior is hard to reason about, for example:
- 1. Order Monitor sends lock tx
- 2. Order Monitor stuck waiting for response
- 3. Lock Tx Event received _before_ order monitor gets response (sets status to PendingProve)
- 4. Order Monitor finally receives response, tries to set status to PendingProve, but its already in that status!


One idea that I'd feel better about would be if the Order Monitor process was the one that owned updating the status even in the case where theres a lock event. E.g. when the order monitor is waiting for confirmation on the lock TX, it also monitors some channel of lock events, and if it sees an event there it proceeds with the regular logic. This way at least the updates to the status are owned by one process and serialized in a predictable way

austinabell:
> My preference would still be that we stick to an actor-style model

Perhaps semantics, but I don't think this violates it, because no two actors modify the order once it is in skipped state. Once skipped it is no longer touched by the order monitor

> 1/ If I set an order from Skipped directly to PendingProving, it seems to mean that it skips the capacity logic. So while the order was marked as "Skipped" i could have accepted another order, making it now possible to overcommit to orders beyond `max_conc_proofs` and `peak_khz`

Yes, but the alternative is to just ignore this order altogether and for sure slash it. Not sure that's better.

> 2/ The order of the different processes can lead to race conditions where the behavior is hard to reason about, for example:
> 
> * 1. Order Monitor sends lock tx
> * 2. Order Monitor stuck waiting for response
> * 3. Lock Tx Event received _before_ order monitor gets response (sets status to PendingProve)
> * 4. Order Monitor finally receives response, tries to set status to PendingProve, but its already in that status!

> One idea that I'd feel better about would be if the Order Monitor process was the one that owned updating the status even in the case where theres a lock event. E.g. when the order monitor is waiting for confirmation on the lock TX, it also monitors some channel of lock events, and if it sees an event there it proceeds with the regular logic. This way at least the updates to the status are owned by one process and serialized in a predictable way

This PR avoids the case where the tx event is received after the order monitor gets the response. The order monitor is already tracking this transaction, so not sure what that would change. Also, how much after each lock tx is sent will the order monitor wait? This loop has to resolve at some point to allow the next batch of locks to potentially be processed, so suggestion is to just add another buffer to wait for these, or maybe refactor the logic of how the order monitor handles?

Either way, seems unclear the benefit and I don't have data or examples to prove that this would be a net benefit in those cases or future cases, so I will just close for now and we can handle each case specifically as they come up

PR #821, opened 2025-06-18
add a retry on block retrieval after lock transaction

austinabell:


crates/broker/src/order_monitor.rs
```diff
@@ -288,14 +292,25 @@ where
...
-            .provider
-            .get_block_by_number(lock_block.into())
-            .await
-            .with_context(|| format!("failed to get block {lock_block}"))?
-            .with_context(|| format!("failed to get block {lock_block}: block not found"))?
-            .header
-            .timestamp;
+        // Fetch the block to retrieve the lock timestamp. This has been observed to return
+        // inconsistent state between the receipt being available but the block not yet.
+        let lock_timestamp = crate::futures_retry::retry(
```
willpote, line 302: Don't we already have retries on RPC calls https://github.com/boundless-xyz/boundless/blob/67a84a717bd65cef92636ab8e521e445b7e08065/crates/broker/src/bin/broker.rs#L43 ?

austinabell: I don't think this is covered, but perhaps possible to integrate it into the retry policy there. Will try and test both before opening this PR again

austinabell: I get that this pattern isn't ideal given there is this other retry mechanism we do not re-use, but this path seems cleaner than some hacky way to include with that trait (which doesn't cover non-error cases like this). Feel free to suggest an alternative, I don't feel strongly about this

willpote (approved):

PR #831, opened 2025-06-23
Add distributor for automating balance topups

willpote:
The distributor is a single address where we can send ETH and stake. It then handles keeping the various accounts topped up on funds.

It runs every 6 hours to minimize risk of nonce issues.


austinabell (approved):
Noting that the flow seems like it would spin and not top up other accounts or types. For example, if the first prover key fails for some non-temporal reason, it will continue to spin indefinitely and other keys or eth balances wouldn't be updated, right? And for temporal issues this service will crash and spin as fast as it is restarted, which seems like a bit of a risk.

Not sure this matters much for a service like this, just noting

willpote:
> Noting that the flow seems like it would spin and not top up other accounts or types. For example, if the first prover key fails for some non-temporal reason, it will continue to spin indefinitely and other keys or eth balances wouldn't be updated, right? And for temporal issues this service will crash and spin as fast as it is restarted, which seems like a bit of a risk.
> 
> Not sure this matters much for a service like this, just noting

Thats a good point. Ideally it would still continue. Will update

willpote:
> > Noting that the flow seems like it would spin and not top up other accounts or types. For example, if the first prover key fails for some non-temporal reason, it will continue to spin indefinitely and other keys or eth balances wouldn't be updated, right? And for temporal issues this service will crash and spin as fast as it is restarted, which seems like a bit of a risk.
> > Not sure this matters much for a service like this, just noting
> 
> Thats a good point. Ideally it would still continue. Will update

Added a couple continues instead of breaks, but it didn't seem worth catching every error for basic things like transfers

PR #846, opened 2025-06-26
BM-1253: Add order commit mode config

austinabell:
This also moves the logic for picking and committing prioritization into a separate file for easier customization.


crates/broker/src/prioritization.rs
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) 2025 RISC Zero, Inc.
+//
+// All rights reserved.
+
```
willpote, line 4: Might make more sense for the tests for these functions to be in this file

willpote (approved):

crates/broker/src/prioritization.rs
austinabell: It's not so clean to move given the test utilities are defined in each respective test module, but will do a bit of a refactor now

PR #847, opened 2025-06-26
BM-1249: Smart contracts and guests for Boundless integration of support Proof of Verifiable Work (PoVW)

nategraf:
This PR includes the smart contract and associated guests for proof of verifiable work (PoVW).

Related: https://github.com/risc0/zirgen/pull/250
Related: https://github.com/risc0/risc0/pull/3220


crates/guest/povw/mint-calculator/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+use boundless_povw_guests::log_updater::IPoVW;
+use boundless_povw_guests::mint_calculator::{
+    FixedPoint, Input, MintCalculatorJournal, MintCalculatorMint, MintCalculatorUpdate,
+};
+use risc0_steel::{ethereum::ANVIL_CHAIN_SPEC, Event};
+
+// The mint calculator ensures:
+// * An event was logged by the PoVW contract for each log update and epoch finalization.
+//   * Each event is counted at most once.
+//   * Events from an unbroken chain from initialCommit to updatedCommit. This constitutes an
```
willpote: form

crates/guest/povw/src/mint_calculator.rs
```diff
@@ -0,0 +1,348 @@
...
+        multiblock_env
+    }
+}
+
+/// An ordered map of block numbers to [EthEvmEnv] that form a subsequence in a single chain.
+pub struct MultiblockEthEvmEnv<Db, Commit>(pub BTreeMap<u64, EthEvmEnv<Db, Commit>>);
+
+impl MultiblockEthEvmEnv<StateDb, Commitment> {
+    /// Ensure that the [EthEvmEnv] in this multiblock env form a subsequence of blocks from a since
+    /// chain, all blocks being an ancestor of the latest block.
```
willpote: from a since chain?

crates/guest/povw/src/mint_calculator.rs
```diff
@@ -0,0 +1,348 @@
...
+        let mut multiblock_env = MultiblockEthEvmEnv(Default::default());
+        for env_input in self.0 {
+            let env = env_input.into_env(chain_spec);
+            if let Some(collision) = multiblock_env.0.insert(env.header().number, env) {
+                // NOTE: This could instead be handled via extending the original, if that was
+                // available in the guest. But keeping things constrained is reasonable.
+                panic!("more than one env input provided for block {}", collision.header().number);
+            };
+        }
+        // Verify the the envs form a subsequence of a since chain. This is a required check, and so
```
willpote: Verify that

contracts/src/povw/Mint.sol
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/// A mint action authorized by the mint calculator.
+struct MintCalculatorMint {
+    /// Address of the recipient for the mint.
+    address recipient;
+    /// Value of the mint towards the recipient, as a fraction of the epoch reward.
+    // NOTE: This may be larger than 1 when aggregating rewards across multiple epochs.
+    // TODO(povw): This only works if the epoch reward is constant per epoch.
+    FixedPoint value;
```
willpote: Do you plan to handle this? Curious if this works if emission rates ever need to be adjusted

contracts/src/povw/Mint.sol
```diff
@@ -0,0 +1,134 @@
...
+        if (journal.povwContractAddress != address(POVW)) {
+            revert IncorrectPovwAddress({expected: address(POVW), received: journal.povwContractAddress});
+        }
+
+        // Ensure the initial commit for each update is correct and update the final commit.
+        for (uint256 i = 0; i < journal.updates.length; i++) {
+            MintCalculatorUpdate memory update = journal.updates[i];
+
+            // On the first mint for a journal, the initialCommit should be equal to the empty root.
+            bytes32 expectedCommit = lastCommit[update.workLogId];
```
willpote: It seems that the Mint Calculator Guest processes every single prover in the epoch, which means that for every mint there could be _a lot_ of updates + mints

How feasible is it for minting to be sybil'd? Concern would be to make mint consume so much gas that the transaction can not fit in a block, which I believe would halt rewards indefinitely (?)

willpote (commented):
Looking good! Would like to spend more time reading and understanding but took a quick pass

contracts/src/povw/Mint.sol
nategraf: I imagine that we need to, since I would assume we do want the ability to change the emissions rate (e.g. on a time curve or via governance). Not totally sure the best way to handle this, and that's partly because the best solution may depend on the rules around updates.

contracts/src/povw/Mint.sol
nategraf: Ah, yes. So if this becomes an issue the transaction can be broken up. The segmenting can be done both between work logs (e.g. some work logs in one tx and some in another) or between updates (e.g. the first x updates in one tx and the next y updates in another tx). The smallest chunk allowed is a single `WorkLogUpdated` event for a single work log (i.e. a single prover).

In the case that there are many low value log updates, which might be so small its not worth paying the gas to mint the tokens, these can simply be excluded. If the owner of those work logs would like their mint, they can post the update themselves.

crates/guest/povw/mint-calculator/src/main.rs
nategraf: 6090df5b

crates/guest/povw/src/mint_calculator.rs
nategraf: 6090df5b

crates/guest/povw/src/mint_calculator.rs
nategraf: 6090df5b

contracts/src/povw/Mint.sol
willpote: Are you saying it can be supported in the current implementation, or its something we could do in a future upgrade? (maybe im missing it but I don't see how you can do it with the current Mint guest)

nategraf: The guest supports this, but the tooling does not exist to run this guest in this way (or any other for that matter, expect in tests). The way its supported in the guest is that it only enforces the completeness of updates within a single account (i.e. work log). So if one run of the guest prepares the mint for work log A, and the second run of the guest prepares the mint for work log B, those two mints can be posted in independent transactions.

nategraf:
Marked this as draft because I added `SCCACHE_RECACHE` to see if that has something to do with the build failure.

PR #857, opened 2025-07-01
Reduce the size of some error enums

c410-f3r:
A large enum used in several places can cause a negative runtime impact. In fact, this is so common that the community created several lints to prevent such a scenario.

- [large_enum_variant](https://rust-lang.github.io/rust-clippy/master/?groups=perf#large_enum_variant)
- [result_large_err](https://rust-lang.github.io/rust-clippy/master/?groups=perf#result_large_err)
- [variant_size_differences](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/types/static.VARIANT_SIZE_DIFFERENCES.html)

This PR cuts 8 bytes from 3 different `Error` enums through the use of `Box<str>`, which doesn't change any intended behavior.

Of course, 8 bytes won't make much of a runtime different in this project but it is still worthwhile IMO.

### DIFF
--- a/crates/boundless-market/src/storage/s3.rs
+++ b/crates/boundless-market/src/storage/s3.rs
@@ -60,7 +60,7 @@ pub enum S3StorageProviderError {
 
     /// Error type for missing configuration parameters.
     #[error("missing config parameter: {0}")]
-    Config(String),
+    Config(Box<str>),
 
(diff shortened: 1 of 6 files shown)


austinabell:
Not interested in this type of change, sorry! The performance is not sensitive to that minimal change in memory usage to warrant this change

PR #861, opened 2025-07-02
BM-1268: Skip max size limit

capossele:
This PR adds logic to skip the max file size limit when downloading input from a URL input type, if the client address is contained in the `allow_skip_mcycle_limit_addresses`
It also adds `0x0466ACfc0F27bBA9fBB7A8508f576527e81E83Bd` to the whitelisted addresses


crates/broker/src/storage.rs
```diff
@@ -74,7 +75,11 @@ pub(crate) async fn create_uri_handler(
             }
             let max_size = {
                 let config = &config.lock_all().expect("lock failed").market;
-                config.max_file_size
+                if skip_max_size_check {
+                    usize::MAX
+                } else {
+                    config.max_file_size
+                }
             };
```
austinabell: This conditional seems like it should be flipped? There should be no need to read the config if skipping the max size?

austinabell:
Main changes I've made since passed over:

- Changed config name so it isn't overloaded
  - This name might be a bit vague, and made backwards compatible to avoid issues -- so suggestions welcome for that as well as the `priority_requestor_address` name
- Added a new address to staging whitelist that I will use to test this config on that network

infra/prover/tomls/bonsai/broker-prod-11155111.toml
```diff
@@ -12,7 +12,7 @@ balance_warn_threshold = "5"
 balance_error_threshold = "1"
 stake_balance_warn_threshold = "15"
 stake_balance_error_threshold = "10" 
-# allow_skip_mcycle_limit_addresses = []
+allow_skip_mcycle_limit_addresses = ["0x0466ACfc0F27bBA9fBB7A8508f576527e81E83Bd"]
```
nategraf: Is this one right?

nategraf (approved):

infra/prover/tomls/bonsai/broker-prod-11155111.toml
austinabell: no, not sure why that one was missed from a search. Thanks

PR #862, opened 2025-07-03
BM-1269: remove skip order on priority above mcycle limit

austinabell:
This check could likely also be removed, if we trust the execution cap limits, but keeping for now unless someone indicates otherwise


austinabell:
draft as I noticed one other change I think that should be made

PR #866, opened 2025-07-03
BM-1272: implement deny list addresses config

austinabell:


crates/broker/src/config.rs
```diff
@@ -611,6 +617,7 @@ max_stake = "0.1"
 max_file_size = 50_000_000
 max_fetch_retries = 10
 allow_client_addresses = ["0x0000000000000000000000000000000000000000"]
+deny_requestor_addresses = ["0x0000000000000000000000000000000000000000"]
```
nategraf, line 620: It seems like it should be an error for there to be overlap in the allow and deny lists. Actually, what is the allow list?

nategraf (approved):

crates/broker/src/config.rs
austinabell: I suppose so, the current behaviour is that if allow lists defined, it will only allow requests from that, and then if any of those are in deny it will ignore those ones. I suppose the semantics could be to error, but assumed there might be a case where the deny list would be dynamically updated and this behaviour seemed reasonable.

PR #870, opened 2025-07-05
Bump development version to 0.13

nategraf:


austinabell (approved):

nategraf:
Still not sure why this PR in particular is causing the composition example to fail 😕

capossele:
> Still not sure why this PR in particular is causing the composition example to fail 😕

Not 100%sure but maybe related to bumping the risc0-groth16 crate? In composition we do verify a groth16 receipt so maybe it's that? Not sure though

austinabell:
I think it was the `groth16_latest` pulling in 2.2 params that caused composition issues (not certain)

PR #878, opened 2025-07-07
BM-1286: remove nvidia runtime requirement from CPU agent types

austinabell:
Will likely leave in PR until I can test on more machine types, but this seems to work on the machines I've tested so far, although on some it seems the cuda version specifically only works on this.

Unblocks being able to run certain agents on CPU only machines, as the aux, snark, exec sub agents do not call into cuda


walirt:
Can GPU agents run on different servers from other components? I tried doing this, but got an error.

image

austinabell:
> Can GPU agents run on different servers from other components? I tried doing this, but got an error.
> 
> image

You need to be running the agents with the same versions. If using main right now, you will be using different verifier params than if you have agents running with the latest stable release, for example.

walirt:
image
image
image
The agent is the same version, but the aux agent seems to need to run in a graphics card environment to work

capossele:
@walirt you might want to double check what `boundless` cli version you are running cause that will dictate what `risc0_zkvm::VERSION` the client is requesting and your bento should be running with 2.3.0.
Just in case, make sure to install the boundless cli directly from main rather than using the latest from crates. We have not released the new version on crates yet

walirt:
> @walirt you might want to double check what `boundless` cli version you are running cause that will dictate what `risc0_zkvm::VERSION` the client is requesting and your bento should be running with 2.3.0. Just in case, make sure to install the boundless cli directly from main rather than using the latest from crates. We have not released the new version on crates yet

Thank you, although the agent version is the same, the hash is different. After keeping it consistent, they work. But after the stress test, peak_prove_khz is significantly reduced. Is this because of the bus?

austinabell:
> > @walirt you might want to double check what `boundless` cli version you are running cause that will dictate what `risc0_zkvm::VERSION` the client is requesting and your bento should be running with 2.3.0. Just in case, make sure to install the boundless cli directly from main rather than using the latest from crates. We have not released the new version on crates yet
> 
> Thank you, although the agent version is the same, the hash is different. After keeping it consistent, they work. But after the stress test, peak_prove_khz is significantly reduced. Is this because of the bus?

Hard to say without knowing about your infra setup. I will note that it is important to have the prove agents connection to the DB be solid (if you can, run on the same machine) as that IO time will likely become a bottleneck. That would be my first guess of what to look into

edit: or do you mean using latest version reduces your proving hz rate?

walirt:
> > > @walirt you might want to double check what `boundless` cli version you are running cause that will dictate what `risc0_zkvm::VERSION` the client is requesting and your bento should be running with 2.3.0. Just in case, make sure to install the boundless cli directly from main rather than using the latest from crates. We have not released the new version on crates yet
> > 
> > 
> > Thank you, although the agent version is the same, the hash is different. After keeping it consistent, they work. But after the stress test, peak_prove_khz is significantly reduced. Is this because of the bus?
> 
> Hard to say without knowing about your infra setup. I will note that it is important to have the prove agents connection to the DB be solid (if you can, run on the same machine) as that IO time will likely become a bottleneck. That would be my first guess of what to look into
> 
> edit: or do you mean using latest version reduces your proving hz rate?

I think it may be an IO problem. When the database and the prover agent are on the same machine, the `peak_prove_khz` of the 2-card 4090 can reach more than 1000. When they are on different machines, this value drops to around 600. 
I don't know how big this gap is when actually running proofs, which makes the benefits of horizontally scaling GPUs even smaller.

austinabell:
> I think it may be an IO problem. When the database and the prover agent are on the same machine, the `peak_prove_khz` of the 2-card 4090 can reach more than 1000. When they are on different machines, this value drops to around 600. I don't know how big this gap is when actually running proofs, which makes the benefits of horizontally scaling GPUs even smaller.

Are those two machines co-located? Are you running a local version of postgres as is or something custom like having some connection pooling or perhaps having replicas? This is not something we have extensively explored yet. 

In aws, we don't notice this amount of fall off when running co-located through RDS, but there should definitely be ways to optimize this to horizontally scale in your case.

walirt:
> > I think it may be an IO problem. When the database and the prover agent are on the same machine, the `peak_prove_khz` of the 2-card 4090 can reach more than 1000. When they are on different machines, this value drops to around 600. I don't know how big this gap is when actually running proofs, which makes the benefits of horizontally scaling GPUs even smaller.
> 
> Are those two machines co-located? Are you running a local version of postgres as is or something custom like having some connection pooling or perhaps having replicas? This is not something we have extensively explored yet.
> 
> In aws, we don't notice this amount of fall off when running co-located through RDS, but there should definitely be ways to optimize this to horizontally scale in your case.

I use one server as the database and other agents running environment, and another server to run the GPU agent. The two machines are in a similar area, and the latency from the GPU server to the database server is about 50ms.

PR #880, opened 2025-07-10
Add deployed signal contract addresses to priority_requestor_addresses for Bonsai

nategraf:


capossele (approved):
Shouldn’t we add the same to the bento toml and template as well?

nategraf:
> Shouldn’t we add the same to the bento toml and template as well?

This is intentional. Our Bento provers don't really have capacity right now. Actually, I probably should add this address to the deny list

PR #881, opened 2025-07-10
BM-1297: update Bonsai API to staging

capossele:


nategraf (approved):

nategraf:
Should this PR include prod as well?

capossele:
> Should this PR include prod as well?

Yeah we can add them now, or we can do that on a second PR once we are happy with it

nategraf:
I'll merge this now to get staging switched over, and try to follow up with a PR for prod shortly.

PR #898, opened 2025-07-14
BM-1329: add more info to broker.toml descriptions

sashaaldrick:
- added a little bit more info for the 'allow_client_addresses' and 'peak_prove_khz' options


austinabell (approved):
Wonder if better to just reference the template file for all definitions, instead of documenting and keeping up to date in two places?

sasha-computer:
Yeah thats a good idea, let me make those changes later today, for now we will merge this

PR #903, opened 2025-07-15
BM-1338: Improve logging for the pricing queue, skip fulfilled already FulfillAfterLockExpire orders

willpote:
- **Log status of order pricing / preflight queue and tasks**
- **Skip already fulfilled orders for FulfillAfterLockExpire**


crates/broker/src/order_picker.rs
```diff
@@ -810,8 +824,18 @@ where
...
+                        tracing::debug!(
+                            "Queued order {} to be priced. Currently {} queued pricing tasks: {}",
+                            order_id,
+                            pending_orders.len(),
+                            pending_orders
+                                .iter()
+                                .map(ToString::to_string)
+                                .collect::<Vec<_>>()
+                                .join(", ")
+                        );
```
austinabell, line 848: This seems like it would clutter logs a bit if a lot of order requests come in. Probably fine for now

crates/broker/src/order_picker.rs
```diff
@@ -97,6 +98,7 @@ pub struct OrderPicker<P> {
     priced_orders_tx: mpsc::Sender<Box<OrderRequest>>,
     stake_token_decimals: u8,
     order_cache: OrderCache,
+    active_tasks: Arc<Mutex<HashMap<String, Box<OrderRequest>>>>,
```
austinabell, line 111: I don't think you need the mutex logic to track these, as you can just remove the entry when the task is joined. I can make that change in a follow up as I'm changing related things

crates/broker/src/order_picker.rs
```diff
@@ -867,10 +892,43 @@ where
...
+                            let mut active_tasks = picker_clone.active_tasks.lock().await;
+                            active_tasks.insert(order_id.clone(), order.clone());
+
+                            // Log current active tasks
+                            let task_details = active_tasks
+                                .values()
+                                .map(ToString::to_string)
+                                .collect::<Vec<_>>()
+                                .join(", ");
+                            tracing::info!("Current pricing tasks: [{}]", task_details);
```
austinabell: This seems to verbose for an info log. Also, why not log outside of the loop for orders? This seems like it would just spam this log for each order added?

austinabell (approved):

crates/broker/src/order_picker.rs
```diff
@@ -867,10 +892,43 @@ where
 
                         let picker_clone = picker.clone();
                         let task_cancel_token = cancel_token.child_token();
+                        let order_id = order.id();
+
+                        // Track this task as active
+                        {
+                            let mut active_tasks = picker_clone.active_tasks.lock().await;
+                            active_tasks.insert(order_id.clone(), order.clone());
```
austinabell, line 910: wait, why do you need the order itself? Doesn't seem like it's used for anything other than logging the id, which is the key?

crates/broker/src/lib.rs
```diff
@@ -172,7 +172,7 @@ fn format_order_id(
 /// Order request from the network.
 ///
 /// This will turn into an [`Order`] once it is locked or skipped.
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
```
austinabell, line 185:
```suggestion
#[derive(Serialize, Deserialize, Debug)]
```

In relation to previous comment, would be nice to avoid this impl to avoid race conditions where we have duplicates of the same order in different states accidentally

crates/broker/src/order_picker.rs
willpote: I did this just so we can log using the Display impl of the order, which includes the useful logging around when the orders expire.

crates/broker/src/order_picker.rs
willpote: Right, I did it such that whenever an order task starts or completes we log. I wanted it such that you can search an order id and see its full flow (when it was added/removed from the queue, start/stopped), and also see which other orders its competing with

crates/broker/src/order_picker.rs
willpote: For now I'd prefer we over-log and reduce over time. Much prefer that to the current situation where its not really possible to track whats going on

Note if particular lines are really noisy you can filter them out in Cloudwatch.

crates/broker/src/lib.rs
willpote: Thats fair. We should only be logging order id + expiry times (via the Display impl), so should be safe, but generally agree

PR #904, opened 2025-07-15
BM-1339: Cancel preflights when locked/fulfilled

austinabell:
Closes BM-1312

### DIFF
--- a/crates/broker/src/order_picker.rs
+++ b/crates/broker/src/order_picker.rs
@@ -975,7 +975,7 @@ where
                                 .values()
                                 .flat_map(|orders| orders.keys().cloned())
                                 .collect();
-                            tracing::info!(
+                            tracing::debug!(
                                 "Removed pricing task {}. Current in-progress pricing tasks: [{}]",
                                 order_id,
                                 task_details.join(", ")
--- a/crates/broker/src/order_picker.rs
+++ b/crates/broker/src/order_picker.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -24,7 +24,7 @@ use crate::{
     provers::{ProverError, ProverObj},
     storage::{upload_image_uri, upload_input_uri},
     task::{RetryRes, RetryTask, SupervisorErr},
-    utils, FulfillmentType, OrderRequest,
+    utils, FulfillmentType, OrderRequest, OrderStateChange,
 };
 use crate::{now_timestamp, provers::ProofResult};
 use alloy::{
@@ -43,7 +43,7 @@ use boundless_market::{
 };
 use moka::future::Cache;
 use thiserror::Error;
-use tokio::sync::{mpsc, Mutex};
+use tokio::sync::{broadcast, mpsc, Mutex};
 use tokio::task::JoinSet;
 use tokio_util::sync::CancellationToken;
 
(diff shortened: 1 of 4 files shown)


crates/broker/src/order_picker.rs
```diff
@@ -825,10 +831,31 @@ where
...
                 read_config().map_err(SupervisorErr::Fault)?;
-            let mut tasks: JoinSet<()> = JoinSet::new();
+            let mut tasks: JoinSet<(String, U256)> = JoinSet::new();
             let mut rx = picker.new_order_rx.lock().await;
+            let mut order_state_rx = picker.order_state_tx.subscribe();
             let mut capacity_check_interval = tokio::time::interval(MIN_CAPACITY_CHECK_INTERVAL);
             let mut pending_orders: Vec<Box<OrderRequest>> = Vec::new();
+            let mut active_tasks: BTreeMap<U256, BTreeMap<String, CancellationToken>> =
+                BTreeMap::new();
+
```
willpote: Is there a reason we're using `RequestId` as opposed to `OrderId` as the key for this map? Although we don't fully support it yet, I believe this design would break "multiple request same id" since the hash of the request is not part of the key.

crates/broker/src/order_picker.rs
```diff
@@ -847,8 +874,70 @@ where
...
-                    _ = tasks.join_next(), if !tasks.is_empty() => {
-                        tracing::trace!("Pricing task completed ({} remaining)", tasks.len());
+                    Ok(state_change) = order_state_rx.recv() => {
+                        match state_change {
+                            OrderStateChange::Locked { request_id, prover } => {
+                                tracing::debug!("Received order state change for request 0x{:x}: Locked by prover {:x}",
+                                    request_id, prover);
+
+                                // Cancel all active tasks for this request_id
+                                cancel_active_tasks(&mut active_tasks, request_id, "locked");
```
willpote: Since on lock we trigger creation of a FulfillAfterLockExpires order, is there a possible race condition here where that order enters pricing before this line is executed, and thus gets cancelled mistakenly?

Generally I'm a bit confused why we want to cancel "all" the active tasks here- wouldn't we just want to focus in on the LockAndFulfill one?

crates/broker/src/order_picker.rs
```diff
@@ -847,8 +874,70 @@ where
...
+                                    active_tasks.remove(&request_id);
+                                }
+                            }
+
+                            // Log current active tasks after removal
+                            let task_details: Vec<String> = active_tasks
+                                .values()
+                                .flat_map(|orders| orders.keys().cloned())
+                                .collect();
+                            tracing::info!(
```
willpote: debug

crates/broker/src/order_picker.rs
austinabell: The order id is the inner map. The reason for two tiered is so that when an order id is locked and fulfilled, all orders with that request id are cancelled (when relevant)

crates/broker/src/order_picker.rs
austinabell: oops, this was a bad refactor. I only want to cancel LockAndFulfill in this case

crates/broker/src/order_picker.rs
austinabell: I just copied as it was, but yeah definitely too loud

crates/broker/src/order_picker.rs
austinabell: https://github.com/boundless-xyz/boundless/pull/904/commits/8da769881b66a853a8299511dcc399ed77691173

crates/broker/src/order_picker.rs
austinabell: https://github.com/boundless-xyz/boundless/pull/904/commits/9593b1f09a710e1f570c8ce7738c647511f72182

crates/broker/src/order_picker.rs
willpote: Ah gotcha

willpote (approved):

PR #910, opened 2025-07-16
refactor: update dotenv import to use ES6 style in slasher/index.ts

dizer-ti:
Replace CommonJS require('dotenv').config() with ES6 import * as dotenv from 'dotenv' and move the config() call inside the exported function. This maintains consistent import style across the codebase and ensures proper initialization of environment variables at runtime


dizer-ti:
After additional analysis, I noticed that the file uses mixed-style modules (CommonJS export with `export =`), so changing the import from require() to ES6 import may be incompatible with the current project configuration. I'm closing this PR

PR #915, opened 2025-07-16
BM-1348: convert the execution limit to mi-cycles thought bonsai

austinabell:
https://github.com/risc0/risc0/blob/723813051d1d2d09a0b6721e1c05ad971f5c20ec/bento/crates/workflow/src/tasks/executor.rs#L353

### DIFF
--- a/crates/broker/src/provers/bonsai.rs
+++ b/crates/broker/src/provers/bonsai.rs
@@ -303,7 +303,7 @@ impl Prover for Bonsai {
     ) -> Result<ProofResult, ProverError> {
         self.retry_only(
             || async {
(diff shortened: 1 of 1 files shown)


crates/broker/src/provers/bonsai.rs
```diff
@@ -303,6 +303,8 @@ impl Prover for Bonsai {
     ) -> Result<ProofResult, ProverError> {
         self.retry_only(
             || async {
+                // Convert cycles to MiB cycles (1024*1024) for Bonsai API
```
austinabell:
```suggestion
                // Convert cycles to Mi-cycles (1024*1024) for Bonsai API
```

willpote (approved):

PR #917, opened 2025-07-17
BM-1350: implement cache for preflights with same image/input

austinabell:
To avoid redundant preflights

~~TODO:~~
- [x] Resolve case where skip because session limit can stop other orders from being preflighted (potential DOS condition)


crates/broker/src/order_picker.rs
```diff
@@ -547,42 +566,148 @@ where
...
+        let preflight_result = loop {
+            let prover = &self.prover;
+            let config = &self.config;
+            let request = &order.request;
+            let order_id_clone = &order_id;
+
+            let result = self
+                .preflight_cache
+                .try_get_with(cache_key.clone(), async move {
+                    // Upload image and input only if not cached
```
willpote: Would be nice to add a note on the magic here that makes it so that even in-progress preflights result in a cache hit:

https://docs.rs/moka/latest/moka/sync/struct.Cache.html#concurrent-calls-on-the-same-key-2

willpote (approved):
Approving, pending validating tokio cancellations would break this

PR #931, opened 2025-07-18
BM-1365: refactoring cycle limits

austinabell:
These changes are to prevent redundant executions when the lock and fulfill session limit is less than the fulfill after lock one and the exec hits session limits. Also refactors the logic of the session limit calculation, to try to simplify with this added complexity of calculating both limits

### DIFF
--- a/crates/broker/src/order_picker.rs
+++ b/crates/broker/src/order_picker.rs
@@ -850,6 +850,12 @@ where
 
     /// Calculates the cycle limit for the preflight and also for the max cycles that this specific
     /// order variant will consider proving for.
+    ///
+    /// The reason for calculating both preflight and prove limits is to execute the order with
+    /// a large enough cycle limit for both lock and fulfill orders as well as for if the order
+    /// expires and to prove after lock expiry so that the execution can be cached and only happen
+    /// once. The prove limit is the limit for this specific order variant and decides the max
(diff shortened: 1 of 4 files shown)


BMG042 (approved):

BMG042:
> The logic I think is still slightly off, will re-read another day but opening for visibility

Bless

BMG042:
Done

crates/broker/src/order_picker.rs
```diff
@@ -945,6 +847,146 @@ where
         let balance = self.market.balance_of_stake(self.provider.default_signer_address()).await?;
         Ok(balance)
     }
+
+    /// Calculates the cycle limit for the preflight and also for the max cycles that this specific
+    /// order variant will consider proving for.
+    fn calculate_exec_limits(
```
willpote: Can you add some documentation about the logic + reasoning here? Doing this only really makes sense in the context of the preflight cache mechanism and that LockAndFulfill are kicked off before FulfillAfterLockExpire, which may not be clear to others

austinabell: https://github.com/boundless-xyz/boundless/pull/931/commits/e2e3176341e0d2d0a2bb761b19a266dec20a7810 good callout, I thought I had documented this more already

BMG042:
prove

PR #933, opened 2025-07-21
BM-1376: bump bento image to 2.3.1

austinabell:
Looking at the diff https://github.com/risc0/risc0/compare/v2.3.0...v2.3.1 it seems not necessary to update our min version in our libraries to be this version. Feel free to share a reason why this might be necessary. (cc @capossele I know you suggested updating the cargo locked deps)

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -11,7 +11,7 @@ x-base-environment: &base-environment
   RUST_BACKTRACE: 1
 
 x-agent-common: &agent-common
-  image: risczero/risc0-bento-agent:2.3.0@sha256:5b029fb8074b3273b45e6e8fb4d6dbd86000a216688d8f1429eb893686cc1ff8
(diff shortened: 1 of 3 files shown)


capossele:
> Looking at the diff [risc0/risc0@v2.3.0...v2.3.1](https://github.com/risc0/risc0/compare/v2.3.0...v2.3.1) it seems not necessary to update our min version in our libraries to be this version. Feel free to share a reason why this might be necessary. (cc @capossele I know you suggested updating the cargo locked deps)

The only reason I can think of for a full upgrade of the lock files would be possibly to drop https://github.com/boundless-xyz/boundless/blob/5b646f3012d73cf968fc2747836ce94c2cf75b87/crates/guest/util/identity/Cargo.toml#L11 from the dependencies of the identity guest program. Aside from that all good

capossele (approved):

austinabell:
> The only reason I can think of for a full upgrade of the lock files would be possibly to drop .. from the dependencies of the identity guest program. Aside from that all good

[`82e3b9a` (#933)](https://github.com/boundless-xyz/boundless/pull/933/commits/82e3b9aaf7817283a44393a462f8063821c68165)

PR #940, opened 2025-07-23
BM-1394: in case of u64 overflow in cycles calculation, use max value

austinabell:
Pointing at #937 so draft until that comes in


crates/broker/src/order_picker.rs
```diff
@@ -898,7 +898,7 @@ where
             let initial_stake_based_limit =
                 (price.saturating_mul(ONE_MILLION).div_ceil(min_mcycle_price_stake_token))
                     .try_into()
-                    .context("Failed to convert U256 exec limit to u64")?;
+                    .unwrap_or(u64::MAX)?;
```
austinabell:
```suggestion
                    .unwrap_or(u64::MAX);
```

PR #953, opened 2025-07-25
BM-1403: clean up preflight cache logic

austinabell:
Intention was to keep the logic simple, but this seems more confusing to read, and was worried about future footguns.


austinabell:
Closing because do not want to make this basically cosmetic change with other changes going on right now. Might revisit later

PR #956, opened 2025-07-25
BM-1408: Improve deployment testing

capossele:
Drop one fulfill test and keep the more specific one that uses a defined selector (must be set as env).
In this way if not set, that test will fail immediately but all the rest will be run.
This is done to avoid being rate limited by the piñata gateway and remove friction to keep the selector updated.

To run the test, after exporting both RPC_URL and CHAIN_KEY:

```bash
SELECTOR=0xbb001d44 FOUNDRY_PROFILE=deployment-test forge test --fork-url $RPC_URL -vvvv
```

Bonsai api key and url can also be set to use bonsai as backend. 
Moreover, optional IPFS_GATEWAY and PINATA_JWT can be set to bypass rate limits if needed

The Release GitHub workflow is dropped as most of those public RPC endpoints do not support being used with anvil in fork-mode


capossele:
Closing as superseded by #962

PR #980, opened 2025-07-31
memoize the order id generation

austinabell:
To avoid redundant hashing on id calls. Keeping the API the same for simplicity.

### DIFF
--- a/crates/broker/src/submitter.rs
+++ b/crates/broker/src/submitter.rs
@@ -818,6 +818,7 @@ mod tests {
             chain_id,
             total_cycles: None,
             proving_started_at: None,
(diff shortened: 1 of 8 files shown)


BMG042 (commented):
Trying

BMG042 (commented):
Trying

crates/broker/src/submitter.rs
```diff
@@ -818,6 +818,7 @@ mod tests {
             chain_id,
             total_cycles: None,
             proving_started_at: None,
```
BMG042, line 820: Start

BMG042 (commented):
Done ✅

PR #987, opened 2025-08-03
Revert "BM-1435: memoize the order id generation"

BMG042:
Reverts boundless-xyz/boundless#980


austinabell:
If this isn't an AI agent, please open an issue with the issue you ran into for the motivation for reverting this commit

BMG042:
> If this isn't an AI agent, please open an issue with the issue you ran into for the motivation for reverting this commit

PR #988, opened 2025-08-04
Increase bonsai alarm threshold, fix stake reward log

willpote:
Bonsai alarms keep going off due to DB locked issues. Given we intend to wind down the bonsai prover within a month, it does not seem worth fixing anymore. Therefore increasing the threshold of errors that are related to DB locked to reduce noise. This includes the specific db locked errors as well as various unexpected errors and supervisor errors that trigger as cascading errors. We leave the Bento alarms untouched, so any broker bugs should still be caught.

Also modifying the stake reward log to account for if the prover is the first secondary prover.


austinabell:
Closes BM-1232

PR #1004, opened 2025-08-12
BM-1466: Ec2/claim digest

ec2:
Add support for a new Predicate variant which matches on the claim digest.

The main motivation for this is to support a new Groth16 which is based on blake3 for BitVM. The current groth16 compression commits to 5 public values. This new Groth16 proof commits to 1 public value, which is the blake3 hash of the mentioned values such as the image id.

Essentially, this adds support for custom claim digests AND adds support for when the requestor does not care about the journal. 

Notable smart contract changes: 
- New `FulfillmentData` struct which replaces `imageId` and `journal` in `Fulfillment` which contains the journal and image id associated with the proof. Callbacks are only supported for fulfillments that provide the imageid and journal. 
- `Fulfillment` struct contains the claimDigest now which is used to verify the integrity of the receipt. Before, it was calculated from the imageId and journal on chain. 
- `imageId` is removed from Requirements. Instead, it is encoded into the `data` field of `Predicate` if the request is NOT claim digest match. This is because otherwise the `imageId` will be unverifiable and untrustworthy data. 

Notable assessor guest changes:
- Remove the verification of the integrity of the Fulfillment. This is because this is already done in the set builder and on chain. Also it would be very difficult to do for custom claim digests.

Things to note on broker/indexer:
- The SQL schema has changed. DBs will have to be nuked. Is this a problem?
- The broker will no longer be able to use a cached program image in the case the request is a ClaimDigestMatch. This is because it no longer has access to the image id. This means that it will always have to be downloaded. It could technically be cached by the imageURL instead, but that may cause problems as links may change. 

TODO: 
- fix docs
- some cleanup


contracts/snapshots/BoundlessMarketBasicTest.json
```diff
@@ -1,40 +1,41 @@
 {
   "ERC20 approve: required for depositStake": "45966",
-  "bytecode size implementation": "24381",
+  "bytecode size implementation": "23636",
```
capossele: Love this!

contracts/src/BoundlessMarket.sol
```diff
@@ -333,13 +337,31 @@ contract BoundlessMarket is
...
 
             uint256 callbackIndexPlusOne = fillToCallbackIndexPlusOne[i];
-            if (callbackIndexPlusOne > 0) {
+            // We do not support callbacks for claim digest matches.
+            if ((fill.predicateType != PredicateType.ClaimDigestMatch) && (callbackIndexPlusOne > 0)) {
+                (bytes32 imageId, bytes calldata journal) = _decodeCallbackData(fill.callbackData);
+
                 AssessorCallback calldata callback = assessorReceipt.callbacks[callbackIndexPlusOne - 1];
-                _executeCallback(fill.id, callback.addr, callback.gasLimit, fill.imageId, fill.journal, fill.seal);
+                _executeCallback(fill.id, callback.addr, callback.gasLimit, imageId, journal, fill.seal);
```
capossele: If I'm reading this correctly, it seems that imageId and journal here are not authenticated.
It is true that we are using the correct values in the assessor guest to compute the claimDigest, however on the contract side an adversary could replace their values. If my line of thinking makes sense we could either commit to it as part of the assessor journal, or alternatively, we could recompute the claimDigest from those values in case the predicate type is != than PredicateType.ClaimDigestMatch

contracts/test/BoundlessMarket.t.sol
```diff
@@ -2538,14 +2619,22 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
-        bytes32 imageId0 = fills[0].imageId;
-        bytes memory journal0 = fills[0].journal;
+        // bytes32 imageId0 = fills[0].imageIdOrClaimDigest;
+        // bytes memory journal0 = fills[0].journal;
+
+        // fills[0].imageIdOrClaimDigest = fills[1].imageIdOrClaimDigest;
+        // fills[1].imageIdOrClaimDigest = imageId0;
+
+        // fills[0].journal = fills[1].journal;
+        // fills[1].journal = journal0;
```
capossele:
```suggestion
```

crates/boundless-cli/src/bin/boundless.rs
```diff
@@ -716,10 +748,11 @@ async fn handle_proving_command(cmd: &ProvingCommands, client: StandardClient) -
             let session_info = execute(&request).await?;
             let journal = session_info.journal.bytes;
 
-            if !request.requirements.predicate.eval(&journal) {
-                tracing::error!("Predicate evaluation failed for request");
-                bail!("Predicate evaluation failed");
-            }
+            // TODO(ec2): how do we check this?
```
capossele: Can't we move this check within the previous match, so that we can then use the `FulfillmentData` enum?
Or we can just fill the fulfillment data there, and then do the check once here

crates/boundless-cli/src/bin/boundless.rs
```diff
@@ -1183,28 +1216,32 @@ where
...
 
-        // Verify predicate
-        ensure!(
-            request.requirements.predicate.eval(&journal),
-            "Preflight failed: Predicate evaluation failed. Journal: {}, Predicate type: {:?}, Predicate data: {}",
-            hex::encode(&journal),
-            request.requirements.predicate.predicateType,
-            hex::encode(&request.requirements.predicate.data)
-        );
+        // TODO(ec2): how do we check when we have custom claim digests?
```
capossele: I think we can still evaluate the predicate here based on the predicate type and potentially the proof type (e.g., the specified selector) no??

contracts/src/BoundlessMarket.sol
capossele: Also now that I think about it, are we only guaranteeing the callbackData in case of a callback is required? Cause in that case the only way to get back a journal now is by requesting a callback and I'm not sure is what we want

crates/boundless-market/src/contracts/artifacts/CallbackData.sol
```diff
@@ -0,0 +1,19 @@
...
+//
+// Use of this source code is governed by the Business Source License
+// as found in the LICENSE-BSL file.
+pragma solidity ^0.8.24;
+
+using CallbackDataLibrary for CallbackData global;
+
+/// @title Callback Struct and Library
+/// @notice Represents a callback configuration for proof delivery
+struct CallbackData {
```
capossele: I'm the worst at naming things, however this would make sense if the imageId and journal are always only used for this purpose. Otherwise it might be confusing if we have not requested a callback and we are getting back a structure called like that, for instance when waiting for fulfillment on the rust side

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -430,9 +433,10 @@ impl ProofRequest {
         }
         Url::parse(&self.imageUrl).map(|_| ())?;
 
-        if self.requirements.imageId == B256::default() {
-            return Err(RequestError::ImageIdIsZero);
-        }
+        // TODO(ec2): fixme
```
capossele: Maybe here we could run a similar check for the predicate, and based on the type check that the corresponding field is not zero

contracts/src/types/Fulfillment.sol
```diff
@@ -15,16 +16,14 @@ struct Fulfillment {
...
-    // TODO: Add a flag in the request to decide whether to post the journal. Note that
-    // if the journal and journal digest do not need to be delivered to the client, imageId will
-    // be replaced with claim digest, since it is captured in the requirements on the request,
-    // checked by the Assessor guest.
-    /// @notice Journal committed by the guest program execution.
-    /// @dev The journal is checked to satisfy the predicate specified on the request's requirements.
-    bytes journal;
+    /// @notice The `PredicateType` of the request that is being fulfilled.
+    /// @dev When the `PredicateType` is `ClaimDigestMatch`, the imageIdOrClaimDigest field is the claim digest,
+    /// and otherwise it is the image ID of the guest that was executed.
```
capossele: Is this still true?

contracts/src/types/Predicate.sol
```diff
@@ -5,7 +5,10 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.24;
 
+import {ReceiptClaim, ReceiptClaimLib} from "risc0/IRiscZeroVerifier.sol";
+
 using PredicateLibrary for Predicate global;
+using ReceiptClaimLib for ReceiptClaim;
 
 /// @title Predicate Struct and Library
 /// @notice Represents a predicate and provides functions to create and evaluate predicates.
```
capossele: We are now overloading a bit the semantic of the data field here, so might be worth expanding the description explaining what data might contain based on the different types?

crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -1154,7 +1155,7 @@ impl<P: Provider> BoundlessMarketService<P> {
         self.query_request_submitted_event(request_id, None, None).await
     }
 
-    /// Returns journal and seal if the request is fulfilled.
+    /// Returns callback data and seal if the request is fulfilled.
```
capossele: I'm not 100% convinced this naming makes sense here, how about `claim_data` or something else but not necessarily related to the callback concept

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -551,22 +550,168 @@ impl Requirements {
...
+                        == fulfillment_data
+                            .image_id()
+                            .expect("fulfillment data doesnt have image id, but is prefix match")
+                            .as_bytes()
+            }
+            PredicateType::ClaimDigestMatch => {
+                // self.data.as_ref()
+                //     == ReceiptClaim::ok(Digest::from_bytes(image_id.0), journal.as_ref().to_vec())
+                //         .digest()
+                //         .as_bytes()
```
capossele:
```suggestion
```

crates/broker/proptest-regressions/db/fuzz_db.txt
capossele, line 1: Is this a new thing?

crates/broker/src/aggregator.rs
```diff
@@ -206,10 +210,20 @@ impl AggregatorService {
...
 
+            let fulfillment_data = match order.request.requirements.predicate.predicateType {
+                PredicateType::ClaimDigestMatch => FulfillmentData::from_claim_digest(
+                    order.request.requirements.predicate.claim_digest().unwrap(),
+                ),
+                _ => FulfillmentData::from_image_id_and_journal(
+                    Digest::from_hex(order.image_id.unwrap()).unwrap(),
+                    journal,
+                ),
+            };
```
capossele: nit: this could be made a method/function as we use it several times within the codebase

crates/broker/src/db/mod.rs
```diff
@@ -1148,23 +1142,23 @@ mod tests {
...
+
+    //     let submit_order: (ProofRequest, Bytes, String, B256, U256, FulfillmentType) =
+    //         db.get_submission_order(&order.id()).await.unwrap();
+    //     assert_eq!(submit_order.0, order.request);
+    //     assert_eq!(submit_order.1, order.client_sig);
+    //     assert_eq!(submit_order.2, order.proof_id.unwrap());
+    //     assert_eq!(submit_order.3, order.request.requirements.imageId);
+    //     assert_eq!(submit_order.4, order.lock_price.unwrap());
+    //     assert_eq!(submit_order.5, order.fulfillment_type);
+    // }
```
capossele: what happened here?

crates/broker/src/storage.rs
```diff
@@ -347,18 +348,20 @@ pub async fn upload_image_uri(
...
-    let required_image_id = Digest::from(request.requirements.imageId.0);
-    let image_id_str = required_image_id.to_string();
-    if prover.has_image(&image_id_str).await? {
-        tracing::debug!(
-            "Skipping program upload for cached image ID: {image_id_str} for request {:x}",
-            request.id
-        );
-        return Ok(image_id_str);
+    let image_id_str = request.requirements.image_id().map(|image_id| image_id.to_string());
+    // When predicate is ClaimDigestMatch, we do not have the image id, so we must always download and upload the image.
```
capossele, line 357: This is true and worth thinking if this might create a (rate limit) issue when used often. One solution could be to make sure the requestor knows the image url must guarantee some high frequency access, or we should think to cache based on the url, however that is risky as the content might change. Anyway, just food for thoughts for now

crates/indexer/migrations/3_proof_requests.sql
```diff
@@ -4,7 +4,6 @@ CREATE TABLE IF NOT EXISTS proof_requests (
     client_address      TEXT      NOT NULL,  -- Ethereum address
     
     -- Requirements fields
-    image_id            TEXT      NOT NULL, -- 32-byte image ID (hex encoded)
```
capossele, line 7: I'm in favor of this approach of keeping things clean. However this will require nuking the existing db as the migration will not be compatible. We could leave the value and use a dummy value as an alternative if we prefer keeping compatibility

crates/guest/assessor/assessor-guest/src/main.rs
```diff
@@ -64,8 +64,7 @@ fn main() {
             fill.verify_signature(&eip_domain_separator).expect("signature does not verify")
         };
         fill.evaluate_requirements().expect("requirements not met");
-        env::verify_integrity(&fill.receipt_claim()).expect("claim integrity check failed");
```
capossele, line 67: Doesn't removing this means we now cannot guarantee that the proof is generated with the actual input and not something else?

capossele: or maybe you are right, and that is guaranteed by recomputing the claim digest in here already

contracts/src/types/Fulfillment.sol
```diff
@@ -15,16 +16,13 @@ struct Fulfillment {
...
-    // TODO: Add a flag in the request to decide whether to post the journal. Note that
-    // if the journal and journal digest do not need to be delivered to the client, imageId will
-    // be replaced with claim digest, since it is captured in the requirements on the request,
-    // checked by the Assessor guest.
-    /// @notice Journal committed by the guest program execution.
-    /// @dev The journal is checked to satisfy the predicate specified on the request's requirements.
-    bytes journal;
+    /// @notice The `PredicateType` of the request that is being fulfilled.
+    /// @dev When the `PredicateType` is `ClaimDigestMatch`, there callbacks are not supported
+    PredicateType predicateType;
```
nategraf: Related to my other comments, I think we can remove this.

contracts/src/types/Callback.sol
```diff
@@ -6,23 +6,27 @@ pragma solidity ^0.8.24;
...
+
 /// @title Callback Struct and Library
 /// @notice Represents a callback configuration for proof delivery
 struct Callback {
     /// @notice The address of the contract to call back
     address addr;
     /// @notice Maximum gas to use for the callback
     uint96 gasLimit;
+    /// @notice The type of callback
+    CallbackType callbackType;
```
nategraf: Should we maybe not support multiple callback types? I am leaning towards saying that we should not, or least not support multiple callback types right now. Maybe this will be requested in the future, but its unclear what the receiver experience would look like and I'm inclined to think we probably won't support it.

contracts/src/types/Predicate.sol
nategraf: I think the way to clarify this is that "a predicate is a function over the claim that determines whether it meets the clients requirements". Prior to this PR, predicates acted over the journal only.

nategraf: The names `DigestMatch` and `PrefixMatch` are a bit confusing with this framing, in that they are with reference to the journal. We could rename when as `JournalDigestMatch` and `JournalPrefixMatch`, although I'm personally ok with keeping them named as is and just adding a comment to each to describe the fact that the apply to the journal. This might be the better option since we want the common-case patterns to have the shorter, friendlier names.

contracts/src/types/Predicate.sol
```diff
@@ -26,36 +34,66 @@ library PredicateLibrary {
     /// @notice Creates a digest match predicate.
     /// @param hash The hash to match.
     /// @return A Predicate struct with type DigestMatch and the provided hash.
-    function createDigestMatchPredicate(bytes32 hash) internal pure returns (Predicate memory) {
-        return Predicate({predicateType: PredicateType.DigestMatch, data: abi.encode(hash)});
+
```
nategraf: This newline is separating the function from its comment.
```suggestion
```

contracts/src/types/Predicate.sol
```diff
@@ -26,36 +34,66 @@ library PredicateLibrary {
...
+            return startsWith(journal, dataJournal);
+        } else if (predicate.predicateType == PredicateType.ClaimDigestMatch) {
+            return bytes32(predicate.data) == ReceiptClaimLib.ok(imageId, journalDigest).digest();
         } else {
             revert("Unreachable code");
         }
     }
 
+    /// Taken from openzepplin util Bytes.sol
+    function sliceToEnd(bytes memory buffer, uint256 start) internal pure returns (bytes memory) {
```
nategraf: Can you move this out of the library namespace?

contracts/src/types/AssessorCallback.sol
```diff
@@ -4,11 +4,15 @@
...
+
 struct AssessorCallback {
     /// @notice The index of the fill in the request
     uint16 index;
     /// @notice The address of the contract to call back
     address addr;
     /// @notice Maximum gas to use for the callback
     uint96 gasLimit;
+    /// @notice The type of callback
+    CallbackType callbackType;
```
nategraf: Instead of just the callback type, we should encode the full callback data here, and ensure it is committed in the assessor journal. This both allows us to skip the recomputation of the receipt claim onchain when we go to send the callback, because it will now be part of what the assessor constructs, and allows for more flexibility in the future in how we evolve the system.

contracts/src/types/Callback.sol
nategraf: Another way to keep the option open for multiple callback types would be to add a version number to the start of the encoded `CallbackData`. So if we increment the version in the future, we could provide a new function for decoding it.

But we don't necessarily need to do that even: because the first 32 bytes are a hash, we can use a special value of `bytes32(uint256(1))` to indicate that this is the new version in the future. This is more ugly, but it can be the plan of record in the eventuality that we do want to support multiple types of callbacks in the future.

contracts/src/types/AssessorJournal.sol
```diff
@@ -17,6 +18,8 @@ struct AssessorJournal {
     /// @notice The (optional) selectors for the requests committed by the assessor.
     /// @dev This is used to verify the fulfillment of the request against its selector's seal.
     Selector[] selectors;
+    /// @notice The list of `PredicateType` for each request.
+    PredicateType[] predicateTypes;
```
nategraf: I don't think we need this.

Also, if we do this should be in `AssessorCommitment` instead. The reason the `callbacks` and `selectors` are outside this Merkle tree of `AssessorCommitment` objects is that most of them will be empty and so it takes up less space to do it this way.

contracts/src/types/Callback.sol
nategraf: Another option for forward compatibility is that we can increment the entire request version, and accordingly create any need code paths we need, switching on that. We have reserved bits in the request ID that can be used for this.

contracts/src/types/AssessorCallback.sol
nategraf: Related to my comments below, I am kind of thinking we should just embed the `imageId` and `journal` directly here to simplify. https://github.com/boundless-xyz/boundless/pull/1004/files#r2277886432

contracts/src/types/Fulfillment.sol
```diff
@@ -15,16 +16,13 @@ struct Fulfillment {
...
+    /// @notice Claim Digest
+    bytes32 claimDigest;
+    /// @notice The callback data, if requested.
+    bytes callbackData;
```
nategraf: Related to my other comments, this should come with the `AssessorReceipt`

contracts/src/BoundlessMarket.sol
```diff
@@ -333,13 +337,39 @@ contract BoundlessMarket is
...
+                (bytes32 imageId, bytes calldata journal) = _decodeCallbackData(fill.callbackData);
+                bytes32 calculatedClaimDigest = ReceiptClaimLib.ok(imageId, sha256(journal)).digest();
+                if (fill.claimDigest != calculatedClaimDigest) {
+                    revert ClaimDigestMismatch(fill.claimDigest, calculatedClaimDigest);
+                }
+
+                if (callbackIndexPlusOne > 0) {
+                    AssessorCallback calldata callback = assessorReceipt.callbacks[callbackIndexPlusOne - 1];
+                    _executeCallback(fill.id, callback.addr, callback.gasLimit, imageId, journal, fill.seal);
+                }
```
nategraf: By placing the callback data into the `AssessorCallback`, we can mostly eliminate the new validation here, and simply decode and execute the callback (reverting if decode fails). If we directly embed the `CallbackData` into `AssessorCallback`, we can avoid the extra decode step as well.

contracts/src/types/Fulfillment.sol
```diff
@@ -15,16 +16,13 @@ struct Fulfillment {
...
-    /// @notice Image ID of the guest that was verifiably executed to satisfy the request.
-    /// @dev Must match the value in the request's requirements.
-    bytes32 imageId;
-    // TODO: Add a flag in the request to decide whether to post the journal. Note that
-    // if the journal and journal digest do not need to be delivered to the client, imageId will
-    // be replaced with claim digest, since it is captured in the requirements on the request,
-    // checked by the Assessor guest.
-    /// @notice Journal committed by the guest program execution.
-    /// @dev The journal is checked to satisfy the predicate specified on the request's requirements.
-    bytes journal;
```
nategraf, line 27: Ultimately, I think removing the journal from the `Fulfillment` and only providing it in the callback is the right call. It will have the effect though that, by default, journals will not have onchain data availability guarantees (in the `ProofDelivered`) event. Most applications we have seen so far will have computed the journal locally (as they must to use the `DigestMatch` predicate) so this should not need needed, however, it may add some friction to the devex if they were planning on fetching the receipt from the chain and have the journal included.

crates/assessor/src/lib.rs
```diff
@@ -77,15 +83,28 @@ impl Fulfillment {
...
     pub fn evaluate_requirements(&self) -> Result<(), Error> {
-        if !self.request.requirements.predicate.eval(&self.journal) {
-            return Err(Error::PredicateEvaluationError);
+        let requirements = &self.request.requirements;
+
+        if requirements.predicate.predicateType == PredicateType::ClaimDigestMatch
+            && requirements.callback.callbackType != CallbackType::None
+        {
+            return Err(Error::RequirementsEvaluationError);
+        }
```
nategraf: I would drop this constraint (partly because my suggestion is to drop `CallbackType`) and instead fail when trying to construct the `AssessorCallback` is the `FullfillmentClaimData` does not contain the image ID and journal. It may need somewhat odd that this would enable a `ClaimDigestMatch` request to actually have a callback, but this is actually fine so long as we check that `ReceiptClaim::ok(image_id, journal).digest()` does in fact meet the requirements, which we do.

crates/guest/assessor/assessor-guest/src/main.rs
nategraf: My assessment is that this does allow the `assessor` to be run on claims that are not in fact true, but it should not be an issue since we check the seal onchain anyway. We need to check the seal onchain for data availability reasons, and because the onchain verifier is actually the canonical one.

crates/indexer/migrations/3_proof_requests.sql
nategraf: What is the consequence of nuking the indexer DB? I don't have a strong opinion of whether or not to keep this field to maintain compatibility. I do feel like we want to start indexing claim digests and callback data though.

contracts/src/types/Callback.sol
ec2: I think it makes sense to just not support other callback types for now then for simplicity

contracts/src/types/Fulfillment.sol
ec2: That means we would only provide the journal if a callback is requested, no?

contracts/src/types/AssessorReceipt.sol
```diff
@@ -5,6 +5,7 @@
 pragma solidity ^0.8.20;
 
 import {AssessorCallback} from "./AssessorCallback.sol";
+
```
nategraf:
```suggestion
```

contracts/src/types/Requirements.sol
```diff
@@ -29,10 +30,10 @@ library RequirementsLibrary {
         return keccak256(
             abi.encode(
                 REQUIREMENTS_TYPEHASH,
-                requirements.imageId,
                 CallbackLibrary.eip712Digest(requirements.callback),
                 PredicateLibrary.eip712Digest(requirements.predicate),
-                requirements.selector
+                requirements.selector,
+                requirements.fulfillmentDataType
```
nategraf: I would rather not include this here, for two reasons:
1. It can conflict with the chosen predicate. In particular, using the claim digest predicate may is potentially incompatible with the `ImageIdAndJournal` fulfillment type. So the request would then be internally inconsistent.
2. It adds another parameter for the user to set.

Instead, I would recommend that we, in the assessor, use the fulfillment type that is provided by the prover. In the case of the two journal-checking predicates, this will force the prover to use `ImageIdAndJournal`.

crates/assessor/src/lib.rs
```diff
@@ -77,15 +83,26 @@ impl Fulfillment {
     }
     /// Evaluates the requirements of the request.
     pub fn evaluate_requirements(&self) -> Result<(), Error> {
-        if !self.request.requirements.predicate.eval(&self.journal) {
-            return Err(Error::PredicateEvaluationError);
+        // If ClaimDigestMatch, a journal cannot be provided because it would be unautheticatable.
+        if self.request.requirements.predicate.predicateType == PredicateType::ClaimDigestMatch {
+            if self.request.requirements.fulfillmentDataType != FulfillmentDataType::None {
+                return Err(Error::RequirementsEvaluationError);
+            }
```
nategraf: We actually can check it, by using `ReceiptClaim::ok`. With this, you can simply always call `eval`, like how this is on `main`

crates/guest/assessor/assessor-guest/src/main.rs
```diff
@@ -40,20 +37,23 @@ fn main() {
...
+    let mut predicate_types = Vec::with_capacity(input.fills.len());
 
     let eip_domain_separator = input.domain.alloy_struct();
     // For each fill we
     // - verify the request's signature
     // - evaluate the request's requirements
-    // - verify the integrity of its claim
     // - record the callback if it exists
     // - record the selector if it is present
+    // NOTE: We no longer verify integrity of the claim. That is done on chain.
```
nategraf: "no longer" will not make sense to anyone who does not already have context on the evolution of this code.
```suggestion
    // NOTE: We do not verify integrity of the claim. That is done on chain.
```

contracts/src/BoundlessMarket.sol
```diff
@@ -333,13 +337,50 @@ contract BoundlessMarket is
...
-                AssessorCallback calldata callback = assessorReceipt.callbacks[callbackIndexPlusOne - 1];
-                _executeCallback(fill.id, callback.addr, callback.gasLimit, fill.imageId, fill.journal, fill.seal);
+
+            if (fill.fulfillmentDataType == FulfillmentDataType.ImageIdAndJournal) {
+                (bytes32 imageId, bytes calldata journal) = _decodeFulfillmentData(fill.fulfillmentData);
+                bytes32 calculatedClaimDigest = ReceiptClaimLib.ok(imageId, sha256(journal)).digest();
+
+                if (fill.claimDigest != calculatedClaimDigest) {
+                    revert ClaimDigestMismatch(fill.claimDigest, calculatedClaimDigest);
+                }
```
nategraf: We don't need this check, as the assessor should have checked it.

contracts/src/types/FulfillmentData.sol
```diff
@@ -0,0 +1,42 @@
...
+/// @notice Represents a fulfillment configuration for proof delivery
+struct FulfillmentData {
+    /// @notice Image ID of the guest that was verifiably executed to satisfy the request.
+    bytes32 imageId;
+    /// @notice Journal committed by the guest program execution.
+    /// @dev The journal is checked to satisfy the predicate specified on the request's requirements.
+    bytes journal;
+}
+
+library FulfillmentDataLibrary {
```
willpote, line 23: Can we add a test file for FulfillmentData?

contracts/src/types/Predicate.sol
```diff
@@ -5,18 +5,26 @@
...
 using PredicateLibrary for Predicate global;
+using ReceiptClaimLib for ReceiptClaim;
 
 /// @title Predicate Struct and Library
-/// @notice Represents a predicate and provides functions to create and evaluate predicates.
+/// @notice A predicate is a function over the claim that determines whether it meets the clients requirements.
+/// The data field is used to store the specific data associated with the predicate.
+/// - DigestMatch: (bytes32, bytes32) -> abi.encodePacked(imageId, journalHash)
+/// - PrefixMatch: (bytes32, bytes) -> abi.encodePacked(imageId, prefix)
+/// - ClaimDigestMatch: (bytes32) -> abi.encode(claimDigest)
```
willpote, line 19: Nit: can we create library functions for decoding? Then we can use them in the smart contract requestor example

contracts/src/BoundlessMarket.sol
```diff
@@ -333,14 +338,49 @@ contract BoundlessMarket is
...
+            // Extract imageId (first 32 bytes after length)
+            imageId := calldataload(add(data.offset, 0x20))
+            // Extract journal offset and create calldata slice
+            let journalOffset := calldataload(add(data.offset, 0x40))
+            let journalPtr := add(data.offset, add(0x20, journalOffset))
+            let journalLength := calldataload(journalPtr)
+            journal.offset := add(journalPtr, 0x20)
+            journal.length := journalLength
+        }
+    }
```
willpote: Can't we use the library function?

crates/boundless-market/src/request_builder/finalizer.rs
```diff
@@ -130,8 +129,15 @@ impl Adapt<Finalizer> for RequestParams {
 
         // As an extra consistency check. verify the journal satisfies the required predicate.
         if let Some(ref journal) = self.journal {
-            if !requirements.predicate.eval(journal) {
-                bail!("journal in request builder does not match requirements predicate; check request parameters.\npredicate = {:?}\njournal = 0x{}", requirements.predicate, hex::encode(journal));
+            if let Some(image_id) = self.image_id {
```
willpote: We should probably check that they aren't requesting a callback + claimDigestMatch also. Not certain if it should live here exactly or elsewhere within the `request_builder` framework.

contracts/src/BoundlessMarket.sol
```diff
@@ -333,14 +338,49 @@ contract BoundlessMarket is
             }
 
             uint256 callbackIndexPlusOne = fillToCallbackIndexPlusOne[i];
-            if (callbackIndexPlusOne > 0) {
-                AssessorCallback calldata callback = assessorReceipt.callbacks[callbackIndexPlusOne - 1];
-                _executeCallback(fill.id, callback.addr, callback.gasLimit, fill.imageId, fill.journal, fill.seal);
+
+            if (fill.fulfillmentDataType == FulfillmentDataType.ImageIdAndJournal) {
```
willpote: I may be missing something, but I don't see how we ensure that the Fulfillment data type is as the proof requestor desires. E.g. if I want a callback, how do i ensure that during fulfillment the prover must set 1FulfillmentDataType1 == 1ImageIdAndJournal1?

crates/boundless-cli/src/bin/boundless.rs
```diff
@@ -661,29 +661,60 @@ async fn handle_request_command(cmd: &RequestCommands, client: StandardClient) -
...
             tracing::info!("Fetching proof for request 0x{:x}", request_id);
-            let (journal, seal) =
+            let (callback_data, seal) =
                 client.boundless_market.get_request_fulfillment(*request_id).await?;
             tracing::info!("Successfully retrieved proof for request 0x{:x}", request_id);
             tracing::info!(
-                "Journal: {} - Seal: {}",
-                serde_json::to_string_pretty(&journal)?,
+                "Callback Data: {} - Seal: {}",
+                serde_json::to_string_pretty(&callback_data)?,
```
willpote: Think this is out of date now? Seems users no longer need to use callbacks due to `FulfillmentDataType` existing on the fill

crates/broker/src/tests/e2e.rs
```diff
@@ -17,14 +17,18 @@ use std::{future::Future, path::PathBuf};
 use crate::{config::Config, now_timestamp, Args, Broker};
 use alloy::{
     node_bindings::Anvil,
```
willpote, line 19: Not seeing a huge amount in terms of broker tests. Would be good to have an E2E test that covers requesting claim digest match, as well as unit tests where appropriate

contracts/src/BoundlessMarket.sol
ec2: I've added a check for if FulfillmentDataType is none, and a callback is requested, to throw

contracts/src/BoundlessMarket.sol
ec2: youre right, removed check.

contracts/src/types/AssessorJournal.sol
ec2: Removed it entirely, youre correct, we dont need it

crates/broker/src/tests/e2e.rs
ec2: Yeah i def need to add more testing... Was trying to nail down smart contract changes first because changing those cascades to everywhere.

contracts/src/BoundlessMarket.sol
ec2: moved this function into library 👍🏼

contracts/src/BoundlessMarket.sol
willpote: Hmm, did we consider making FulfillmentDataType explicitly specified by the requestor as part of the `ProofRequest`, and committing to it in the Assessor Journal?

Currently its kind of implicitly specified by the use of callback whether you want the journal, which works, but imo made more sense when we had the super minimal changes approach which didn't have the new `FulfillmentDataType`. 

Now that we have it, i'm wondering if it would make sense to be fully explicit and just add it to ProofRequest also, than rely on the callback.

willpote: Also, currently it seems the only effect of `FulfillmentDataType` is whether a callback would be executed or not. One big part of the fulfillment of a proof request is emitting the ProofDelivered event, which currently emits the full journal. If FulfillmentDataType == NONE, then I think we should skip emitting the journal/image as part of the ProofDelivered event (and only emit the request id).

contracts/src/BoundlessMarket.sol
```diff
@@ -333,9 +334,26 @@ contract BoundlessMarket is
...
+                (bytes32 imageId, bytes calldata journal) =
+                    FulfillmentDataLibrary.decodeFulfillmentData(fill.fulfillmentData);
```
nategraf: The way this is factored, this decode always occurs even when a callback was not requested. If move the `callbackIndexPlusOne` check to be on the outside, we can avoid this. I also think it is more intuitive that way.

contracts/src/BoundlessMarket.sol
```diff
@@ -333,9 +334,26 @@ contract BoundlessMarket is
...
+                // We should only get to this point if the journal has been authenticated because the claimDigest
+                // passed the integrity check in verifyDelivery and we just recalculated it from the imageId and journal
+                // and compared to make sure they match.
+                // That is to say, if the claimDigest isn't a Risc0 zkvm claimDigest (e.g. blake3), journals
+                // cannot be authenticated and therefore we cannot use them in a callback and we revert with ClaimDigestMismatch.
```
nategraf: This is not accurate, since the digest is not checked in `verifyDelivery`. I feel like I might have contributed to the confusion here over the iterations. Right now, the code is missing a key check that the `fill.claimDigest` actually matches the fulfillment data.

There are two options here:
1. Verify that `fill.fulfillmentData` matches `fill.claimDigest` onchain by computing the claim digest again.
2. Add a hash of the fulfillment data to the `AssessorCommit`.

After some thinking, I favor #2. If we use `keccak` to has the `fulfillmentData` to include in the `AsessorCommit`, we can save a decent amount of gas onchain. Additionally, this approach provides a path supporting callbacks for BLAKE3 hashed claims in the future without needing BLAKE3 in solidity. It does mean that if the assessor is broken then we an attacker could cause this to mismatch. This would be caught as long as the user of the proof is additionally verifying it themselves, which is highly recommended.

nategraf: Can you make this comment a bit more clear and concise? I had to read it a couple of times to understand what you mean.

ec2: I see the error in my ways. Option 2. makes sense to me. Should it actually be in the AssessorReceipt though?

nategraf: We use the `AssessorReceipt` to transmit the data that is required for verification, but is not part of the fulfillment. Since the `fulfillmentData` is in `Fulfillment`, it does not need to be in `AssessorReceipt`. The key is that it is committed in the `AssesorJournal`. Including it in `AssessorCommit` is the right place, because each assessor commit is bound to a single fulfillment, and then the list of commits is combined to form the batch root that is placed into `AssessorJournal.root`.

ec2: I think my comments there are a bit outdated and irrelevant now, so imma just remove them.

contracts/src/BoundlessMarket.sol
ec2: I think adding `FulfillmentDataType` for the requestor makes things potentially messier as the requestor has more ways to craft unfulfillable requests unless you add a bunch more checks. 

I like the current implicit approach because it allows for provers to potentially fulfill requests in more than one way which opens up flexibility in the future.

ec2: The `ProofDelivered` event emits a `Fulfillment` not the journal. So if it is the case where FulfillmentDataType is None, then the `fulfillmentData` bytes field should be empty.

crates/guest/assessor/assessor-guest/src/main.rs
```diff
@@ -64,23 +61,30 @@ fn main() {
...
             index: U256::from(index),
             id: fill.request.id,
             requestDigest: request_digest.into(),
             claimDigest: <[u8; 32]>::from(claim_digest).into(),
+            fulfillmentDataDigest: <[u8; 32]>::from(fill.fulfillment_data_digest()).into(),
         }
         .eip712_hash_struct();
         leaves.push(Digest::from_bytes(*commit));
+
+        let callback = &fill.request.requirements.callback;
```
willpote, line 79: On L80 can also use

contracts/src/BoundlessMarket.sol
willpote: > I like the current implicit approach because it allows for provers to potentially fulfill requests in more than one way which opens up flexibility in the future.

What sort of thing do you have in mind here? I guess my thinking is that rational provers will always just fulfill requests in the gas-cheapest way possible to maximize margins (which I guess is typically "None" type). Curious about a case where provers would want to fulfill in more than one way

@nategraf also curious for your thoughts here

nategraf: @willpote , take a look at the latest version of this code. It's somewhat less flexible in that if a callback is requested, it must always be delivered. Additionally, the system is now bound such that whatever fulfillment data the assessor is given must be delivered onchain. So the assessor controls what is allowable, with the additional restriction that if a callback is requested the prover _must_ deliver the image ID and journal so that the callback can proceed. I agree that the provers will take the cheapest route available to them. I view the flexibility in terms A) separation of concerns, to enforce each check in the place it makes the most sense and B) how hard will it be to change the system later.

contracts/src/types/AssessorJournal.sol
```diff
@@ -5,6 +5,7 @@
 pragma solidity ^0.8.20;
 
 import {AssessorCallback} from "./AssessorCallback.sol";
+import {PredicateType} from "./Predicate.sol";
```
nategraf: Unused import
```suggestion
```

contracts/src/types/FulfillmentData.sol
```diff
@@ -0,0 +1,62 @@
...
+using FulfillmentDataLibrary for FulfillmentData global;
+
+enum FulfillmentDataType {
+    None,
+    ImageIdAndJournal
+}
+
+/// @title FulfillmentData Struct and Library
+/// @notice Represents a fulfillment configuration for proof delivery
+struct FulfillmentData {
```
nategraf: We should probably call this something like `FulfillmentDataImageIdAndJournal` to disambiguate the fact that it is one of potentially multiple types of fulfillment data.

contracts/src/types/FulfillmentData.sol
```diff
@@ -0,0 +1,62 @@
...
+    /// @notice Image ID of the guest that was verifiably executed to satisfy the request.
+    bytes32 imageId;
+    /// @notice Journal committed by the guest program execution.
+    /// @dev The journal is checked to satisfy the predicate specified on the request's requirements.
```
nategraf: This comment is slightly misleading in that the predicate now acts over the entire claim, rather than just the journal.
```suggestion
```

contracts/src/types/FulfillmentData.sol
```diff
@@ -0,0 +1,62 @@
...
+    bytes journal;
+}
+
+library FulfillmentDataLibrary {
+    /// @notice Decodes a bytes calldata into a FulfillmentData struct.
+    /// @param data The bytes calldata to decode.
+    /// @return fillData The decoded FulfillmentData struct.
+    function decode(bytes calldata data) public pure returns (FulfillmentData memory fillData) {
```
nategraf: This function returns the struct in memory (rather than `calldata`) anyway, so should this just be  an ABI decode? If so, it seems like it can at least use `decodeFulfillmentData` to avoid duplicating this assembly block. I'm also not personally familiar with any safety considerations here.  It does feel like someone could send malformed to make this go wonky, but it's not clear to me whether it would have nay value to them to do so.

contracts/src/types/Fulfillment.sol
```diff
@@ -15,19 +16,28 @@ struct Fulfillment {
...
+    /// @return The keccak256 digest of the fulfillmentData.
+    function fulfillmentDataDigest(Fulfillment memory fulfillment) internal pure returns (bytes32) {
+        if (fulfillment.fulfillmentDataType == FulfillmentDataType.None) {
+            return bytes32(0);
+        } else if (fulfillment.fulfillmentDataType == FulfillmentDataType.ImageIdAndJournal) {
+            return keccak256(abi.encodePacked(uint8(fulfillment.fulfillmentDataType), fulfillment.fulfillmentData));
+        } else {
+            revert("Unknown fulfillment data type");
+        }
+    }
```
nategraf: We do need to constrain that the `fulfillmentData` is empty if `fulfillmentDataType` is `None`. One way to do this would be to always hash here instead. This would lead to the simple implementation of.

```suggestion
    function fulfillmentDataDigest(Fulfillment memory fulfillment) internal pure returns (bytes32) {
            return keccak256(abi.encodePacked(uint8(fulfillment.fulfillmentDataType);
        }
    }
```

Another option would be to check the length:


```suggestion
    function fulfillmentDataDigest(Fulfillment memory fulfillment) internal pure returns (bytes32) {
        if (fulfillment.fulfillmentDataType == FulfillmentDataType.None) {
            require(fulfillment.fulfillmentData.length == 0, InvalidFulfillmentData());
            return bytes32(0);
        } else if (fulfillment.fulfillmentDataType == FulfillmentDataType.ImageIdAndJournal) {
            return keccak256(abi.encodePacked(uint8(fulfillment.fulfillmentDataType), fulfillment.fulfillmentData));
        } else {
            revert("Unknown fulfillment data type");
        }
    }
```

I'm happy with either, so I'll let you pick the direction.

contracts/src/IBoundlessMarket.sol
```diff
@@ -165,6 +165,10 @@ interface IBoundlessMarket {
     /// @dev selector efc954a6
     error BatchSizeExceedsLimit(uint256 batchSize, uint256 limit);
 
+    /// @notice Error when the fulfillment has a unfulfillable callback
+    /// TODO(ec2): selector
```
nategraf: You can use the following command to get this value:
```
forge inspect IBoundlessMarket errors
```

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -430,9 +437,23 @@ impl ProofRequest {
...
         Url::parse(&self.imageUrl).map(|_| ())?;
 
-        if self.requirements.imageId == B256::default() {
-            return Err(RequestError::ImageIdIsZero);
+        match self.requirements.predicate.predicateType {
+            PredicateType::DigestMatch | PredicateType::PrefixMatch => {
+                if self.requirements.predicate.data.len() < 32 {
+                    return Err(RequestError::MalformedPredicateData);
+                }
+                if self.requirements.image_id() == Some(<[u8; 32]>::from(B256::default()).into()) {
```
nategraf:
```suggestion
                if self.requirements.image_id() == Some(Digest::ZERO) {
```

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -430,9 +437,23 @@ impl ProofRequest {
         }
         Url::parse(&self.imageUrl).map(|_| ())?;
 
-        if self.requirements.imageId == B256::default() {
-            return Err(RequestError::ImageIdIsZero);
+        match self.requirements.predicate.predicateType {
+            PredicateType::DigestMatch | PredicateType::PrefixMatch => {
+                if self.requirements.predicate.data.len() < 32 {
```
nategraf: The `DigestMatch` predicate can be restricted further to ensure the data length is exactly `64` bytes. But perhaps this should attempt to decode the data, and return an error if that fails to avoid including these magic constants here.

crates/boundless-market/src/contracts/mod.rs
nategraf: Also, what happens if this returns `None`? We should probably error on that case.

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -510,18 +536,8 @@ impl ProofRequest {
...
-            callback: Callback::default(),
-            selector: UNSPECIFIED_SELECTOR,
-        }
-    }
-
-    /// Sets the image ID.
-    pub fn with_image_id(self, image_id: impl Into<Digest>) -> Self {
-        Self { imageId: <[u8; 32]>::from(image_id.into()).into(), ..self }
+    pub fn new(predicate: Predicate) -> Self {
+        Self { predicate, callback: Callback::default(), selector: UNSPECIFIED_SELECTOR }
```
nategraf: This is a useful Rust-ism that will allow us to make this API a bit more flexible / less verbose at the call site.
```suggestion
    pub fn new(predicate: impl Into<Predicate>) -> Self {
        Self { predicate: predicate.into(), callback: Callback::default(), selector: UNSPECIFIED_SELECTOR }
```

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -551,22 +567,165 @@ impl Requirements {
             false => Self { selector: FixedBytes::from(Selector::Groth16V2_2 as u32), ..self },
         }
     }
+
+    /// Returns image id from the predicate type. Returns none if claim digest match. Panics if
+    /// the predicate data is not long enough.
```
nategraf: It doesn't seem like this API should ever panic. One thing you could do if you want to differentiate "malformed data" vs "claim digest predicate" is return `Result` and include a specific error for "this is a claim digest match predicate"

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -551,22 +567,165 @@ impl Requirements {
...
+    pub fn image_id(&self) -> Option<Digest> {
+        self.predicate.image_id()
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+/// The data that is used to construct the claim for a fulfillment.
+pub enum FulfillmentClaimData {
```
nategraf: It would be nice for this to just be `FulfillmentData`
```suggestion
pub enum FulfillmentData {
```

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -551,22 +567,165 @@ impl Requirements {
...
+    /// There are certain types of fulfillments where journals cannot be authenticated and
+    /// so we can only check the integrity of the receipt.
```
nategraf: I don't think this comment would make sense to an outsider.
```suggestion
    /// Fulfillment data including on the claim digest.
    ///
    /// This variant is used when the journal does not need to be delivered onchain (e.g. a large journal that would be expensive to deliver), or when the claim type is not `ReceiptClaim` (e.g. BLAKE3 hashed claims).
```

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -551,22 +567,165 @@ impl Requirements {
...
+
+    /// Returns image id from the predicate type. Returns none if claim digest match. Panics if
+    /// the predicate data is not long enough.
+    pub fn image_id(&self) -> Option<Digest> {
+        self.predicate.image_id()
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+/// The data that is used to construct the claim for a fulfillment.
```
nategraf: Please put your comments above annotations.
```suggestion
/// The data that is used to construct the claim for a fulfillment.
#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
```

crates/boundless-market/src/contracts/mod.rs
nategraf: Also, do we actually use this variant? It seems like this variant should be empty and called `None` to match the contract.

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -551,22 +567,165 @@ impl Requirements {
...
+    }
+    /// Get the claim digest or calculates it if journal and image id exists.
+    pub fn claim_digest(&self) -> Option<Digest> {
+        match self {
+            Self::ClaimDigest(digest) => Some(*digest),
+            Self::ImageIdAndJournal(image_id, journal) => {
+                Some(ReceiptClaim::ok(*image_id, journal.as_ref().to_vec()).digest())
+            }
+        }
+    }
```
nategraf: Instead of these, I would provide a `from_bytes` method that returns `Self` and the caller can `match` on it to get the underlying values.

crates/boundless-market/src/contracts/mod.rs
nategraf: This is important if we ever want to add new variants in the future
```suggestion
#[non_exhaustive]
pub enum FulfillmentClaimData {
```

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -551,22 +567,165 @@ impl Requirements {
...
+
+    /// Returns claim digest if the predicate type is ClaimDigestMatch.
+    pub fn claim_digest(&self) -> Option<Digest> {
+        if self.predicateType == PredicateType::ClaimDigestMatch {
+            Some(Digest::from_bytes(self.data.as_ref().try_into().unwrap()))
+        } else {
+            None
+        }
+    }
```
nategraf: Similar to my comment above, we might want to prefer uses a decode method.

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -551,22 +567,165 @@ impl Requirements {
...
+
+    #[inline]
```
nategraf: Why `#[inline]`?
```suggestion
```

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -551,22 +567,165 @@ impl Requirements {
...
+                self.data.as_ref()
+                    == fulfillment_data
+                        .claim_digest()
+                        .expect(
+                            "fulfillment data doesnt have claim digest, but is claim digest match",
+                        )
+                        .as_bytes()
+            }
+            PredicateType::__Invalid => panic!("invalid PredicateType"),
+        }
```
nategraf: Could you implement a predicate enum similar to fulfillment data and decode into that and match instead?

crates/boundless-market/src/request_builder/finalizer.rs
```diff
@@ -128,10 +127,18 @@ impl Adapt<Finalizer> for RequestParams {
             .context("failed to build request: offer is incomplete")?;
         let request_id = self.require_request_id().context("failed to build request")?.clone();
 
-        // As an extra consistency check. verify the journal satisfies the required predicate.
-        if let Some(ref journal) = self.journal {
-            if !requirements.predicate.eval(journal) {
-                bail!("journal in request builder does not match requirements predicate; check request parameters.\npredicate = {:?}\njournal = 0x{}", requirements.predicate, hex::encode(journal));
+        // If a callback is requested, we should check that the journal satisfies the required predicate.
+        if !requirements.callback.is_none() {
```
nategraf: We also want to check the predicate when the type is ClaimDigestMatch, and we always want to check the predicate when possible, regardless of whether a callback is requested.

crates/boundless-cli/src/bin/boundless.rs
```diff
@@ -687,29 +688,60 @@ async fn handle_request_command(cmd: &RequestCommands, client: StandardClient) -
...
+                        .await
+                        .map_err(|_| anyhow::anyhow!("Verification failed"))?;
+                }
+                _ => {
+                    bail!(
+                        "Unsupported predicate type for verification: {:?}",
+                        predicate.predicateType
+                    );
+                }
+            }
```
nategraf: This block is too large to do what should be quite simple. Could you move some of these details into the library so that this can be more concise?

crates/guest/assessor/assessor-guest/src/main.rs
nategraf: This is a very important invariant change that we will want to bring to the auditor if we get someone to review it. My understanding that it should be fine, as we need to check the delivered seal onchain anyway. But perhaps there are implications I am not seeing.

crates/guest/assessor/assessor-guest/src/main.rs
```diff
@@ -64,23 +61,30 @@ fn main() {
...
+
+        // Note that this allows for callbacks to be specified even in the case of
+        // ClaimDigestMatch predicates where the journal cannot be provided.
+        // This is ok because it will fail on chain checks if a journal is indeed provided.
```
nategraf: This is not quite accurate, because it _is_ possible for a prover to provide the image ID and journal fulfillment data in response to a `ClaimDigestMatch` request. In this case, if a callback was requested as well, it will be delivered. This is all correct behavior, even though we don't expect provers to actually do this.

crates/assessor/src/lib.rs
```diff
@@ -77,15 +83,40 @@ impl Fulfillment {
...
-        ReceiptClaim::ok(image_id, self.journal.clone())
+
+    /// Returns the claim digest for the fulfillment.
+    pub fn claim_digest(&self) -> Result<Digest, Error> {
+        match self.fulfillment_data {
+            FulfillmentClaimData::ClaimDigest(digest) => Ok(digest),
+            FulfillmentClaimData::ImageIdAndJournal(image_id, ref journal) => {
+                Ok(ReceiptClaim::ok(image_id, <Vec<u8>>::from(journal.clone())).digest())
+            }
+        }
```
nategraf: Since the FulfillmentData onchain uses `Nonce` and `ImageIdAndJournal`, I think its best we do the same here. What that means for this function is that it would take the claim digest from the requirements rather than the `FulfillmentData` in the case of `FulfillmentData::None`. In the case of the prefix match predicate, the claim digest can only be calculated from the fulfillment data. So there is an akward situation where the source of this data is not consistent. This kind of inconsistency, especially when there are two possible sources that may disagree is prolific ground for bugs.

What I would suggest is to combine this method with `evaluate_requirements` by having that function return the claim digest, after it has checked that everything is consistent. That way there is only one way to get the claim digest, and it ensures consistency.

crates/boundless-market/src/contracts/mod.rs
ec2: It should never be none here since the previous check looks at if the predicate data len is < 32

contracts/src/types/Fulfillment.sol
ec2: i chose option 1

contracts/src/types/FulfillmentData.sol
ec2: I changed it to call the other function. I don't know for certain if the assembly block is ok or not. But theres no other way that I know of to avoid copying the calldata into memory for the journal

contracts/src/types/FulfillmentData.sol
ec2: I've changed it to what you suggested. Please look at the library portion of this file as I didnt rename that... I'm not sure of the solidity idioms and if im following it in this case

crates/boundless-market/src/contracts/mod.rs
ec2: I dont think we have a data structure to decode it to. I think it makes sense to check the length instead.

crates/guest/assessor/assessor-guest/src/main.rs
ec2: Oops i left an old comment there

crates/assessor/src/lib.rs
ec2: I'm a little confused here. What you've suggested is make 2 new enums which reflect what is on chain.

``` 
pub enum Predicate {
    DigestMatch { image_id: Digest, journal_digest: Digest },
    PrefixMatch { image_id: Digest, journal_prefix: Bytes },
    ClaimDigestMatch { claim_digest: Digest },
}

pub enum FulfillmentData {
    /// No fulfillment data delivered
    None,
    /// Proofs fulfilled with both image id and journal and the claim is calculated from them.
    ImageIdAndJournal(Digest, Bytes),
}
```

This `enum Predicate` would then have a method `fn eval(&self, f: &FulfillmentData) -> bool`.

If the FulfillmentData should reflect whats on chain, and when its a ClaimDigestMatch AND `FulfillmentData::None`, what you are suggesting is taking the claim digest from the requirements, i.e. from the predicate data, basically comparing something with itself?

willpote:
@nategraf @ec2 
> _

> So the assessor controls what is allowable, with the additional restriction that if a callback is requested the prover must deliver the image ID and journal so that the callback can proceed.

Just to rephrase to check I understand what you are saying- it sounds like your point is that the assessor can dictate what fulfillment data types are valid for each particular ProofRequests? So it can, for example, just fail if the prover tries to do `FulfillmentDataType == NONE` for a ProofRequest that asks for a callback?

However currently we do not do that- the assessor would allow a FulfillmentDataType == NONE for a ProofRequest that wants a callback, and we instead do this check on-chain and revert there. Should we move this check to the asssesor?

ec2:
> @nategraf @ec2
> 
> > _
> 
> > So the assessor controls what is allowable, with the additional restriction that if a callback is requested the prover must deliver the image ID and journal so that the callback can proceed.
> 
> Just to rephrase to check I understand what you are saying- it sounds like your point is that the assessor can dictate what fulfillment data types are valid for each particular ProofRequests? So it can, for example, just fail if the prover tries to do `FulfillmentDataType == NONE` for a ProofRequest that asks for a callback?
> 
> However currently we do not do that- the assessor would allow a FulfillmentDataType == NONE for a ProofRequest that wants a callback, and we instead do this check on-chain and revert there. Should we move this check to the asssesor?

You're right. I think we don't need to move the check to the assessor though. In the future, we may want to introduce another callback type that can be called where the FulfillmentDataType == NONE. By enforcing the check in the contract instead of the assessor, we won't need to redeploy (although I guess the less stuff we do on chain, the cheaper things get?).

Anyways, checking on chain vs in assessor both seem reasonable to me. What do you think? @nategraf ?

willpote:
> > @nategraf @ec2
> > > _
> > 
> > 
> > > So the assessor controls what is allowable, with the additional restriction that if a callback is requested the prover must deliver the image ID and journal so that the callback can proceed.
> > 
> > 
> > Just to rephrase to check I understand what you are saying- it sounds like your point is that the assessor can dictate what fulfillment data types are valid for each particular ProofRequests? So it can, for example, just fail if the prover tries to do `FulfillmentDataType == NONE` for a ProofRequest that asks for a callback?
> > However currently we do not do that- the assessor would allow a FulfillmentDataType == NONE for a ProofRequest that wants a callback, and we instead do this check on-chain and revert there. Should we move this check to the asssesor?
> 
> You're right. I think we don't need to move the check to the assessor though. In the future, we may want to introduce another callback type that can be called where the FulfillmentDataType == NONE. By enforcing the check in the contract instead of the assessor, we won't need to redeploy (although I guess the less stuff we do on chain, the cheaper things get?).
> 
> Anyways, checking on chain vs in assessor both seem reasonable to me. What do you think? @nategraf ?

We would always need to redeploy the contract regardless, since the assessor image ID and IPFS url is set in the contract itself. But yes, part of the reason the assessor exists is to move expensive validation off-chain. This check isn't particularly expensive, so less concerned about that, but moreso where this type of validation should live long term

ec2:
> > > @nategraf @ec2
> > > > _
> > > 
> > > 
> > > > So the assessor controls what is allowable, with the additional restriction that if a callback is requested the prover must deliver the image ID and journal so that the callback can proceed.
> > > 
> > > 
> > > Just to rephrase to check I understand what you are saying- it sounds like your point is that the assessor can dictate what fulfillment data types are valid for each particular ProofRequests? So it can, for example, just fail if the prover tries to do `FulfillmentDataType == NONE` for a ProofRequest that asks for a callback?
> > > However currently we do not do that- the assessor would allow a FulfillmentDataType == NONE for a ProofRequest that wants a callback, and we instead do this check on-chain and revert there. Should we move this check to the asssesor?
> > 
> > 
> > You're right. I think we don't need to move the check to the assessor though. In the future, we may want to introduce another callback type that can be called where the FulfillmentDataType == NONE. By enforcing the check in the contract instead of the assessor, we won't need to redeploy (although I guess the less stuff we do on chain, the cheaper things get?).
> > Anyways, checking on chain vs in assessor both seem reasonable to me. What do you think? @nategraf ?
> 
> We would always need to redeploy the contract regardless, since the assessor image ID and IPFS url is set in the contract itself. But yes, part of the reason the assessor exists is to move expensive validation off-chain. This check isn't particularly expensive, so less concerned about that, but moreso where this type of validation should live long term

Hey! What youre saying makes sense, so I'm now doing the check in the assessor as well.

contracts/src/types/Predicate.sol
```diff
@@ -83,3 +106,17 @@ library PredicateLibrary {
         return keccak256(abi.encode(PREDICATE_TYPEHASH, predicate.predicateType, keccak256(predicate.data)));
     }
 }
+
+/// Taken from Openzepplin util Bytes.sol
+function _sliceToEnd(bytes memory buffer, uint256 start) pure returns (bytes memory) {
+    // sanitize
+    uint256 end = buffer.length;
```
nategraf: This drops a check that end <= start from that library. Why not just use the library?
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.4.0/contracts/utils/Bytes.sol#L86-L99

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -430,9 +450,9 @@ impl ProofRequest {
         }
         Url::parse(&self.imageUrl).map(|_| ())?;
 
-        if self.requirements.imageId == B256::default() {
-            return Err(RequestError::ImageIdIsZero);
-        }
+        // The conversion from RequestPredicate to Predicate will validate
+        <Predicate>::try_from(self.requirements.predicate.clone())?;
```
nategraf: I don't _think_ there is any reason for this syntax
```suggestion
        Predicate::try_from(self.requirements.predicate.clone())?;
```

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -553,20 +567,248 @@ impl Requirements {
...
+    pub fn image_id(&self) -> Option<Digest> {
+        match self {
+            Predicate::DigestMatch(image_id, ..) => Some(*image_id),
+            Predicate::PrefixMatch(image_id, ..) => Some(*image_id),
+            Predicate::ClaimDigestMatch(..) => None,
+        }
+    }
+
+    /// Evaluates the predicate against the fulfillment data and returns the claim digest if eval passes.
+    pub fn eval(&self, fulfillment_data: &FulfillmentData) -> bool {
```
nategraf: The comments doesn't match the actual return. It might simplify the implementation of the evaluate requirements function in `Fulfillment` if this actually did return the claim digest.

crates/boundless-market/src/request_builder/finalizer.rs
```diff
@@ -128,11 +129,29 @@ impl Adapt<Finalizer> for RequestParams {
...
+                let eval_data =
+                    FulfillmentData::from_image_id_and_journal(image_id, journal.bytes.clone());
+                predicate.eval(&eval_data)
             }
+            (Some(journal), None) => {
+                tracing::debug!(
+                    "Image id not provided, only ensuring the journal matches predicate"
+                );
+                predicate.check_journal(&journal.bytes)
+            }
```
nategraf: I don't think there should ever be a case where the journal is provided, but not he image ID. I think we can just drop this arm of the match and not run the check, it's not critical as this is just a sanity check to catch errors sooner.

crates/boundless-market/src/request_builder/requirements_layer.rs
```diff
@@ -66,12 +66,15 @@ pub struct RequirementParams {
 
 impl From<Requirements> for RequirementParams {
     fn from(value: Requirements) -> Self {
+        // TODO(ec2): unwrap
+        let predicate = Predicate::try_from(value.predicate).unwrap();
```
nategraf: It seems this implementation needs to be a TryFrom now

crates/guest/assessor/assessor-guest/src/main.rs
```diff
@@ -45,9 +42,9 @@ fn main() {
     // For each fill we
     // - verify the request's signature
     // - evaluate the request's requirements
-    // - verify the integrity of its claim
     // - record the callback if it exists
     // - record the selector if it is present
+    // NOTE: We do not verify integrity of the claim. That is done on chain.
```
nategraf:
```suggestion
    // NOTE: We do not verify soundness of the claim. That is done on chain.
```

examples/counter/apps/src/main.rs
```diff
@@ -99,33 +98,44 @@ async fn run(args: Args) -> Result<()> {
...
+        client
+            .new_request()
+            .with_program(ECHO_ELF)
+            .with_stdin(echo_message.as_bytes())
+            .with_requirements(
+                RequirementParams::builder()
+                    .predicate(Predicate::claim_digest_match(r0_claim_digest))
+                    .build()
+                    .unwrap(),
+            )
```
nategraf: Why change this example to use claim digest match? We generally want the examples to show the simplest way to get the job done, and this is kind of considered an advanced feature right now.

crates/boundless-market/src/contracts/mod.rs
nategraf: I saw this somewhere else too. It would be good to grep for it

examples/counter/apps/src/main.rs
ec2: True! I kinda used the example just to see how things would work before. Ill revert the changes.

crates/boundless-cli/src/bin/boundless.rs
```diff
@@ -687,29 +687,62 @@ async fn handle_request_command(cmd: &RequestCommands, client: StandardClient) -
...
+                (Predicate::ClaimDigestMatch(claim_digest), FulfillmentData::None) => {
+                    verifier
+                        .verifyIntegrity(Receipt {
+                            seal,
+                            claimDigest: <[u8; 32]>::from(*claim_digest).into(),
+                        })
+                        .call()
+                        .await
+                        .map_err(|_| anyhow::anyhow!("Integrity verification failed"))?;
+                }
```
nategraf: The command interface has kind of painted us into a corner here. This is not right, because we did not actually check the given image ID, and there is no way we can. Given the interface, all we can do here is error since we can't actually fulfill the contract of ensuring it matches the image ID.

ec2: Make sense!

ec2: I error here now.

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -553,20 +567,249 @@ impl Requirements {
...
+}
+
+impl TryFrom<RequestPredicate> for Predicate {
+    type Error = PredicateError;
+
+    fn try_from(value: RequestPredicate) -> Result<Self, Self::Error> {
+        match value.predicateType {
+            PredicateType::DigestMatch => {
+                // Don't need to check the length because that is checked when converting to Digest
+                let (image_id, journal_digest) = value.data.as_ref().split_at(32);
```
nategraf: This will panic if the predicate data is less than 32 bytes. This could actually be meaningful in that an attacker can submit malformed data to the network that will cause the Broker to have a panic in the task. A the very least, this would be a performance issue. Let's use `.split_at_checked` here and below.
```suggestion
                let (image_id, journal_digest) = value.data.as_ref().split_at_checked(32).ok_or(PredicateError::DataMalformed)?;
```

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -553,20 +567,249 @@ impl Requirements {
...
+    pub fn eval(&self, fulfillment_data: &FulfillmentData) -> Option<Digest> {
+        let claim_digest_data = match fulfillment_data {
+            FulfillmentData::None => None,
+            FulfillmentData::ImageIdAndJournal(image_id, journal) => {
+                Some(ReceiptClaim::ok(*image_id, journal.to_vec()).digest())
+            }
+        };
+        let claim_digest_predicate = match self {
+            Predicate::DigestMatch(image_id, journal) => {
+                Some(ReceiptClaim::ok(*image_id, MaybePruned::Pruned(*journal)).digest())
```
nategraf, line 798: I missed a check when I wrote this, to require the prover to deliver the journal.
```suggestion
                // With the DigestMatch predicate, we require the delivery of the journal.
                // If is checked to match the predicate by the claim digest comparison below.
                let FulfillmentData::ImageIdAndJournal(_, _) = &fulfillment_data else {
                    return None;
                };
                Some(ReceiptClaim::ok(*image_id, MaybePruned::Pruned(*journal)).digest())
```

crates/guest/assessor/assessor-guest/src/main.rs
```diff
@@ -63,24 +60,34 @@ fn main() {
...
+
+        if let FulfillmentData::ImageIdAndJournal(_, _) = &fill.fulfillment_data {
+            if callback.addr != Address::ZERO {
+                callbacks.push(AssessorCallback {
+                    index: index.try_into().expect("callback index overflow"),
+                    addr: callback.addr,
+                    gasLimit: callback.gasLimit,
+                });
+            }
         }
```
nategraf: As discussed on the call, it would be good to have the assessor fail if a callback is requested but cannot be completed. As is, the callback would actually be skipped which means the client would request a callback and not get it.
```suggestion
        if callback.addr != Address::ZERO {
            match &fill.fulfillment_data {
                 FulfillmentData::ImageIdAndJournal(_, _) => {}
                 _ => panic!("callback requested but no fulfillment data provided");
            }
            callbacks.push(AssessorCallback {
                index: index.try_into().expect("callback index overflow"),
                addr: callback.addr,
                gasLimit: callback.gasLimit,
            });
        }
```

nategraf (commented):
Found a few non-trivial issues. I believe these are the last issues.

contracts/src/types/FulfillmentData.sol
willpote: Bump!

crates/boundless-market/tests/e2e.rs
```diff
@@ -18,11 +18,12 @@ use alloy::{
     providers::Provider,
     sol_types::eip712_domain,
 };
+use alloy_primitives::Bytes;
```
willpote, line 21: Would really like to see an e2e test for this new functionality

PR #1011, opened 2025-08-12
BM-1473: Zeroecco/bento

zeroecco:
🎉 move bento over to boundless
🎉 start redis before starting bento stuff
🎉 update nvidia to 12.9.1 
🐛 drop sm_code=90 as it messes with mixed envs

### DIFF
--- a/crates/slasher/src/lib.rs
+++ b/crates/slasher/src/lib.rs
@@ -93,6 +93,7 @@ pub struct SlashServiceConfig {
     pub balance_error_threshold: Option<U256>,
     pub skip_addresses: Vec<Address>,
     pub tx_timeout: Duration,
+    pub max_block_range: u64,
 }
 
 impl SlashService<ProviderWallet> {
@@ -152,12 +153,25 @@ where
                         continue;
                     }
 
-                    tracing::info!("Processing blocks from {} to {}", from_block, to_block);
+                    // Cap the processing range to max_block_range
+                    let chunk_to =
+                        std::cmp::min(from_block + self.config.max_block_range - 1, to_block);
 
-                    match self.process_blocks(from_block, to_block).await {
+                    if chunk_to < to_block {
+                        tracing::info!(
+                            "Processing blocks from {} to {} (chunked, current block: {})",
+                            from_block,
+                            chunk_to,
+                            to_block,
+                        );
+                    } else {
+                        tracing::info!("Processing blocks from {} to {}", from_block, chunk_to);
+                    }
+
+                    match self.process_blocks(from_block, chunk_to).await {
                         Ok(_) => {
                             attempt = 0;
-                            from_block = to_block + 1;
+                            from_block = chunk_to + 1;
                         }
                         Err(e) => match e {
                             // Irrecoverable errors
--- a/crates/slasher/src/main.rs
+++ b/crates/slasher/src/main.rs
@@ -60,6 +60,9 @@ struct MainArgs {
     /// Transaction timeout in seconds.
     #[clap(long, default_value = "120")]
     tx_timeout: u64,
+    /// Maximum block range to query in a single request.
+    #[clap(long, default_value = "500")]
+    max_block_range: u64,
 }
 
 fn parse_address(s: &str) -> Result<Address, String> {
@@ -98,6 +101,7 @@ async fn main() -> Result<()> {
             balance_error_threshold: args.error_balance_below,
             skip_addresses: args.skip_addresses,
             tx_timeout: Duration::from_secs(args.tx_timeout),
+            max_block_range: args.max_block_range,
         },
     )
     .await?;
--- a/infra/slasher/Pulumi.prod-11155111.yaml
+++ b/infra/slasher/Pulumi.prod-11155111.yaml
@@ -10,6 +10,7 @@ config:
   slasher:DOCKER_TAG: latest
   slasher:INTERVAL: "60"
   slasher:LOG_LEVEL: "boundless_slasher=trace,boundless_market=debug"
+  slasher:MAX_BLOCK_RANGE: "500"
   slasher:RETRIES: "5"
   slasher:SKIP_ADDRESSES: ""
   slasher:SLACK_ALERTS_TOPIC_ARN: arn:aws:sns:us-west-2:968153779208:boundless-alerts-topic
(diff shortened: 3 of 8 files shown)


infra/kailua-order-generator/Pulumi.dev.yaml
```diff
@@ -0,0 +1,98 @@
+encryptionsalt: v1:c+qkUDk4J5U=:v1:v2sUQtg3aH/C4Ghk:vqM6TFhW45Y9+FmlBOVZf51DRK9eFQ==
```
willpote: Mistake?

Btw https://github.com/boundless-xyz/boundless/blob/main/infra/README#L30 :)

infra/kailua-order-generator/Pulumi.dev.yaml
```diff
@@ -0,0 +1,98 @@
...
+  kailua-order-generator-offchain:RAMP_UP: "0.2"
+  kailua-order-generator-offchain:LOCK_TIMEOUT: "60"
+  kailua-order-generator-offchain:TIMEOUT: "1800"
+  kailua-order-generator-offchain:SECONDS_PER_MCYCLE: "20"
+  kailua-order-generator-offchain:INTERVAL: "30"
+  # Kailua-specific configuration
+  kailua-order-generator-base:KAILUA_DEVNET_URL:
+    secure: v1:vRqhmJr08pYRh5xs:UyCB20dPgwx/TV6LM36BWNi4Ws17Oy5FSjN6D5YQmU3hVk/XOkmZ3VTnFOns2mLcncnxla/n57EvDsQuNQ8leOwpZLqQYFVADwaKsCMPSSrnpBibvZfQAc5Zx0qNCN7Vadvb
+  kailua-order-generator-base:KAILUA_L1_URL: https://ancient-ultra-pool.quiknode.pro/0f7b97e0e4258b667c3156233dfde6e935ddf343/
+  kailua-order-generator-base:KAILUA_L2_URL: https://evocative-blissful-moon.optimism.quiknode.pro/b52a373470f753a59b129684feecb2973ec8af24/
```
willpote: Whoops, should rotate

infra/kailua-order-generator/Pulumi.dev.yaml
zeroecco: yeah, that was due to a pr on top of a pr while moving from boundless to another repo. should have done it clean :(

infra/kailua-order-generator/Pulumi.dev.yaml
zeroecco: understood

zeroecco: yup

compose.yml
```diff
@@ -75,6 +87,13 @@ services:
       - 6379:6379
     volumes:
       - redis-data:/data
+    command: redis-server --maxmemory 2gb --maxmemory-policy allkeys-lru --save 900 1 --appendonly yes
```
austinabell: Where does the 2gb number come from? Wondering the reason for this default? This seems like it would cause issues for provers locking a lot of large orders

compose.yml
```diff
@@ -11,12 +11,17 @@ x-base-environment: &base-environment
   RUST_BACKTRACE: 1
 
 x-agent-common: &agent-common
-  image: risczero/risc0-bento-agent:2.3.1@sha256:7873f18005efff03fc5399f1bdcb6760cda7ffbd4fdd4d9c39aedee8972e0a0d
+  build:
+    context: .
+    dockerfile: dockerfiles/agent.dockerfile
```
austinabell, line 16: Noting that this change will add a lot of friction for provers building. It is already incredibly time consuming to build just the broker and we have been getting a lot of complaints.

Seems like this could be done in two ways, either also including the image in the declaration, so if you don't pass `--build` it uses prebuilt, and if you do it would rebuild. Alternative to have a `compose.local.yml` to look something like:

```
services:
  exec_agent0:
    image: null
    build:
      context: .
      dockerfile: dockerfiles/agent.dockerfile

  exec_agent1:
    image: null
    build:
      context: .
      dockerfile: dockerfiles/agent.dockerfile

  aux_agent:
    image: null
    build:
      context: .
      dockerfile: dockerfiles/agent.dockerfile

  gpu_prove_agent0:
    image: null
    build:
      context: .
      dockerfile: dockerfiles/agent.dockerfile

  snark_agent:
    image: null
    build:
      context: .
      dockerfile: dockerfiles/agent.dockerfile 
```

(haven't extensively tested both, there could be a reason why one might be worse)

compose.yml
```diff
@@ -26,7 +31,6 @@ x-exec-agent-common: &exec-agent-common
   cpus: 3
   environment:
     <<: *base-environment
-    LD_LIBRARY_PATH: ${LD_LIBRARY_PATH:-/usr/local/cuda-12.2/compat/}
```
austinabell, line 29: Why remove this instead of updating to 12.9?

austinabell (changes requested):
The nvidia compat is the only thing I think is blocking, other comments are somewhat opinionated

compose.yml
zeroecco: not needed in my testing with 12.9 on ampere and blackwell cards

compose.yml
zeroecco: It is my suggestion that we have really two options here:
1. pre build the binaries and pull them from somewhere directly into the container
2. the baking of images.

I favor the first option I am presenting. if this compiling is a larger issue we should pre compile the binaries and post them in releases IMHO.

zeroecco: if we agree on my suggestion I am fine to begin setting up the binaries build process

compose.yml
austinabell: Seems like this is the case, although the specific use case is being able to run these on machines without a CUDA GPU, and not certain it works on all machines -- harder to test. Perhaps fine for this to be removed and see if anyone complains

aridder: I had to remove this line to make it work with 5090

compose.yml
zeroecco: `b7d1f2e6`

compose.yml
zeroecco: punting this to a follow up PR

PR #1023, opened 2025-08-20
fix: stake token max denom (and bench)

austinabell:
Main reason for fix is the max stake tokens config, but when doing a scan for any other missed I noticed the bench program (unused) had one so swapped that just to avoid any issues in the future


austinabell:
> I think we should drop entirely this bench crate

Can do in a future PR, could still be useful to test optimizations when hitting higher order counts

PR #1026, opened 2025-08-21
BM-1492: make exec rate configurable for order generators, adjust values

austinabell:
Flat period hasn't been long enough for provers to process, giving a little bit of breathing room and making configurable to adjust


zeroecco (approved):
small nit, newlines

else looks ok to me 5mhz is safe for all

austinabell:
> else looks ok to me 5mhz is safe for all

that was just for if the value was set to 0 to avoid a panic, but this isn't reasonably hit and confusing, so I just removed that handling [`3e47465` (#1026)](https://github.com/boundless-xyz/boundless/pull/1026/commits/3e47465cf8dc56aab51acb92043451033628f9e2)

PR #1034, opened 2025-08-27
BM-1507: Update risc0-zkvm and risc0-ethereum to 3.0

nategraf:


nategraf:
Note that the bump to Rust 1.89 is required by downstream crate `wide` bumping their MSRV in `0.7.34` relative to `0.7.33`. This was released today. Another option would be to pin to `wide 0.7.33`.

Also note that the host requires `1.89` while the latest RISC Zero toolchain we provide for the guest is `1.88`. So they are on different versions of rust now. This is not actually an issue, but is a bit non-intuitive.

nategraf:
A couple more issues:
* Building `cuda` support now required having `protoc` (the protobuf compiler) installed. Additionally, `risc0-groth16` needs to be installed via `rzup` or proving will fail at runtime.
* After updating the `risc0` crates to 3.0 and running `cargo test`, the system got into a bad state such that execution of the guests was failing with a combination of host-side I/O errors and guest-side illegal instruction errors (possibly a panic in the guest kernel). Running the following incantation to run `cargo update` on all workspace (including guests) fixed this issue locally `ag -G Cargo.toml '\[workspace\]' -l | xargs -n1 cargo update --manifest-path`

nategraf:
I figured out the reason for this. When I updated the `Cargo.toml` files are ran `cargo test` the `Cargo.lock` got into a state where the workspace dependency on `risc0-zkos-v1compat` was on `2.0.2` and the guest dependency on `risc0-zkvm-platform` was on `2.1.0`. These versions are not compatible, and their moving out of sync was partially facilitated by the fact that they two workspaces are independent.

nategraf:
Looks like that solved the issue in the assessor, but not for the set builder in the examples.

nategraf:
In the examples, we use the `guest-set-builder` crate and build the set-builder when building the test. This uses the lock file from `risc0-ethereum` for the guest, but is built with `risc0-build` and associated deps based on the workspace (in this repo). As a result, there was a mismatch which should be solved by https://github.com/risc0/risc0-ethereum/pull/678

ec2:
Might be a good idea to update the toolchain files in the examples as well?

nategraf:
> Might be a good idea to update the toolchain files in the examples as well?

Ah, indeed. I will do that.

UPDATE: It turns out these are all symlinks to the root `rust-toolchain.toml`

nategraf:
Holding this as merging it will effect the staging deployment. Additionally, the `Cargo.toml` files should be bumped to reference 3.0.3 since that's now the latest version.

austinabell (approved):

PR #1044, opened 2025-08-28
BM-1513: BM-1516: Zeroecco/pvw

zeroecco:


bento/crates/api/src/lib.rs
```diff
@@ -54,6 +50,18 @@ impl std::fmt::Display for ErrMsg {
     }
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+pub struct WorkReceiptInfo {
+    pub key: String,
```
nategraf, line 55: What is the key? Is that what the user should send back when they want to fetch the receipt?

bento/crates/api/src/lib.rs
```diff
@@ -54,6 +50,18 @@ impl std::fmt::Display for ErrMsg {
     }
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+pub struct WorkReceiptInfo {
```
nategraf, line 54: Could you include the PoVW log ID and PoVW job number in the API response to the user? The client needs some way to tell which receipts they've already added to their log, and the best way for them to do that is the job number. I'd also like to have the log ID associated with the stored receipt so that if the prover changes it (by restarting bento with a new one) it will be possible to differentiate what log ID a given receipt in storage was produced with.

As I understand it, the S3 storage may be the only record we have of this receipt. Two ideas I have on how to work with this are:
1. Store an index file along with the receipts with the requires information. In order to avoid pulling hundreds of files, we could use a single index file for all receipts.
2. Store the log ID and job number in the object key, and parse it from there. I don't want the client to parse the keys, so that they just treat the key an opaque identifier, but having the prover parse them seems less bad.

Maybe you have other, better ideas

bento/crates/api/src/lib.rs
```diff
@@ -665,6 +647,57 @@ async fn groth16_download(
...
+        .context("Failed to list work receipts")?;
+
+    let mut receipts = Vec::new();
+    for key in objects {
+        // Extract filename from key
+        if let Some(filename) = key.rsplit('/').next() {
+            receipts.push(WorkReceiptInfo {
+                key: filename.to_string(),
+                size: 0, // We don't have size info from list_objects
+                last_modified: "Unknown".to_string(), // We don't have timestamp info from list_objects
```
nategraf: Should we just not include these in the response then?

bento/crates/bento-client/src/bento_cli.rs
```diff
@@ -73,29 +67,15 @@ async fn main() -> Result<()> {
...
-        bytemuck::cast_slice(&to_vec(&input).expect("Failed to r0 to_vec")).to_vec();
-    let (session_uuid, _receipt_id) =
-        stark_workflow(&client, image, input, vec![receipt_id], args.exec_only)
-            .await
-            .context("STARK proof workflow failure")?;
-
-    // snarkify
-    if args.snarkify {
-        stark_2_snark(session_uuid, client).await?;
-    }
```
nategraf, line 97: Should we keep this, or do we just not want to support SNARK proving via the bento CLI? I don't think its a hard requirement.

bento/crates/workflow/src/tasks/executor.rs
```diff
@@ -565,6 +562,14 @@ pub async fn executor(agent: &Agent, job_id: &Uuid, request: &ExecutorReq) -> Re
     let exec_task: JoinHandle<anyhow::Result<SessionData>> =
         tokio::task::spawn_blocking(move || {
             let mut env = ExecutorEnv::builder();
+            if std::env::var("POVW_ENABLED").unwrap_or_default() == "true" {
+                let log_id = std::env::var("POVW_LOG_ID").context("POVW_LOG_ID is not set")?;
```
nategraf: Should we drop `POVW_ENABLED` and instead use whether or not `POVW_LOG_ID` is set as the signal on whether to use PoVW?

bento/crates/workflow/src/tasks/executor.rs
```diff
@@ -565,6 +562,14 @@ pub async fn executor(agent: &Agent, job_id: &Uuid, request: &ExecutorReq) -> Re
...
+                let povw_log_id = PovwLogId::from_str(&log_id)
+                    .map_err(|e| anyhow::anyhow!("Failed to parse POVW_LOG_ID: {}", e))?;
```
nategraf: Could we additionally check this on startup, so that the operator learns about the issue sooner.

bento/crates/workflow/src/tasks/resolve_povw.rs
```diff
@@ -0,0 +1,175 @@
...
+    // Deserialize as POVW receipt
+    let povw_receipt: SuccinctReceipt<WorkClaim<ReceiptClaim>> =
+        deserialize_obj::<SuccinctReceipt<WorkClaim<ReceiptClaim>>>(&receipt)
+            .context("Failed to deserialize as POVW receipt")?;
+
+    // Unwrap the POVW receipt to get the ReceiptClaim for processing
+    let mut conditional_receipt: SuccinctReceipt<ReceiptClaim> =
+        if let Some(prover) = agent.prover.as_ref() {
+            prover.unwrap_povw(&povw_receipt).context(
+            "POVW unwrap method not available - POVW functionality requires RISC Zero POVW support",
```
nategraf: Is proving fails, this seems like it might be a confusing error messages, since their are other reasons this could fail. Also, because the prover is compiled into this binary I believe if PoVW is not available, the build will fail.

bento/LICENSE-BSL
nategraf, line 1: We have a license in the root of the repo that should apply
https://github.com/boundless-xyz/boundless/blob/main/LICENSE-BSL

bento/crates/workflow/src/tasks/resolve_povw.rs
```diff
@@ -0,0 +1,175 @@
...
+    .await
+    .context("Failed to set root receipt key with expiry")?;
+
+    // Save the resolved receipt to work receipts bucket for later consumption
+    let work_receipt_key = format!("{WORK_RECEIPTS_BUCKET_DIR}/{job_id}.bincode");
+    tracing::debug!("Saving resolved POVW receipt to work receipts bucket: {work_receipt_key}");
+
+    agent
+        .s3_client
+        .write_to_s3(&work_receipt_key, &receipt)
```
nategraf: Would you mind actually taking the `povw_receipt`, wrapping it as `GenericReceipt::Succinct` [1] and _then_ saving it to the bucket? That is the struct `r0vm` produces and this would be nice for integration.

[1] https://docs.rs/risc0-zkvm/latest/risc0_zkvm/enum.GenericReceipt.html#variant.Succinct

nategraf (commented):
Looks good to me! I do have a couple of comments, thinking about integration with the CLI I am writing.

bento/crates/workflow/src/tasks/resolve_povw.rs
zeroecco: 21ede8d4

bento/crates/workflow/src/tasks/executor.rs
zeroecco: this is a good point that I couldn't answer for myself. I can change it but to me it feels like "6 over there and half a dozen on the other side". I defer the decision to the reviewer

bento/LICENSE-BSL
zeroecco: lychee got mad at me unless I put the file here :(

bento/crates/workflow/src/tasks/executor.rs
zeroecco: b02e61a9

bento/crates/workflow/src/tasks/resolve_povw.rs
zeroecco: b02e61a9

bento/crates/api/src/lib.rs
zeroecco: [4ef3317](https://github.com/boundless-xyz/boundless/pull/1044/commits/4ef331774d118fc2a7709a2e4ff1242d16b45694)

bento/crates/bento-client/src/bento_cli.rs
zeroecco: nah, needs to be re-impl'd imo

bento/crates/workflow/src/tasks/executor.rs
zeroecco: fixed.

bento/crates/api/src/lib.rs
zeroecco: receipt identifier that uniquely identifies a specific work receipt. if a user wants a specific reciept this is the key they need.

compose.yml
```diff
@@ -168,9 +169,6 @@ services:
   exec_agent0:
     <<: *exec-agent-common
 
-  exec_agent1:
-    <<: *exec-agent-common
```
austinabell: This was an intentional choice given 1 executor by default can be very prohibitive and conflicting between preflight and proving. Was there a specific reason for removing? Noting that above depends on this exec agent also, so will cause issues when starting up with broker as is

zeroecco: changed.

PR #1050, opened 2025-08-29
BM-1519: Rename stake -> collateral, biddingStart -> rampUpStart

willpote:


lib/risc0-ethereum
nategraf, line 1: Did you mean to update the `risc0-ethereum` submodule?

nategraf (approved):

PR #1052, opened 2025-08-30
Add deprecated assessor

capossele:
To make contracts upgrades less disruptive when updating the assessor image id, we introduce the `DEPRECATED_ASSESSOR_ID` and the `DEPRECATED_ASSESSOR_EXPIRES_AT` fields.  These values are used to keep active the previous version of the assessor until its expiration. In this way, contract upgrades can be performed without disrupting ongoing fulfillments.


contracts/test/BoundlessMarket.t.sol
```diff
@@ -1122,6 +1172,43 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
...
+        vm.prank(testProverAddress);
+        boundlessMarket.lockRequest(request, clientSignature);
+
+        (Fulfillment memory fill, AssessorReceipt memory assessorReceipt) =
+            createDeprecatedFillAndSubmitRoot(request, APP_JOURNAL, testProverAddress);
+
+        Fulfillment[] memory fills = new Fulfillment[](1);
+        fills[0] = fill;
+
+        if (block.timestamp <= boundlessMarket.DEPRECATED_ASSESSOR_EXPIRES_AT()) {
```
willpote, line 1193: The way this test is written it feels easy to accidently never test one of the branches. E.g. if the test were to only warp forward by 5 seconds instead of 5 minutes, we'd never test the revert branch, and we'd never realise.

Would prefer for the test to explicitly test the various cases without an if statement like this

contracts/test/BoundlessMarket.t.sol
```diff
@@ -1346,6 +1433,12 @@ contract BoundlessMarketBasicTest is BoundlessMarketTest {
         );
     }
 
+    function testFulfillDeprecatedAssessor() public {
+        _testFulfillDeprecatedAssessor(1);
+        vm.warp(block.timestamp + 5 minutes);
```
willpote: prefer something like: `+ DEPRECATED_ASSESSOR_DURATION + 1 minute`

austinabell (approved):

PR #1053, opened 2025-08-31
feat: propagate error for client

### LINKED ISSUE #1042: Better error message propagation
Currently I am locally running boundless, and while uploading the program to S3 I get the following error message:

```Failed to upload program```

and that is it, I do not know if it is about an access issue, size issue , bucket issue I have no idea.

I had to manually debug to find out what was the actual issue about s3 upload, turns out I did not have access to create bucket (tries to create bucket in ensure_bucket_init fn), which could never be derived from the error message above.

Is there any chance you can just propagate errors directly as a string next to generic error messages? Something like this would be really nice:

```Failed to upload program, Reason: Bucket Create Access Denied```

varun-doshi:
Fixes #1042 

Propagates errors up namely for two functions: `upload_program` and `upload_input`

### DIFF
--- a/crates/boundless-market/src/client.rs
+++ b/crates/boundless-market/src/client.rs
@@ -627,7 +627,7 @@ where
             .context("Storage provider not set")?
             .upload_program(program)
             .await
-            .map_err(|_| anyhow!("Failed to upload program"))?)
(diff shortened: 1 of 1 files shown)


austinabell:
already fixed with https://github.com/boundless-xyz/boundless/pull/1016

PR #1054, opened 2025-08-31
BM-1523: update bento rest API health check

austinabell:
Will scan through PRs to see if there is a health check planned to be added soon, but this whiffs and causes issues on startup


austinabell:
Closing because added in #1044  (https://github.com/boundless-xyz/boundless/pull/1044/commits/378221a98c44ccaa19c34fdac5465ec713bbe51e)

PR #1056, opened 2025-09-01
Make PoVW Upgradable, add deployment scripts [do not merge]

willpote:
Also includes Ethereum Sepolia Staging and Prod deployments


nategraf:
Note that `cargo test -p boundless-povw-guests` is currently failing because test-utils needs to be updated with the new deployment procedure.

PR #1060, opened 2025-09-02
BM-1525: lower error log on session limit

austinabell:
This error:
```
exec_agent1-1       | 2025-09-02T00:06:13.118698Z ERROR workflow::tasks::executor: Failed to run executor
exec_agent1-1       | 2025-09-02T00:06:13.118848Z ERROR workflow: Failure during task processing: Executor failed
exec_agent1-1       |
exec_agent1-1       | Caused by:
exec_agent1-1       |     0: execution failed failed
exec_agent1-1       |     1: Session limit exceeded: 250609664 >= 250609664
```
Was very confusing and worrying for provers and not clear that it is intended. This just avoids error logging in those cases and removes the double `failed` word in the error context

### DIFF
--- a/bento/crates/workflow/src/tasks/executor.rs
+++ b/bento/crates/workflow/src/tasks/executor.rs
@@ -602,13 +602,13 @@ pub async fn executor(agent: &Agent, job_id: &Uuid, request: &ExecutorReq) -> Re
                     total_cycles: session.total_cycles,
                     journal: session.journal,
                 }),
-                Err(err) => {
+                Err(mut err) => {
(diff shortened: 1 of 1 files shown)


austinabell:
sorry @zeroecco I realized this would still be verbose from the broker side without https://github.com/boundless-xyz/boundless/pull/1060/commits/8a59119467002af752f9e1a5e196f4cebc935039 -- which seemed like the simplest way to update. Sorry for pushing a change after your review

zeroecco:
still approved

PR #1083, opened 2025-09-05
Backport 1082

capossele:


willpote (approved):

austinabell:
This was intentional to avoid the polling subscriptions going dead and not recreating issue in alloy. Suppose fine to just have the lockfile be updated, but wasn't sure if other impls use and was being careful not to reintroduce that issue.

austinabell:
Closing this assuming we don't want to support that boundless version anyway

PR #1093, opened 2025-09-09
update set verifier address in .env.base

sashaaldrick:


willpote (approved):

austinabell:
Disabling auto merge to also apply to testnet -- hope you don't mind!

austinabell:
https://github.com/boundless-xyz/boundless/pull/1093/commits/dc252ff86d5a55ca118dff0d76bf60470fb3c288

Perhaps worth removing the env files entirely, this was previously a stop-gap until we had the default system we have now (seems like this has and will cause footguns)

PR #1102, opened 2025-09-11
BM-1571: Fix order picker error context formatting

austinabell:
Underlying error is not printed with this format for most of the errors, removed the backtrace from others as it's multiline and hard to grep for.


austinabell:
actually going to double check formatting when wrapped in Arc before pulling in

PR #1106, opened 2025-09-12
BM-1573: remove network env files and update docs

austinabell:
- Updates release branch recommendation to 0.14
- Removes network env files
- Removes unnecessary env vars from broker env template (we can remove this too, we just have the ability to run the prover with an env file that I would like to not break ATM

### DIFF
--- a/documentation/site/pages/provers/broker.mdx
+++ b/documentation/site/pages/provers/broker.mdx
@@ -84,19 +84,16 @@ Below are some `broker.toml` settings:
 
 ### Benchmarking Bento
 
-Load environment variables for the target network:
-
-```bash [Terminal]
-# For example, to benchmark an order on Ethereum Sepolia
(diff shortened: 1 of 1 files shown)


documentation/site/pages/provers/quick-start.mdx
```diff
@@ -36,7 +36,7 @@ To get started, first clone the Boundless monorepo on your proving machine, and
 ```bash [Terminal]
 git clone https://github.com/boundless-xyz/boundless
 cd boundless
-git checkout release-0.13
+git checkout release-0.14
```
willpote, line 39: May as well set this to 1.0 at this point

austinabell: Just don't want docs to be broken before the release goes out. Going to pull this in now to remove the env files and will follow up with that change if not done with the release PR

PR #1109, opened 2025-09-12
BM-1578: Move to AccessControlUpgradeable

willpote:


willpote:
Upgradability and deployment script failures are expected for this

PR #1137, opened 2025-09-15
Backport bento fixes

austinabell:
This is quite noisy and leads to missing context, so would prefer provers having this now


austinabell:
pushed a 1.0.1 tag and updated the default, should this PR to come in

austinabell:
Going to rebuild with the other backported change first https://github.com/boundless-xyz/boundless/actions/runs/17838578154/job/50721930262

zeroecco (approved):

PR #1138, opened 2025-09-15
BM-1603: Add povw/staking/delegation API

willpote:
Indexes rewards information and serves the data via an API. Will be integrated into the explorer and the CLI. To do this I:

* Split the current indexer crate into two binaries- 
  * market-indexer (the previously existing indexer code, which indexes Boundless Market)
  * reward-indexer (new, which indexes ZKC, staking, povw, etc)
* Introduced a new crates "rewards" which contains logic for fetching all raw data from chain that relates to rewards. THe indexer uses this crate to fetch data, then stores it in the DB. This crate could also be re-used in other places (e.g. the CLI)
* Refactored the indexer infra package to deploy both `market-indexer` and `reward-indexer`, conditionally based on the Pulumi config. E.g. on Eth Mainnet, we don't have a Boundless market, so we only deploy the `reward-indexer` in `Pulumi.config.1.yaml`

> Originally this PR contained a `summary` command for the CLI, I have removed it as we are building a UI with the information, and we are about to refactor the CLI massively.


crates/boundless-cli/src/commands/zkc/mod.rs
```diff
@@ -36,6 +37,7 @@ pub use get_epoch_end_time::{get_epoch_end_time, ZkcGetEpochEndTime};
 pub use get_rewards_delegates::{get_rewards_delegates, ZkcGetRewardsDelegates};
 pub use get_staked_amount::{get_staked_amount, ZkcGetStakedAmount};
 pub use stake::ZkcStake;
+pub use summary::ZkcSummary;
```
capossele: I think we should maybe hide this one behind a `experimental` or `unstable` feature?

sasha-computer: This is a highly requested feature, and would be one of the most used commands (as it can be used multiple times an epoch across many epochs). I would recommend, if we can, to make this feature stable asap; see my comment below.

sasha-computer:
This is really helpful @willpote; I have had exact feedback asking for this CLI feature many times over the past 48 hours. I would suggest we make this feature live as soon as we can; this will really help provers understand and therefore help faciliate participation in the market and PoVW. 

I wouldn't mark it as experimental, the only other option really is a web UI where provers can connect various addresses to see this visually, which has also been requested. If we do decide to go that route, that would take some time, and regardless there are always provers who prefer CLI commands like this for scripting reasons etc.

crates/boundless-cli/src/commands/zkc/mod.rs
Nordinian: > This is a highly requested feature, and would be one of the most used commands (as it can be used multiple times an epoch across many epochs). I would recommend, if we can, to make this feature stable asap; see my comment below.

Agree, looking forward to this, urgently needed.

willpote:
> This is really helpful @willpote; I have had exact feedback asking for this CLI feature many times over the past 48 hours. I would suggest we make this feature live as soon as we can; this will really help provers understand and therefore help faciliate participation in the market and PoVW.
> 
> I wouldn't mark it as experimental, the only other option really is a web UI where provers can connect various addresses to see this visually, which has also been requested. If we do decide to go that route, that would take some time, and regardless there are always provers who prefer CLI commands like this for scripting reasons etc.

Sounds good. Will try and get it stable today.

dockerfiles/rewards-indexer.dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+SHELL ["/bin/bash", "-c"]
+
+RUN cargo build --release --bin rewards-indexer
+
+FROM init AS runtime
+
+COPY --from=builder /src/target/release/rewards-indexer /app/rewards-indexer
+
+ENTRYPOINT ["/app/rewards-indexer"]
```
zeroecco: nit: newline

infra/builder/configure_builder.sh
```diff
@@ -15,7 +15,7 @@ else
 fi
 
 INSTANCE_ID=$(aws ec2 describe-instances \
-  --filters "Name=tag:Name,Values=builder-local" \
+  --filters "Name=tag:Name,Values=builder-local-v2" \
```
zeroecco, line 18: I worry about this in the future. finding this 6 months from now when we are on a new version will hurt.

infra/builder/pulumiResources.ts
```diff
@@ -10,9 +10,9 @@ export const createPulumiState = (): {
   keyAlias: aws.kms.Alias,
 } => {
   const bucket = new aws.s3.BucketV2(
-      'boundless-builder-state-bucket',
-      {
-      bucketPrefix: 'boundless-builder-state',
+    'boundless-builder-state-bucket-v2',
```
zeroecco, line 13: IIRC this will replace all the infra and not abandon it. is that the intent?

infra/indexer/Pulumi.l-prod-1.yaml
```diff
@@ -0,0 +1,14 @@
...
+  indexer:DOCKER_DIR: ../../
+  indexer:DOCKER_TAG: latest
+  indexer:START_BLOCK: "23250070"
+  indexer:BASE_STACK: organization/bootstrap/services-prod
+  indexer:POVW_ACCOUNTING_ADDRESS: 0x319bd4050b2170a7aE3Ead3E6d5AB8a5c7cFBDF8
+  indexer:VEZKC_ADDRESS: 0xE8Ae8eE8ffa57F6a79B6Cbe06BAFc0b05F3ffbf4
+  indexer:ZKC_ADDRESS: 0x000006c2A22ff4A44ff1f5d0F2ed65F781F55555
+  indexer:RUST_LOG: "info"
+  indexer:CHAIN_ID: "1"
+
```
zeroecco, line 15: nit: newline

infra/indexer/Pulumi.l-prod-11155111.yaml
```diff
@@ -7,6 +7,9 @@ config:
   indexer:CHAIN_ID: "11155111"
   indexer:DOCKER_DIR: ../../
   indexer:DOCKER_TAG: latest
+  indexer:POVW_ACCOUNTING_ADDRESS: 0xC5E956732F4bA6B1973a859Cf382244db6e84D0b
```
zeroecco, line 10: should we abstract these vars into a helper across the pulumi? like read the deployment.toml for example

infra/indexer/components/indexer-infra.ts
```diff
@@ -0,0 +1,210 @@
...
+                'ecr:BatchCheckLayerAvailability',
+                'ecr:GetDownloadUrlForLayer',
+                'ecr:BatchGetImage',
+              ],
+              Resource: '*',
+            },
+            {
+              Effect: 'Allow',
+              Action: ['logs:CreateLogStream', 'logs:PutLogEvents'],
+              Resource: '*',
```
zeroecco: we need to find a way to neck this down.

infra/indexer/components/indexer-infra.ts
```diff
@@ -0,0 +1,210 @@
...
+          Statement: [
+            {
+              Effect: 'Allow',
+              Action: [
+                'ecr:GetAuthorizationToken',
+                'ecr:BatchCheckLayerAvailability',
+                'ecr:GetDownloadUrlForLayer',
+                'ecr:BatchGetImage',
+              ],
+              Resource: '*',
```
zeroecco: we need to find a way to neck this down.

infra/indexer/components/rewards-indexer.ts
```diff
@@ -0,0 +1,260 @@
...
+      dockerfile: {
+        location: `${dockerDir}/dockerfiles/rewards-indexer.dockerfile`,
+      },
+      builder: dockerRemoteBuilder
+        ? {
+          name: dockerRemoteBuilder,
+        }
+        : undefined,
+      buildArgs: {
+        S3_CACHE_PREFIX: 'private/boundless/rust-cache-docker-Linux-X64/sccache',
```
zeroecco: this prefix could conflict with other docker impls. worth considering rescoping, but not a blocker.

infra/builder/configure_builder.sh
willpote: I had to rename this due to some shenanigans with the migration from R0 account to Boundless account. We shouldn't need to rename it again.

infra/builder/pulumiResources.ts
willpote: Yes the intent was to start the builder from scratch. I forget the exact problem, but it was something related to the R0 setup of AWS Organizations vs the new Boundless setup.

Side note: I will probably remove this entire `builder` infra once I get my 5090 build. The only reason I was using it was to avoid doing Docker builds on Mac, and once I have that build I will just do Docker there.

infra/indexer/Pulumi.l-prod-11155111.yaml
willpote: That would be nice, out of scope for this PR though I think? Would be a big change.

infra/indexer/components/indexer-infra.ts
willpote: Scoped these down to the repo

infra/indexer/components/indexer-infra.ts
willpote: Refactored so we create the Cloudwatch policy in the Market indexer + Reward indexer stack themselves, rather than this shared infra stack, so I can get the log group names

PR #1144, opened 2025-09-18
BM-1611: Zeroecco/rm beta

zeroecco:
downing mainnet beta resources


willpote:
CI is failing as prover-cluster/index.ts is referencing deleted files

PR #1152, opened 2025-09-19
BM-1620: Switch pvw commands to use ETH_RPC_URL to disambiguate deployments

austinabell:
This change is to avoid the footgun of mixing up RPC_URL from other commands, as the same variable is used to point to the market deployment and also for the accounting contract. Name isn't perfect, because very possible the accounting contract is not on Ethereum, and the market they are interacting on is also Ethereum, which could cause other confusions. 

Naming suggestions welcome, couldn't think of one that would prevent the footguns we have been seeing more than this.


crates/boundless-cli/src/commands/povw/claim.rs
```diff
@@ -87,6 +88,10 @@ pub struct PovwClaim {
 
     #[clap(flatten, next_help_heading = "Prover")]
     prover_config: ProverConfig,
+
+    /// The RPC URL to use to submit the reward claim transaction.
+    #[clap(long, env = "ETH_RPC_URL")]
```
derrix060, line 93: nit: maybe rename it to `REWARD_RPC_URL` so in case it's decided to use a different chain for the reward (for example base due to tx fees in eth), it's a no brainer :)

austinabell: yeah agree that might be a better name. @willpote you good with this, also because similarity with us using `REWARD_ADDRESS` env vars now?

austinabell:
Closing in favour of #1207 which switches this env along with others

PR #1155, opened 2025-09-22
Fix log ID check to correctly insert seen IDs

derrix060:
The condition if !seen_log_ids.insert(info_log_id) was inverted.

HashSet::insert() returns:
- true if the value was not previously in the set (first time seeing this log ID)
- false if the value was already in the set (seen before)

Before: The warning was shown for every receipt with log ID 0, not just the first.
After: The warning is shown only once per unique log ID.


derrix060:
any news?

PR #1156, opened 2025-09-22
BM-1627: Add RPS limit to powv cli

Wollac:
Adds the 'rps' argument to the CLI, limiting the number of requests sent to RPC providers per second. The default is 25, which matches the Alchemy free tier.


willpote (approved):

crates/boundless-cli/src/config.rs
```diff
@@ -34,6 +34,10 @@ pub struct GlobalConfig {
     #[clap(long, env = "RPC_URL", global = true)]
     pub rpc_url: Option<Url>,
 
+    /// Limit the requests per second to the RPC endpoint
+    #[clap(long, env = "RPS", global = true, default_value_t = 25)]
```
derrix060, line 38: Not sure defaulting the value to 25 is a good idea.  I'd say that the majority of provers are not using a free RPC endpoint, so they would be affected.

Also for backwards compatibility, set the default to unlimited (current behaviour), but allow one configure to a different value

PR #1158, opened 2025-09-23
BM-1636: Extend flat period at max price. Debug dynamic adjustment timeout adjustment

willpote:
Currently if a prover waits until max price before locking an order, they have a fixed amount of time to prove the order (i.e. delta between rampUpTime and lockTimeout set in Pulumi config). This time does not scale based on order size since both rampUpTime and lockTimeout are adjusted by the same value based on cycle count.

This change makes reduce the amount that rampUpTime scales based on the number of cycles in the order. This essentially enables provers to lock at max price, and still have a dynamically set amount of time to prove the order before lock expiry.


derrix060:
Thank you folks!

PR #1160, opened 2025-09-23
BM-1638: [V-BND-VUL-002] Unify client signature validation via `_verifyClientSignature`

capossele:
This PR routes `lockRequestWithSignature` path through `_verifyClientSignature` to remove duplicate checks in `_verifyClientSignatureAndExtractProverAddress`. This change ensures uniform revert behavior and bounded gas across the two entrypoints.


contracts/snapshots/BoundlessMarketBasicTest.json
```diff
@@ -17,14 +17,14 @@
...
-  "lockinRequest: base case": "147097",
-  "lockinRequest: with prover signature": "156739",
-  "priceAndFulfill: a single request": "109202",
-  "priceAndFulfill: a single request (smart contract signature)": "115396",
-  "priceAndFulfill: a single request (with selector)": "111513",
-  "priceAndFulfill: a single request that was not locked": "109202",
-  "priceAndFulfill: a single request that was not locked fulfilled by prover not in allow-list": "109202",
-  "priceAndFulfill: fulfill already fulfilled was locked request": "107510",
+  "lockinRequest: base case": "147100",
+  "lockinRequest: with prover signature": "166667",
```
willpote: I think this is why it was implemented as is, new implementation hashes twice.

If theres a clean way to refactor to only hash once that would be ideal

capossele: I've made a tweak and now is slightly better but still more expensive than before. I don't think the issue is with additional hash though, rather due to calling the internal function

capossele: You were right, I've refactored and now things look better

contracts/src/BoundlessMarket.sol
```diff
@@ -904,19 +904,17 @@ contract BoundlessMarket is
...
+        view
+        returns (address proverAddress)
+    {
         bytes32 proofRequestEip712Digest = request.eip712Digest();
         bytes32 lockRequestHash =
             _hashTypedDataV4(LockRequestLibrary.eip712DigestFromPrecomputedDigest(proofRequestEip712Digest));
         proverAddress = ECDSA.recover(lockRequestHash, proverSignature);
 
-        return (requestHash, proverAddress);
+        return proverAddress;
```
willpote: No need for this return statement as we're using a named return

PR #1161, opened 2025-09-23
BM-1639: [V-BND-VUL-003] Add check to reject non-canonical signatures

capossele:
This PR adds a check to the rust library to enforce canonical signatures. This matches with the same rule implemented by OpenZeppelin’s `ECDSA.recover`.


crates/boundless-cli/src/bin/boundless-ffi.rs
```diff
@@ -119,6 +119,10 @@ async fn main() -> Result<()> {
         .map_err(|_| anyhow::anyhow!("Failed to decode ProofRequest from input"))?;
     let signature =
         Signature::try_from(Bytes::from_hex(args.signature.trim_start_matches("0x"))?.as_ref())?;
+    // Check if the signature is non-canonical.
+    if signature.normalize_s().is_some() {
+        bail!("invalid signature: non-canonical s value");
```
Wollac: 'Non-canonical' sounds a bit vacuous. I think 'not normalised s-value' would be clearer.

PR #1170, opened 2025-09-24
BM-1651: Rename POVW_LOG_ID to REWARDS_ADDRESS

willpote:
Renames POVW_LOG_ID to REWARDS_ADDRESS. For env variables in Bento maintains backward compatibility.

For the CLI there are some breaking changes- we are planning a massive refactor to the CLI anyway, so I did not think it was worth the investment to make this version fully backward compatible (though I did most of it anyway)

Updates documentation also


willpote:
This was merged into another branch

PR #1171, opened 2025-09-24
[V-BND-VUL-005] add partial payment

capossele:
This PR refactors the `_fulfillAndPayWasLocked` function to partially pay the prover in the case a second request with same id but different hash than the originally locked one is priced and fulfilled, and the client does not have enough balance to cover for the delta between original lockPrice and the new price.


contracts/src/BoundlessMarket.sol
```diff
@@ -566,9 +573,12 @@ contract BoundlessMarket is
 
         requestLocks[id].setProverPaidAfterLockDeadline(assessorProver);
         if (MARKET_FEE_BPS > 0) {
-            price = _applyMarketFee(price);
+            finalPrice = _applyMarketFee(finalPrice);
+        }
+        accounts[assessorProver].balance += finalPrice;
+        if (partialPayment) {
```
willpote, line 590: I _think_ if `finalPrice != price`, it must be a partial payment, so could drop this variable

PR #1175, opened 2025-09-26
BM-1662: Zeroecco/prover cluster

zeroecco:
this re-works the prover cluster to use a pre-baked packer AMI. 
Overall this brings boundless _closer_ to being able to use agnostic providers.
uses vector to send logs to cw
uses prebuilt binaries
pipelines etc in place
fix env vars in the client that are not working


infra/packer/config_files/vector.yaml
```diff
@@ -0,0 +1,24 @@
...
+sinks:
+  cloudwatch_bento:
+    type: aws_cloudwatch_logs
+    inputs: ["extract_message"]
+    group_name: "/boundless/bento"
+    stream_name: "%Y-%m-%d"
+    region: "us-west-2"
+    encoding:
+      codec: json
+
```
willpote, line 24: Does this mean that if we are running multiple clusters in the same AWS account, they will all log to the same log group? 

We definitely need some way to enable each cluster to log to its own group- aside from pure usability issues (note that we deploy infra for multiple markets to the same account, so we'd have cluster logs from different chains all being mixed up with the current setup!), having them all flow to the same group will make setting up alerts based on logs very difficult

infra/packer/config_files/vector.yaml
```diff
@@ -0,0 +1,24 @@
...
+    type: remap
+    inputs: ["journald"]
+    source: |
+      . = .message
+
+sinks:
+  cloudwatch_bento:
+    type: aws_cloudwatch_logs
+    inputs: ["extract_message"]
+    group_name: "/boundless/bento"
```
willpote, line 19: I just had a peak at the log group and don't see the Bento logs there- just broker logs. How can we get bento logs? And should this group be called `/boundless/broker`?

infra/packer/README.md
```diff
@@ -0,0 +1,205 @@
...
+- **RISC0**: Groth16 prover support
+- **Vector**: Log forwarding to CloudWatch
+- **AWS Tools**: CLI v2, SSM Agent, CloudWatch Agent
+
+### **Logging Configuration**
+
+- **Vector**: Configured to forward logs to CloudWatch
+- **Destination**: `/boundless/bento` log group
+- **Streams**: Daily rotation with format `%Y-%m-%d`
+- **Services**: Captures logs from `bento.service` and `bento-broker.service`
```
willpote, line 155: Oh? I didnt see Bento in your dev stacks log group, maybe you disabled logging from it?

infra/packer/README.md
```diff
@@ -0,0 +1,205 @@
...
+  bento.pkr.hcl
+```
+
+## AMI Sharing
+
+### **Automatic Sharing**
+
+- AMIs are automatically shared with service accounts after build
+- Lambda function handles cross-account sharing
+- EventBridge triggers sharing on successful build
```
willpote, line 128: Really? Adding a note to swing back to this

infra/pipelines/pipelines/packer.ts
```diff
@@ -0,0 +1,165 @@
...
+        development: string;
+        staging: string;
+        production: string;
+    };
+}
+
+// The name of the app that we are deploying. Must match the name of the directory in the infra directory.
+const APP_NAME = "packer";
+// The branch that we should deploy from on push.
+const BRANCH_NAME = "zeroecco/prover_cluster";
```
willpote: Reminder to change to main before merge

infra/pipelines/pipelines/packer.ts
```diff
@@ -0,0 +1,165 @@
...
+      - echo "Building AMI with Packer and sharing with service accounts..."
+      - cd infra/$APP_NAME
+      - echo "Initializing Packer plugins..."
+      - packer init bento.pkr.hcl
+      - echo "Building AMI with Packer..."
+      - packer build -var "aws_region=us-west-2" -var "boundless_bento_version=$BOUNDLESS_BENTO_VERSION" -var "boundless_broker_version=$BOUNDLESS_BROKER_VERSION" -var "service_account_ids=[\"$DEVELOPMENT_ACCOUNT_ID\",\"$STAGING_ACCOUNT_ID\",\"$PRODUCTION_ACCOUNT_ID\"]" bento.pkr.hcl
+  post_build:
+    commands:
+      - echo "AMI build and sharing completed successfully"
+      - echo "AMI ID will be available in the build logs"
```
willpote, line 44: I don't quite see how we do the sharing cross-account. Maybe I missed something in the packer infra, but if you could break it down that would be useful

infra/pipelines/pipelines/prover-cluster.ts
```diff
@@ -0,0 +1,233 @@
...
+    // Stack name mappings for different environments
+    stagingStackName?: string;
+    stagingStackName2?: string;
+    productionStackName?: string;
+}
+
+// The name of the app that we are deploying. Must match the name of the directory in the infra directory.
+const APP_NAME = "prover-cluster";
+// The branch that we should deploy from on push.
+const BRANCH_NAME = "zeroecco/prover_cluster";
```
willpote: .

infra/pipelines/pipelines/prover-cluster.ts
```diff
@@ -0,0 +1,233 @@
...
+                        {
+                            name: "BOUNDLESS_BENTO_VERSION",
+                            type: "PLAINTEXT",
+                            value: boundlessBentoVersion || "v1.0.1"
+                        },
+                        {
+                            name: "BOUNDLESS_BROKER_VERSION",
+                            type: "PLAINTEXT",
+                            value: boundlessBrokerVersion || "v1.0.0"
+                        },
```
willpote: Why do we need this here in Codebuild? Shouldn't this just be config in the prover-cluster Pulumi stack?

infra/prover-cluster/index.ts
```diff
@@ -1,7 +1,661 @@
...
+        {
+            name: "name",
+            values: [`boundless-v*-ubuntu-24.04-nvidia-*`]
+        }
+    ]
+});
+
+const imageId = boundlessAmi.then(ami => ami.id);
+
+// 1) Instance role & profile (SSM access)
```
willpote: I see, we always deploy the latest AMI, which must be the latest _released_ version?

Also seems we don't support multiple clusters?

Will leave my thoughts in the overall review comments

willpote (commented):
Overall this looks awesome!!! Very excited about this!

I do have a few asks though:

1/ Currently the provers that we run from our pipelines build and deploy off main. This means that we are constantly testing the latest changes to the code in staging/prod, and being alerted if there are any bugs. This helps us catch bugs before we do a release, so is a super important property, and something I think we should keep. 

Is it possible to have packer build an AMI on every merge to main, that our prover cluster can use? Alternatively could do a nightly image release of whatever is on main, and have our cluster point to that?

2/ That being said, we do definitely want the ability to spin up a "stable" cluster, based off the release branch, that we can use for example as a backstop for specific customers. E.g. for the Bob request, we would want to service them using stable rather than a nightly release. So I would also like the ability for our prover cluster pipeline to deploy a cluster based on the stable release AMIs also!

Which brings me to my third ask..

3/ Can we make the prover-cluster Pulumi stack:
  a/ able to deploy multiple clusters
  b/ where each cluster can be configured to use latest stable AMI, or latest beta AMI (based off main)

Essentially I want it to be able to do something similar to what we do here: https://github.com/boundless-xyz/boundless/blob/main/infra/prover/index.ts#L55-L103 for spinning up new clusters. So making the cluster code a re-usable Pulumi component where you can specify image, toml, etc. Then we can do stuff like:

```
const nightlyCluster = new ProverCluster({ name: 'nightly-cluster', ami: nightly, toml: "../beta-toml.toml"});
const stableCluster = new ProverCluster({ name: 'stable-cluster', ami: latest, toml: "../stable-toml.toml"});
const clusterForBob = new ProverCluster({ name: 'bob-cluster', ami: latest, toml: "../prioritize-bob.toml"});
```

infra/pipelines/pipelines/prover-cluster.ts
zeroecco: `1c0be7b2`

infra/packer/config_files/vector.yaml
zeroecco: it is actually replaced here:
https://github.com/boundless-xyz/boundless/blob/zeroecco/prover_cluster/infra/prover-cluster/index.ts#L303

infra/packer/README.md
zeroecco: packer does this under the hood:
https://github.com/boundless-xyz/boundless/blob/zeroecco/prover_cluster/infra/packer/bento.pkr.hcl#L30-L34

infra/pipelines/pipelines/packer.ts
zeroecco: packer does this under the hood:
https://github.com/boundless-xyz/boundless/blob/zeroecco/prover_cluster/infra/packer/bento.pkr.hcl#L30-L34

infra/pipelines/pipelines/prover-cluster.ts
zeroecco: yup can convert it over once ready for merge.

infra/prover-cluster/index.ts
zeroecco: image is def the latest. I can easily add a versioning system.

infra/packer/README.md
zeroecco: the original dev stack doesn't support this but I couldn't tear it down due external factors. a redeployment of my dev stack will fix it.

infra/pipelines/pipelines/prover-cluster.ts
```diff
@@ -0,0 +1,223 @@
...
+                        {
+                            name: "GITHUB_TOKEN",
+                            type: "PLAINTEXT",
+                            value: githubToken
+                        },
+                        ...(amiId ? [{
+                            name: "AMI_ID",
+                            value: amiId,
+                        }] : []),
```
willpote, line 102: Don't think this is used

infra/prover-cluster/README.md
```diff
@@ -1,296 +1,144 @@
...
-
-```bash
-pulumi login --local
-pulumi stack init <stack-name>
-pulumi stack select <stack-name>
-pulumi config set aws:region us-west-2
+infra/prover-cluster/
+├── templates/            # Cloud-init template files
+│   ├── manager.yaml      # Manager node cloud-init template
+│   └── prover.yaml       # Prover node cloud-init template
```
willpote, line 36: ?

infra/prover-cluster/Pulumi.prod-8453.yaml
```diff
@@ -0,0 +1,33 @@
...
+  prover-cluster:minioPassword:
+    secure: v1:qdMFDop9G60QnG9t:HOT+pX2MyrELiYUt3YgsaHMdxTxWgKEpRAto1jh6pszFNu0=
+  prover-cluster:executionCount: "4"
+  prover-cluster:auxWorkerCount: "4"
+  prover-cluster:proverWorkerCount: "15"
+  prover-cluster:privateKey:
+    secure: v1:bZaP8gZLLn+7PG9z:9u7k5/ch7P3WCrI6oaFzUpsljSw3ZiOBKcrM3mQL9Tg09HdrjDK8kM5FpWJ8nSfTouzr2Ghfiyj+KGcJgXYWFQQXbj2SXGVc8HMl0LbPdkk=
+  prover-cluster:ethRpcUrl:
+    secure: v1:lFT0b4jaqt9DTTz8:QUmzfFkJTx753kerYTsYteef43jKTjRsk4GUjgW2vfV41yRzarZ1YVGFJ3QyXfmC8fsfTezQ7JwcEFdprqmkkTu+NaN9wrkb2aZsEdwahxxEIuLUD+DMR1vSO7IgVbd3MmunrP2Dm7ufmVlbzhsTXQ==
+  prover-cluster:orderStreamUrl: "https://testnet.boundless.network/"
```
willpote: base-mainnet.boundless.network

.github/workflows/nightly-ami.yml
```diff
@@ -0,0 +1,96 @@
+name: Nightly AMI Build
```
willpote: Putting the nightly build into GH actions puts us in a state where we do our release builds on AWS (+ all our other builds for all our other services), and only our nightly build in GH actions. 

I think we should move this to CodePipeline also to keep everything consistent and let us re-use the existing code, and use an event bridge rule to trigger the build nightly: https://docs.aws.amazon.com/codepipeline/latest/userguide/pipelines-trigger-source-schedule.html

.github/workflows/nightly-ami.yml
```diff
@@ -0,0 +1,96 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
```
willpote: Another reason- we really want to minimize the creds we need in Github actions. That was one of the big motivations for us putting all our deployment infra into AWS rather than GH actions

infra/packer/config_files/vector.yaml
zeroecco: there is a sed in userdata to replace this.

PR #1177, opened 2025-09-29
feat: Implement log ID validation for receipts

derrix060:
Add log ID check before pushing receipt to work_receipts. There are some cases where the log id might differ.

For example:

```
 2025-09-21T16:52:12.760155Z  INFO boundless_cli::commands::povw::prepare: Loaded 625 work receipts
 Error: Failed to prove work log update
 ﻿
 Caused by:
 work claim nonce min log id is not equal to the expected log id: 918822xx != 1347xx
```


crates/boundless-cli/src/commands/povw/prepare.rs
```diff
@@ -112,7 +112,15 @@ impl PovwPrepare {
...
-                Ok(receipt) => work_receipts.push(receipt),
+                Ok(receipt) => {
+                    let log_id = receipt.claim().value().unwrap().work.value().unwrap().nonce_min.log;
+                    if log_id != state.log_id {
+                        tracing::warn!("Skipping receipt with log ID: {:?} != state log ID: {:?}", log_id, state.log_id);
+                    }
+                    else {
+                        work_receipts.push(receipt.clone());
+                    }
+                }
```
austinabell:
```suggestion
                Ok(receipt) => match check_work_receipt(state.log_id, &state.work_log, receipt) {
                    Ok(receipt) => {
                        work_receipts.push(receipt);
                    }
                    Err(err) => {
                        tracing::warn!("Skipping receipt: {:#}", err);
                    }
                },
```

Just to re	use existing logic and avoid a panic in case of malformed receipt

austinabell:
thank you!

PR #1182, opened 2025-09-29
Clarify requirements for povw state file 

mintybasil:
This change seeks to clarify that a Log ID can effectively only be associated with a single state file. The current wording otherwise suggests that the state file only needs to be maintained until the work is submitted.


sasha-computer:
thanks for this, this has been added in https://github.com/boundless-xyz/boundless/pull/1190.

PR #1183, opened 2025-09-30
BM-1669: Auto-mining with unused capacity support

austinabell:
Still have to test the prepare/submit proofs through this, to ensure a 0 reservation job doesn't conflict with it.

### DIFF
--- a/bento/crates/api/src/helpers.rs
+++ b/bento/crates/api/src/helpers.rs
@@ -14,13 +14,12 @@ use workflow_common::{
 /// Format: "reserved:N" where N is a positive number (e.g., "reserved:5" or "reserved:10")
 /// Returns 0 if no reserved value is specified
 pub fn extract_reserved(api_key: &str) -> i32 {
-    if let Some(reserved_str) = api_key.strip_prefix("reserved:") {
-        if let Ok(reserved) = reserved_str.parse::<i32>() {
+    if let Some(reserved_str) = api_key.strip_prefix("reserved:")
+        && let Ok(reserved) = reserved_str.parse::<i32>() {
             if reserved >= 0 {
                 return reserved;
             }
         }
-    }
     0
 }
 
--- a/bento/crates/api/src/helpers.rs
+++ b/bento/crates/api/src/helpers.rs
@@ -11,15 +11,13 @@ use workflow_common::{
 };
 
 /// Extract reserved value from api_key string
-/// Format: "reserved:N" where N is a positive number (e.g., "reserved:5" or "user123-reserved:10")
-/// Returns 1 if no reserved value is specified
+/// Format: "reserved:N" where N is a positive number (e.g., "reserved:5" or "reserved:10")
+/// Returns 0 if no reserved value is specified
 pub fn extract_reserved(api_key: &str) -> i32 {
-    for part in api_key.split('-') {
-        if let Some(reserved_str) = part.strip_prefix("reserved:") {
-            if let Ok(reserved) = reserved_str.parse::<i32>() {
-                if reserved > 0 {
-                    return reserved;
-                }
+    if let Some(reserved_str) = api_key.strip_prefix("reserved:") {
+        if let Ok(reserved) = reserved_str.parse::<i32>() {
+            if reserved >= 0 {
+                return reserved;
             }
         }
     }
--- a/bento/crates/api/src/helpers.rs
+++ b/bento/crates/api/src/helpers.rs
@@ -11,15 +11,15 @@ use workflow_common::{
 };
 
 /// Extract reserved value from api_key string
-/// Format: "reserved:N" where N is a positive number (e.g., "reserved:5" or "reserved:10")
+/// Format: "v1:reserved:N" where N is a positive number (e.g., "v1:reserved:5" or "v1:reserved:10")
 /// Returns 0 if no reserved value is specified
 pub fn extract_reserved(api_key: &str) -> i32 {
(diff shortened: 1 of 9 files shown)


crates/boundless-cli/src/config.rs
```diff
@@ -175,8 +175,8 @@ impl ProverConfig {
         if let Some(ref api_key) = self.bento_api_key {
             std::env::set_var("BONSAI_API_KEY", api_key);
         } else {
-            tracing::debug!("No API key provided. Setting BONSAI_API_KEY to empty string");
-            std::env::set_var("BONSAI_API_KEY", "");
+            tracing::debug!("No API key provided. Setting BONSAI_API_KEY to reserved:1");
+            std::env::set_var("BONSAI_API_KEY", "reserved:1");
```
ec2: This probably doesn't matter, but before, by default, the value was 0, so maybe its worth maintaining that behaviour?

crates/broker/src/lib.rs
```diff
@@ -825,7 +825,7 @@ where
             tracing::info!("Configured to run with Bento backend");
 
             Arc::new(
-                provers::Bonsai::new(config.clone(), bento_api_url.as_ref(), "")
+                provers::Bonsai::new(config.clone(), bento_api_url.as_ref(), "reserved:999")
```
ec2: This can starve out all the proof requests made by the boundless-cli (reserved=1) unless people know to manually set the it when making a boundless-cli call. I can see a bunch of people complaining about why their povw prepares and submits are failing or never completing lol

ec2 (changes requested):

crates/boundless-cli/src/config.rs
austinabell: why? This would mean these proofs would never get any resources or complete if orders are taken? This changes the behaviour from what currently exists, where they are equal priority

crates/broker/src/lib.rs
austinabell: It just means it will only use one prove agent at a time, to avoid starving the prove agents from completing orders. I could probably increase the reserved value a bit so that it doesn't cap out at 1, and so that when capped it doesn't conflict with mining, but I don't see what issue you have with setting this high of a reservation. This is needed to ensure any non-reserved work does not use resources for orders.

bento/crates/api/src/helpers.rs
```diff
@@ -10,18 +10,33 @@ use workflow_common::{
     AUX_WORK_TYPE, COPROC_WORK_TYPE, EXEC_WORK_TYPE, ExecutorResp, JOIN_WORK_TYPE, PROVE_WORK_TYPE,
 };
 
+/// Extract reserved value from api_key string
+/// Format: "reserved:N" where N is a positive number (e.g., "reserved:5" or "reserved:10")
+/// Returns 0 if no reserved value is specified
+pub fn extract_reserved(api_key: &str) -> i32 {
+    if let Some(reserved_str) = api_key.strip_prefix("reserved:") {
```
willpote: Any reason we make it so it has to be a prefix? Just feels more brittle if we ever need to use api_key/user_id to encode additional data

bento/crates/api/src/helpers.rs
```diff
@@ -10,18 +10,33 @@ use workflow_common::{
     AUX_WORK_TYPE, COPROC_WORK_TYPE, EXEC_WORK_TYPE, ExecutorResp, JOIN_WORK_TYPE, PROVE_WORK_TYPE,
 };
 
+/// Extract reserved value from api_key string
+/// Format: "reserved:N" where N is a positive number (e.g., "reserved:5" or "reserved:10")
+/// Returns 0 if no reserved value is specified
+pub fn extract_reserved(api_key: &str) -> i32 {
```
willpote, line 16: Not too familiar with convention in Bento, but it seems `api-key == user-id`? If so, lets just use one term in each code base, so rename to user id here?

crates/broker/src/lib.rs
willpote: Have the same concern as Eric. I don't know the details of how the reserved mechanism works, but it sounds like you are saying that even if there is a backlog of proofs with '999 reserved', a prepare/submit proof with 1 reserved will get weaved in in a reasonable time frame - if this is the case then sounds good!

bento/crates/api/src/helpers.rs
austinabell: Would this not be backwards compatible? Could just parse a v2 prefix if we wanted to support more in the future? What is the suggestion to do instead?

bento/crates/api/src/helpers.rs
austinabell: Felt more correct to make it referenced as API key because the user id semantics might be a subset of the API key (currently a bit unused). Happy to switch though

crates/broker/src/lib.rs
austinabell: That is correct. 1 reserved would be prioritized if at least 1 GPU prove agent is being used for the other. Because avoiding taskdb changes ATM, this leads to a bit unideal situation with only 1 GPU, because it would just be ordered by segment creation time. I changed this value to 5, so this would by default use up to 5 gpus, and I'm biasing to lower amount to not disrupt the broker capacity assumption, but I think I might increase it more such that it doesn't get delayed by mining too much

bento/crates/api/src/helpers.rs
willpote: Versioning alleviates my concern here, think we should include v1 prefix in this just to keep things clean

PR #1210, opened 2025-10-03
Support requestor lists in broker

willpote:
Create 'requestor lists' - a format for creating lists of Boundless requestors, inspired by Token Lists.

Adds support for fetching and refreshing requestor lists to the broker, specifically to be used in the priority requestor functionality.

With this we can publish our recommended lists of priority requestors, and provers can opt into consuming one/many of the lists we publish. This will allow us to onboard new partners quicker, as we can simply update the list to enable provers to start prioritizing important work.


crates/broker/src/requestor_monitor.rs
```diff
@@ -0,0 +1,459 @@
...
+        };
+
+        if urls.is_empty() {
+            return Ok(());
+        }
+
+        tracing::debug!("Refreshing {} requestor priority lists", urls.len());
+
+        // Clear existing entries before refreshing
+        self.priority_requestors.clear();
```
austinabell: Optional: This seems problematic, because on refresh where a fetch fails, it will clear existing configs. Is this intentional? Probably fine as is, but seems like mapping URL to the config could allow this to be handled better in case that their URL they are using goes down for any reason

crates/requestor-lists/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    /// Fetch and parse a requestor list from a URL
+    pub async fn fetch_from_url(url: &str) -> Result<Self, ValidationError> {
+        let response =
+            reqwest::get(url).await.map_err(|e| ValidationError::FetchError(e.to_string()))?;
+
+        let json = response.text().await.map_err(|e| ValidationError::FetchError(e.to_string()))?;
+
+        Self::from_json(&json)
```
austinabell, line 212: https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.json might be cleaner than going to text first

crates/requestor-lists/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+        version: Version,
+        requestors: Vec<RequestorEntry>,
+    ) -> Self {
+        Self { name, description, schema_version: CURRENT_SCHEMA_VERSION, version, requestors }
+    }
+
+    /// Validate the list according to schema rules
+    pub fn validate(&self) -> Result<(), ValidationError> {
+        // Check schema version compatibility
+        if self.schema_version.major != CURRENT_SCHEMA_VERSION.major {
```
austinabell, line 150: just want to confirm avoiding minor version compatibility is intended?

crates/broker/src/lib.rs
```diff
@@ -957,6 +964,15 @@ where
...
         });
 
+        // Start the RequestorMonitor to periodically fetch priority lists
+        let requestor_monitor = Arc::new(requestor_monitor::RequestorMonitor::new(
+            self.priority_requestors.clone(),
+            non_critical_cancel_token.clone(),
+        ));
+        let monitor_handle = requestor_monitor.clone().spawn();
+        supervisor_tasks
+            .spawn(async move { monitor_handle.await.context("Requestor list monitor panicked")? });
```
austinabell: Is this intentionally not being wrapped in the `Supervisor` struct like the others? Just that this won't restart on a panic I think would be the main difference in this case.

crates/requestor-lists/src/lib.rs
austinabell: resolving, not cleaner because of the validate, and not worth the change for potential micro opt

crates/broker/src/requestor_monitor.rs
willpote: I went back and forth on what the correct behavior here should be, as removing a list publisher may want to just delete a list from circulation to try and stop consumers from using it immediately.

I think what you propose is likely safer, and list publishers can just update their lists to be empty when they want to unpublish a list, so have updated the code to track by URL

crates/broker/src/lib.rs
willpote: This was a miss. Thanks for adding this

PR #1218, opened 2025-10-07
add db clean up script

zeroecco:
🎉 find stuck jobs, and unstick them.
🎉 clean up the DB of completed jobs.


bento/crates/taskdb/migrations/3_maintenance.sql
```diff
@@ -0,0 +1,95 @@
...
+
+
+-- clear out completed jobs
+CREATE OR REPLACE FUNCTION clear_completed_jobs()
+RETURNS INTEGER AS $$
+DECLARE
+    total_cleared INTEGER := 0;
+BEGIN
+    DELETE FROM task_deps WHERE job_id IN (SELECT id FROM jobs WHERE state = 'done');
+    DELETE FROM tasks WHERE job_id IN (SELECT id FROM jobs WHERE state = 'done');
```
austinabell, line 90: Does this sql get deduplicated/optimized, or would it not make sense for the select to only be called once?

Sth like:
```
      SELECT ARRAY_AGG(id) INTO completed_job_ids
      FROM jobs WHERE state = 'done';
      DELETE FROM task_deps WHERE job_id = ANY(completed_job_ids);
      DELETE FROM tasks WHERE job_id = ANY(completed_job_ids);
      DELETE FROM jobs WHERE id = ANY(completed_job_ids);
```

Main concern being partial deletions if a job is completed between these calls, where some data would never be deleted. Assuming you didn't wrap in a transaction for perf

bento/crates/workflow/src/lib.rs
```diff
@@ -247,6 +247,26 @@ impl Agent {
...
+                    .expect("Stuck task maintenance failed")
+            });
+
+            // Enable completed job cleanup for aux workers
+            let term_sig_copy = term_sig.clone();
+            let db_pool_copy = self.db_pool.clone();
+            tokio::spawn(async move {
+                Self::poll_for_completed_job_cleanup(term_sig_copy, db_pool_copy)
+                    .await
+                    .expect("Completed job cleanup failed")
```
austinabell: These errors aren't actually handled and are ignored. Is this intended?

bento/crates/workflow/src/lib.rs
```diff
@@ -419,4 +439,66 @@ impl Agent {
...
+    /// Check for tasks that are stuck in pending state but should be ready
+    /// because all their dependencies are complete, and fix them.
+    async fn poll_for_stuck_tasks(term_sig: Arc<AtomicBool>, db_pool: PgPool) -> Result<()> {
+        while !term_sig.load(Ordering::Relaxed) {
+            tracing::debug!("Checking for stuck pending tasks...");
+
+            // First check if there are any stuck tasks
+            let stuck_tasks = taskdb::check_stuck_pending_tasks(&db_pool).await?;
+            if !stuck_tasks.is_empty() {
+                tracing::info!("Found {} stuck pending tasks", stuck_tasks.len());
```
austinabell:
```suggestion
                tracing::warn!("Found {} stuck pending tasks", stuck_tasks.len());
```

Opinionated, if you think this shouldn't be, feel free to not make this change

bento/crates/workflow/src/lib.rs
```diff
@@ -419,4 +439,66 @@ impl Agent {
...
+
+                // Fix the stuck tasks
+                let fixed_count = taskdb::fix_stuck_pending_tasks(&db_pool).await?;
+                if fixed_count > 0 {
+                    tracing::info!("Fixed {} stuck pending tasks", fixed_count);
+                }
+            }
+
+            // Sleep for 1 minute before next check
+            time::sleep(tokio::time::Duration::from_secs(60)).await;
```
austinabell: Having this be configurable might be nice, in case someone were to not want this to run or if we find it causes issues

bento/crates/workflow/src/lib.rs
```diff
@@ -419,4 +439,66 @@ impl Agent {
...
+        while !term_sig.load(Ordering::Relaxed) {
+            tracing::debug!("Cleaning up completed jobs...");
+
+            let cleared_count = taskdb::clear_completed_jobs(&db_pool).await?;
+            if cleared_count > 0 {
+                tracing::info!("Cleared {} completed jobs", cleared_count);
+            }
+
+            // Sleep for 1 hour before next cleanup
+            time::sleep(tokio::time::Duration::from_secs(60 * 60)).await;
```
austinabell: same as above comment, especially so since someone may be using bento to create proofs with assumptions and might need to turn this off to re-use completed old receipts.

bento/crates/workflow/src/lib.rs
```diff
@@ -419,4 +439,66 @@ impl Agent {
...
+    /// Remove completed jobs and their associated tasks and dependencies
+    /// to prevent database bloat over time.
+    async fn poll_for_completed_job_cleanup(
+        term_sig: Arc<AtomicBool>,
+        db_pool: PgPool,
+    ) -> Result<()> {
+        while !term_sig.load(Ordering::Relaxed) {
+            tracing::debug!("Cleaning up completed jobs...");
+
+            let cleared_count = taskdb::clear_completed_jobs(&db_pool).await?;
```
austinabell, line 553: Also one last concern is that if someone hadn't run this in a while, this might be a slow query when they upgrade. Unsure what the perf would look like, just making sure you have accounted for that

bento/crates/workflow/src/lib.rs
zeroecco: yes. if clean up fails it shouldn't take aux down

bento/crates/workflow/src/lib.rs
zeroecco: you are correct, def a def a slow query, but this shouldn't take an enormous amount of time and the only thing that will happen is that psql will be a slow till everything is cleaned up

bento/crates/taskdb/migrations/3_maintenance.sql
zeroecco: super nerd moment 😆 
A function always gets called from one statement. postgresql, to treat the function as a single action will wrap the function and all the functions it calls again run in a single transaction as the statement that called the initial function, UNLESS you explicitly add breaks and seperate logic.

the beauty of this call is that it is fully point in time and locked to only completed jobs and its dependencies. A job cannot be complete until all its deps are complete.

bento/crates/workflow/src/lib.rs
zeroecco: be2c4f26

bento/crates/workflow/src/lib.rs
zeroecco: be2c4f26

bento/crates/workflow/src/lib.rs
austinabell: not saying that it should, just saying it doesn't even log the error that you are generating, am I misreading?

zeroecco: good point:
45fbfa9a

austinabell: Sorry, I just meant that the error would have just been printed to stderr and would not run again, this commit is actually a regression because it doesn't even print out the panic anymore (context just wraps the error with that prefix, but then it gets dropped because nothing awaiting that task).

If you would like it to fail once and not start up, you probably want to just revert that last commit, but intuitively feels like you want something like:
```
      loop {
          if let Err(e) = poll_for_completed_job_cleanup(term_sig_copy, db_pool_copy) {
              tracing::error!("Completed job cleanup failed: {:#}", e);
              
              // And maybe a sleep to avoid it from spinning
              time::sleep(time::Duration::from_secs(60)).await;
          }
      }

```

Given that some temporal errors could be hit within taskdb? Maybe there is a good reason for some of these not to retry after one error?

PR #1228, opened 2025-10-09
BM-1723: add sanity checks into broker for aggregation proofs

austinabell:
Has been noticed that some proofs are transiently generated invalid and only caught by on-chain verification. This change will help us see where but also be able to salvage partial batches or retry g16 generation if that was a failure. Retrying a full job is more complicated and would have to happen in another change.


bento/crates/workflow-common/src/lib.rs
```diff
@@ -26,8 +26,8 @@ pub const KECCAK_RECEIPT_PATH: &str = "keccak_receipts";
 /// join worker stream identifier
 pub const JOIN_WORK_TYPE: &str = "join";
 
-pub const SNARK_RETRIES: i32 = 0;
-pub const SNARK_TIMEOUT: i32 = 60 * 2;
+pub const SNARK_RETRIES_DEFAULT: i32 = 3;
+pub const SNARK_TIMEOUT_DEFAULT: i32 = 60;
```
willpote, line 30: Not able to comment on these numbers, @zeroecco should have a look

austinabell:
> How often is this happening? At the moment these errors will flag as unexpected errors, so wondering if this will trigger lots of alerts, and whether we have a custom error for it?

It should be very uncommon, as it is very difficult to reproduce. In these cases we would still get the error, but on submit of batch rather than during prove. I don't think we have seen this with any of our provers in our infra, but it's possible one was missed.

bento/crates/workflow-common/src/lib.rs
austinabell: We agreed on these values in DMs and this has been tested for about a week now from a prover that was hitting this error more frequently. Will give some time for Rich to comment, but I don't think we need to delay pulling in if we want to release soon.

PR #1236, opened 2025-10-14
BM-1741: Add API key with higher rate limits

willpote:
Will be used for backfills by the explorer

### DIFF
--- a/infra/distributor/index.ts
+++ b/infra/distributor/index.ts
@@ -2,7 +2,7 @@ import * as aws from '@pulumi/aws';
 import * as awsx from '@pulumi/awsx';
 import * as pulumi from '@pulumi/pulumi';
 import * as docker_build from '@pulumi/docker-build';
-import { ChainId, getServiceNameV1, getEnvVar, Severity } from '../util';
+import { getServiceNameV1, getEnvVar, Severity } from '../util';
 import * as crypto from 'crypto';
 require('dotenv').config();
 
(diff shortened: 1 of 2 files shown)


infra/order-stream/components/order-stream.ts
```diff
@@ -265,92 +271,151 @@ export class OrderStreamInstance extends pulumi.ComponentResource {
...
-        allow: {},
-      },
-      name: `${serviceName}-acl`,
-      description: `WebACL for order stream service ${chainId}`,
-      customResponseBodies: [
-        {
-          key: 'rate-limit-body',
-          content: '{"error": "Rate limit exceeded. Please retry after 60 seconds."}',
-          contentType: 'APPLICATION_JSON',
+    // Build WAF rules with tiered rate limiting
```
zeroecco, line 274: only thought is that priority 0 is lower in the list then priority 1, I beleive this could lead to issues around people understanding the actual waf rule base

willpote: So the existing priority: 0 is the priority for text transformations, not for the rules as a whole. Text transformations seem to have their own priorities.

Our rules were indexed from 1. I changed them to be 0 indexed though to make things clearer

willpote: [35ba077](https://github.com/boundless-xyz/boundless/pull/1236/commits/35ba07797e9118578b08cdb2bb0351d4e38c6da6)

PR #1239, opened 2025-10-15
BM-1714: add override IPFS gateway

capossele:
Overrides the IPFS gateway in a given URL. The new gateway URL is taken from the `IPFS_GATEWAY_URL` environment variable, or defaults to `BOUNDLESS_IPFS_GATEWAY_URL` if the variable is not set

### DIFF
--- a/crates/broker/src/lib.rs
+++ b/crates/broker/src/lib.rs
@@ -662,7 +662,7 @@ where
                             Ok(bytes) => bytes,
                             Err(_) => {
                                 let overridden_url = override_gateway(&contract_url);
-                                tracing::debug!("Retrying with overridden URL: {}", overridden_url);
+                                tracing::debug!("Retrying with overridden URL: {overridden_url}");
                                 self.download_image(&overridden_url, "gateway fallback").await?
                             }
                         };
@@ -679,7 +679,7 @@ where
                         Ok(bytes) => bytes,
                         Err(_) => {
                             let overridden_url = override_gateway(&contract_url);
-                            tracing::debug!("Retrying with overridden URL: {}", overridden_url);
+                            tracing::debug!("Retrying with overridden URL: {overridden_url}");
                             self.download_image(&overridden_url, "gateway fallback").await?
                         }
                     };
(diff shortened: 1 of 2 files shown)


crates/broker/src/lib.rs
```diff
@@ -612,9 +613,16 @@ where
...
+            "assessor",
+            prover,
+            image_id,
+            override_gateway(&image_url_str),
+            path,
+            default_url,
+        )
+        .await
+        .context("uploading assessor image")?;
         Ok(image_id)
```
willpote: Ideally we try URL, if fail try custom gateway, if fail try contract specified gateway. Since our custom gateway that we pay for is now a point of failure

capossele: Good catch, addressed with https://github.com/boundless-xyz/boundless/pull/1239/commits/d95ad89bbd7c6afb4a6608c2b31b882c296e149f

crates/boundless-market/src/storage/fetch.rs
```diff
@@ -49,3 +53,43 @@ async fn fetch_file(url: &Url) -> anyhow::Result<Vec<u8>> {
...
+    }
+
+    let parts: Vec<&str> = url.splitn(2, "/ipfs/").collect();
+    if parts.len() != 2 {
+        return url.to_string();
+    }
+
+    let gateway_url =
+        std::env::var("IPFS_GATEWAY_URL").unwrap_or(BOUNDLESS_IPFS_GATEWAY_URL.to_string());
+    let new_url = format!("{gateway_url}/ipfs/{}", parts[1]);
```
austinabell, line 75:
```suggestion
    let Some((_, cid)) = url.split_once("/ipfs/") else {
        return url.to_string();
    };

    let gateway_url =
        std::env::var("IPFS_GATEWAY_URL").unwrap_or(BOUNDLESS_IPFS_GATEWAY_URL.to_string());
    let new_url = format!("{gateway_url}/ipfs/{cid}");
```

Completely optional pass-by suggestion to make this a bit cleaner

PR #1247, opened 2025-10-16
BM-1754: Zeroecco/stuck jobs

zeroecco:
the api layer for authenticated queries

### DIFF
--- a/bento/crates/workflow/src/tasks/finalize.rs
+++ b/bento/crates/workflow/src/tasks/finalize.rs
@@ -79,11 +79,5 @@ pub async fn finalize(agent: &Agent, job_id: &Uuid, request: &FinalizeReq) -> Re
         .await
         .context("Failed to upload final receipt to obj store")?;
 
-    tracing::debug!("Deleting the keyspace {job_prefix}:*");
-    // remove the redis keys after job is completed
-    redis::scan_and_delete(&mut conn, &job_prefix)
-        .await
-        .context("Failed to delete all redis keys")?;
-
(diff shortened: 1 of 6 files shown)


infra/prover-cluster/components/ApiGatewayComponent.ts
```diff
@@ -0,0 +1,177 @@
...
+        // Set the ALB URL for external access
+        this.albUrl = this.alb.dnsName.apply(dnsName => `http://${dnsName}`);
+    }
+
+    private generateApiKey(): string {
+        // Generate a random 32-character API key
+        const chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+        let result = '';
+        for (let i = 0; i < 32; i++) {
+            result += chars.charAt(Math.floor(Math.random() * chars.length));
```
willpote: Not the best practice to use a pseudorandom number generator for keys. Can we just set the API key in Pulumi state and pass it in (as we do in our other stacks for secrets),  or use something like `crypto.getRandomValues`

willpote (commented):
Whats the motivation for this? We can use SSM to ssh into machines using AWS for auth already can't we?

infra/prover-cluster/components/ApiGatewayComponent.ts
zeroecco: 944df018

willpote (approved):

PR #1249, opened 2025-10-17
BM-1756: Use replicas feature in compose.yml

zeroecco:


compose.yml
```diff
@@ -169,11 +169,10 @@ services:
       - redis
       - minio
 
-  exec_agent0:
-    <<: *exec-agent-common
-
-  exec_agent1:
+  exec_agent:
```
austinabell, line 170: Do you mind also updating the `depends_on` section to reflect this change?

zeroecco: https://github.com/boundless-xyz/boundless/pull/1249/commits/88926ed3e74ac08ec476c74f5d225a167fe3d522

PR #1251, opened 2025-10-18
BM-1758: clean up order picker logs and add context for skipped orders

austinabell:
I got annoyed with reading logs and tracing why orders are locked. Passby also changed all stake naming to collateral and lowered the warnings given when images already exist.

### DIFF
--- a/crates/broker/src/order_picker.rs
+++ b/crates/broker/src/order_picker.rs
@@ -718,7 +718,7 @@ where
                     let required_price_per_mcycle =
                         available_eth.saturating_mul(ONE_MILLION) / U256::from(proof_cycles);
                     format!(
-                        "mcycle_price currently {} ETH/Mcycle, needs to be <= {} ETH/Mcycle",
+                        "mcycle_price currently {} ETH/Mcycle, order required <= {} ETH/Mcycle",
                         format_ether(*mcycle_price_eth),
                         format_ether(required_price_per_mcycle)
                     )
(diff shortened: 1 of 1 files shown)


crates/broker/src/order_picker.rs
```diff
@@ -78,6 +78,48 @@ const PREFLIGHT_CACHE_TTL_SECS: u64 = 3 * 60 * 60; // 3 hours
...
+            ProveLimitReason::ConfigCap { max_mcycles } => {
+                write!(f, "broker max_mcycle_limit config setting ({} Mcycles)", max_mcycles)
+            }
+            ProveLimitReason::DeadlineCap { time_remaining_secs, peak_prove_khz } => {
+                write!(
+                    f,
+                    "deadline: {}s remaining with peak_prove_khz config ({} kHz)",
+                    time_remaining_secs, peak_prove_khz
+                )
+            }
```
willpote, line 118: Maybe out of scope, but I don't find these logs particularly user friendly as the units are all out of sync, e.g. would be better if we computed the mcycle price that the collateral reward gives, and compare it to the mycle price they've set in their config in the logs

austinabell: ah, that's a good take. Yeah I suppose that would be more readable than showing the calculation. Will adjust

austinabell: I can't adjust this for all, because if the session limit is hit across lock and fulfill after lock, we don't know how many cycles the order is (and thus don't know what the setting needs to be). Can improve the logs, in a slightly unclean way, when we do finish execution

austinabell: [`6b273e2` (#1251)](https://github.com/boundless-xyz/boundless/pull/1251/commits/6b273e2a3336b68726627aa46da42d1a1c62f5e4)

Testing briefly before merging, but this is what you had in mind?

PR #1253, opened 2025-10-20
BM-1761: Add new threshold for alerts for distributor

willpote:
Allows us to be notified earlier for distributor balance issues

### DIFF
--- a/crates/boundless-cli/src/bin/boundless.rs
+++ b/crates/boundless-cli/src/bin/boundless.rs
@@ -86,6 +86,7 @@ use boundless_market::{
         FulfillmentData, Offer, Predicate, ProofRequest, RequestInputType, Selector,
     },
     input::GuestEnv,
+    override_gateway,
     request_builder::{OfferParams, RequirementParams},
     selector::ProofType,
     storage::{fetch_url, StorageProvider, StorageProviderConfig},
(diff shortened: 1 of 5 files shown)


infra/distributor/index.ts
```diff
@@ -379,8 +407,8 @@ export = () => {
     alarmActions,
   });
 
-  new aws.cloudwatch.MetricAlarm(`${serviceName}-stake-alarm-${Severity.SEV2}`, {
-    name: `${serviceName}-log-stake-${Severity.SEV2}`,
+  new aws.cloudwatch.MetricAlarm(`${serviceName}-stake-alarm-${Severity.SEV1}`, {
+    name: `${serviceName}-log-stake-${Severity.SEV1}`,
```
austinabell: Distributor and then a prover running out of stake seems unlikely that it's worth being a sev1, I thought we were avoiding those for our provers?

crates/boundless-market/src/contracts/bytecode.rs
capossele, line 1: Why this change?

infra/distributor/index.ts
willpote: True, fixed

crates/boundless-market/src/contracts/bytecode.rs
willpote: Fixed

PR #1254, opened 2025-10-21
BM-1764: improve maintenance for streams

zeroecco:
:broom: clean up streams too.

### DIFF
--- a/bento/crates/taskdb/migrations/4_clear_completed_streams.sql
+++ b/bento/crates/taskdb/migrations/4_clear_completed_streams.sql
@@ -3,7 +3,6 @@ CREATE OR REPLACE FUNCTION clear_completed_jobs()
 RETURNS INTEGER AS $$
 DECLARE
     total_cleared INTEGER := 0;
-    streams_cleared INTEGER := 0;
 BEGIN
     -- Store the count before deletion
     SELECT COUNT(*) INTO total_cleared FROM jobs WHERE state = 'done';


bento/crates/taskdb/migrations/4_clear_completed_streams.sql
```diff
@@ -0,0 +1,26 @@
+-- clear out completed jobs
+CREATE OR REPLACE FUNCTION clear_completed_jobs()
+RETURNS INTEGER AS $$
+DECLARE
+    total_cleared INTEGER := 0;
+    streams_cleared INTEGER := 0;
```
austinabell: This seems unused?

austinabell (approved):
Curious the motivation for doing this just once? Is this not something we would want cleaned up on the cleanup cron?

bento/crates/taskdb/migrations/4_clear_completed_streams.sql
zeroecco: removing that deadwood

zeroecco:
> Curious the motivation for doing this just once? Is this not something we would want cleaned up on the cleanup cron?

I am just adding the streams to the function placed in the third migration call. Should already be covered in the aux job.

bento/crates/taskdb/migrations/4_clear_completed_streams.sql
zeroecco: 5f9fde9c

austinabell:
> > Curious the motivation for doing this just once? Is this not something we would want cleaned up on the cleanup cron?
> 
> I am just adding the streams to the function placed in the third migration call. Should already be covered in the aux job.

I misread, 👍

PR #1255, opened 2025-10-22
BM-1765: mine by default, opt out with env

austinabell:
Passby also reduces the default mining job size to ~1h on 5090, I've found that when consistently participating in the market, 4h is too long where it would be inconsistent where it is included in the epoch. This is an updated guess of what the default could/should be.

- Updates `just broker` to `just prover` (broker was already overloaded given it started bento also)
- Removes `just mine` command and rolls it into `just prover`
  - `BOUNDLESS_MINING=false` to disable the mining process
- Adjusts default mining job size as mentioned above

### DIFF
--- a/justfile
+++ b/justfile
@@ -431,6 +431,9 @@ prover action="up" env_file="" detached="true":
     if [ "{{action}}" = "logs" ]; then
         # Ignore mining process logs by default
         PROFILE_FLAGS="--profile broker"
+    elif [ "{{action}}" = "down" ] || [ "{{action}}" = "clean" ]; then
+        # Always include miner profile when shutting down to ensure no lingering mining process
+        PROFILE_FLAGS="--profile broker --profile miner"
     elif [ "$BOUNDLESS_MINING" = "false" ]; then
         PROFILE_FLAGS="--profile broker"
     else


compose.yml
```diff
@@ -280,7 +280,7 @@ services:
     environment:
       <<: *base-environment
       RUST_LOG: ${RUST_LOG:-info}
-    entrypoint: /bin/bash -c "while sleep 1; do /app/bento_cli --iter-count 2000000; done"
```
willpote, line 283: Ideally we want this value as low as possible correct? What risks do we add or trade-off when lowering it?

Separately would it make sense to set this dynamically (via the wizard) based on cluster size?

willpote (approved):

compose.yml
austinabell: > What risks do we add or trade-off when lowering it?
- If running a lot of GPUs, the prove rate will be lower with smaller jobs (smaller effect for smaller provers)
- Less efficient for minio storage of completed proofs
- More proofs to roll up when doing povw prepare

Hard to pick a default that works for all systems, but this is just a guess at what most provers would find unoffensive. Perhaps something configurable with the config wizard, but unsure it's worth configuring until we know what proof sizes provers prefer.

PR #1262, opened 2025-10-24
BM-1774: update pricing modes to weight randomness by collateral reward

austinabell:
- Updates default to profitability mode
- Updates the pricing mode names to also apply to fulfill after expiry with some weighted shuffling to avoid contention


broker-template.toml
```diff
@@ -79,7 +79,9 @@ max_concurrent_proofs = 2
 # - "random": Process orders in random order to distribute competition among provers (default)
```
austinabell:
```suggestion
# - "random": Process orders in random order to distribute competition among provers
```

PR #1264, opened 2025-10-24
BM-1781: refactor CLI docs after CLI refactor (say that 10x quick)

sasha-computer:


documentation/site/pages/zkc/mining/enable.mdx
```diff
@@ -68,7 +65,7 @@ x-base-environment: &base-environment
   RUST_LOG: ${RUST_LOG:-info}
   RISC0_HOME: /usr/local/risc0
   RUST_BACKTRACE: 1
-  POVW_LOG_ID: "0x1234...5678" // [!code hl] [!code focus]
```
willpote, line 71: Went with REWARD_ADDRESS

documentation/site/pages/zkc/mining/walkthrough.mdx
```diff
@@ -4,53 +4,65 @@
...
-After running `prepare`, provers are ready to submit onchain with:
+```bash
+boundless rewards setup
+```
+
+which would store the state file location, such that provers can simply run:
 
 ```bash
-boundless povw submit --state ${STATE_FILE_LOCATION} --rpc-url ${RPC_URL} --private-key ${PRIVATE_KEY} --povw-private-key ${POVW_PRIVATE_KEY}
+boundless rewards prepare-mining --from-bento
```
willpote: i removed --from-bento as it made some assumptions on where you are running the command,

`--bento-api-url <URL>`

sasha-computer: that makes sense, cleaner this way, thanks for spotting this

PR #1265, opened 2025-10-24
BM-1782: combine market monitor RPC queries

austinabell:
- Combines on-chain event queries to be more RPC efficient
- Fixes the manual deployment chain ID check for permits
- Other passby localnet fixes/improvements

### DIFF
--- a/.env.localnet-template
+++ b/.env.localnet-template
@@ -1,5 +1,4 @@
 # Default anvil node and order stream endpoint
-export RUST_LOG=info,broker=debug
 export RPC_URL="http://localhost:8545"
 export ORDER_STREAM_URL="http://localhost:8585"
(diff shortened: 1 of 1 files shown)


.env.localnet-template
```diff
@@ -1,6 +1,8 @@
 # Default anvil node and order stream endpoint
+export RUST_LOG=info,broker=debug
```
zeroecco: debug is really really verbose, are we sure we want this?

austinabell: for localnet, seems like a reasonable default? Easy to override, but this is just used for testing so I would prefer debug on by default

austinabell:
```suggestion
```

Changed my mind, don't care to die on this hill, I'll just remove if opinions :D

PR #1284, opened 2025-11-03
BM-1813: Update dependencies

capossele:
Pending #1282 

Note: `cargo_bin` has been deprecated, so this PR suggests switching to `escargot` as well as caching the binary path via `lazy_static`

### DIFF
--- a/crates/boundless-cli/src/commands/rewards/submit_mining.rs
+++ b/crates/boundless-cli/src/commands/rewards/submit_mining.rs
@@ -114,9 +114,9 @@ impl RewardsSubmitMining {
 
         ensure!(
             Address::from(state.log_id) == work_log_signer.address(),
-            "Signer does not match the state log ID: signer: {}, state: {}",
+            "Signer does not match the state log ID: signer: {}, state: {:x}",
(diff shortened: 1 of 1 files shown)


crates/indexer/tests/market/basic.rs
```diff
@@ -33,8 +31,14 @@ use boundless_test_utils::{
     guests::{ASSESSOR_GUEST_ELF, ECHO_ID, ECHO_PATH, SET_BUILDER_ELF},
     market::create_test_ctx,
 };
+use lazy_static::lazy_static;
 use sqlx::Row;
 
+lazy_static! {
+    static ref INDEXER_BIN_PATH: PathBuf =
+        escargot::CargoBuild::new().bin("market-indexer").run().unwrap().path().to_path_buf();
+}
```
austinabell, line 40: Noting that this might be deprecated in the future for `LazyLock` in the std lib (https://doc.rust-lang.org/nightly/std/sync/struct.LazyLock.html) -- but if recommended to do with this specifically perhaps there is a reason

capossele:
superseded by #1308

PR #1285, opened 2025-11-03
Remove redundant rust install from bento_cli dockerfile

mintybasil:
This step is unnecessary as the base image is the rust builder image. Also potentially conflicts with the version set by the `RUST_IMG` variable.


austinabell (approved):

mintybasil:
Amended commit with signature. The pains of remote dev...

PR #1294, opened 2025-11-05
BM-1854: fix(bento): aux worker cleanup/connection handling

austinabell:
Various fixes/changes to bento cleanup:
- https://github.com/boundless-xyz/boundless/commit/1ddbcfb1e74900c83cf1ec3d70292437113f632d: Env to allow turning off either cleanup cron job (if any issues in future or if intending to manually handle db maintenance)
- https://github.com/boundless-xyz/boundless/commit/afcc4286e8c6ebee6409bb72f953797ab615acac: Increase aux max client connections default to handle the 4 concurrent tasks. All other workers are sequentially processing so this should be sufficient for now
- https://github.com/boundless-xyz/boundless/commit/2bd97550ce38378b62cf8d5b8bf67b33508f52af: to avoid running on startup to avoid issues with the spinning on startup where all try to connect at one time

Separated into diff commits if we only want a subset of these changes

### DIFF
--- a/bento/crates/workflow/src/redis.rs
+++ b/bento/crates/workflow/src/redis.rs
@@ -4,7 +4,6 @@
 // as found in the LICENSE-BSL file.
 
 use anyhow::{Context, Result};
-use deadpool_redis::Connection;
 pub use deadpool_redis::{Config, Pool as RedisPool, Runtime, redis::AsyncCommands};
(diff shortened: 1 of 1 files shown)


zeroecco:
lgtm

zeroecco (approved):

PR #1297, opened 2025-11-06
BM-1858: add removed sanity check from refactor

austinabell:
Adding the `check_work_receipt` sanity check, which was removed in the CLI refactor should be sufficient, but I noticed there were some other sanity checks that did not match upstream r0 so I added those to future proof in case those others could be hit in the future in some case.


austinabell:
This has been confirmed to have fixed the issue for the external prover

willpote (approved):

PR #1298, opened 2025-11-06
BM-1859: Clean Bento Inputs 

ec2:
Deletes the inputs which can be quite large (e.g. Signal) when the Order is completed or failed.
It will also delete when it fails preflights in the OrderPicker and when the OrderMonitor decides to skip.

Still testing on personal machine to see how effective this is.

### DIFF
--- a/crates/broker/src/aggregator.rs
+++ b/crates/broker/src/aggregator.rs
@@ -517,7 +517,7 @@ impl AggregatorService {
 
                 if let Err(err) = self
                     .db
-                    .set_order_failure(&order.order_id, "Expired before aggregation", &self.prover)
+                    .set_order_failure(&order.order_id, "Expired before aggregation")
                     .await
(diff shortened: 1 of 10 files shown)


zeroecco:
bento section looks fine to me

austinabell:
realizing there is a flaw with this in that currently we cache inputs to try to deduplicate, so deleting one may interfere with another if in-between preflight and prove. Will explore alternatives to see if a clean way to get around

austinabell:
Experimented with uploading duplicate input for each order variant, can work but is pretty messy and very unideal to put more memory pressure on the REST API and unnecessary IO.

Intuition is that perhaps checking the other variant is skipped/failed/completed before deleting might be a better way around this, with the tradeoff that on the race condition both check at the same time, storage is leaked. Will explore that when I work on this next

PR #1310, opened 2025-11-10
BM-1884: Zeroecco/internal reqs

zeroecco:


infra/prover-cluster/components/ApiGatewayComponent.ts
```diff
@@ -76,12 +77,44 @@ export class ApiGatewayComponent extends BaseComponent {
             enableDeletionProtection: false,
             securityGroups: [config.securityGroupId],
             tags: {
-                Name: `${this.config.stackName}-prover-internal`,
+                Name: `${this.config.stackName}-internal`,
```
ncocchiaro, line 80: Should the names (and related tags) for these be more specific to the service?

infra/prover-cluster/components/ApiGatewayComponent.ts
```diff
@@ -94,14 +127,20 @@ export class ApiGatewayComponent extends BaseComponent {
...
             port: 80,
             protocol: "HTTP",
             defaultActions: [{
                 type: "forward",
-                targetGroupArn: this.targetGroup.arn,
+                targetGroupArn: this.internalTargetGroup.arn,
             }],
+        }, {
+            dependsOn: [this.internalTargetGroup],
+            replaceOnChanges: ["defaultActions"],
```
ncocchiaro: Replacing the listener can cause some amount of downtime, is that acceptable here?

infra/prover-cluster/components/ApiGatewayComponent.ts
zeroecco: I can't extend the name any further 32 character limit :(

infra/prover-cluster/components/ApiGatewayComponent.ts
zeroecco: that should be fine yes. Anything consuming this should be able to retry

infra/prover-cluster/components/ApiGatewayComponent.ts
ncocchiaro: Ah true. Oh well.

infra/prover-cluster/Pulumi.prod-nightly-8453.yaml
```diff
@@ -2,6 +2,8 @@ secretsprovider: awskms:///arn:aws:kms:us-west-2:968153779208:alias/pulumi-secre
 encryptedkey: AQICAHhnm/1/W7/xeF2uxmXOGjFzOf6jjMX+KgbMb0K+LSCbsAGA9yGzYmX18KEh3qL46b5UAAAAfjB8BgkqhkiG9w0BBwagbzBtAgEAMGgGCSqGSIb3DQEHATAeBglghkgBZQMEAS4wEQQM9oZ6wye2fTuMWV0GAgEQgDuEacQklGHw6WJN3N4UIAg9ChU4WeCn02cR3th7uPUpnjOJ6LMK9i1Fl8uP29fyw8O/lxuUeBZZi5E8MQ==
 config:
   base:BASE_STACK: organization/bootstrap/services-prod
+  base:SLACK_ALERTS_TOPIC_ARN: arn:aws:sns:us-west-2:968153779208:boundless-alerts-topic-prod
```
ncocchiaro: I think this needs to be `boundless-alerts-topic-launch`, same for the other occurrences of `-prod`

PR #1315, opened 2025-11-13
BM-1894: Add application verifier

capossele:
In preparation for the bitVM integration, this PR adds a new immutable verifier used for verifying application seals.
It also drops the `withdrawFromTreasury` and `withdrawFromCollateralTreasury` method from the contract due to bytecode size limitations


contracts/src/BoundlessMarket.sol
```diff
@@ -111,23 +115,32 @@ contract BoundlessMarket is
...
         // Validate non-zero critical params
         require(address(verifier) != address(0), "Invalid verifier");
+        require(address(applicationVerifier) != address(0), "Invalid application verifier");
         require(assessorId != bytes32(0), "Invalid assessor image");
         require(collateralTokenContract != address(0), "Invalid collateral token");
         if (deprecatedAssessorDuration > 0) {
             require(deprecatedAssessorId != bytes32(0), "Invalid deprecated assessor image");
         }
 
         VERIFIER = verifier;
```
willpote, line 143: I guess it would make the migration more difficult if we rename to ASSESSOR_VERIFIER? Can we at least update the comments on the field that its only used for the assessor?

capossele: Yeah it might break the upgrade, but added a comment

PR #1319, opened 2025-11-14
BM-1900: add VerifierLayeredRouter

capossele:
Adds the `VerifierLayeredRouter` contract.
It also adds a copy of the `RiscZeroVerifierRouter` as we need to add the `virtual` property to the methods we override. Can be removed once those changes land to `risc0-ethereum`  (https://github.com/risc0/risc0-ethereum/pull/703)

### DIFF
--- a/justfile
+++ b/justfile
@@ -133,7 +133,7 @@ test-db action="setup":
     fi
 
 # Run all formatting and linting checks
-check: check-links check-license check-format check-clippy check-deployments
+check: check-links check-license check-format check-clippy
 
 # Check links in markdown files
 check-links:


contracts/test/VerifierLayeredRouter.t.sol
```diff
@@ -0,0 +1,402 @@
...
+        vm.expectCall(address(verifierMockA), abi.encodeCall(IRiscZeroVerifier.verifyIntegrity, TEST_RECEIPT_A), 1);
+        vm.expectCall(
+            address(verifierMockA), abi.encodeCall(IRiscZeroVerifier.verifyIntegrity, TEST_MANGLED_RECEIPT_A), 1
+        );
+        vm.expectCall(address(verifierMockB), new bytes(0), 2);
+        vm.expectCall(address(verifierMockB), abi.encodeCall(IRiscZeroVerifier.verifyIntegrity, TEST_RECEIPT_B), 1);
+        vm.expectCall(
+            address(verifierMockB), abi.encodeCall(IRiscZeroVerifier.verifyIntegrity, TEST_MANGLED_RECEIPT_B), 1
+        );
+        layeredRouter.verifyIntegrity(TEST_RECEIPT_A);
```
willpote, line 115: I'm finding it hard to follow the `expectCall`'s here and below.

Its hard to map each call to a specific invocation of verifyIntegrity, and it says we expect 2 calls but there are 6 vm.expectCall invocations.

Any way to split the tests or make it clearer somehow?

capossele: So that, specifically `vm.expectCall(address(verifierMockA), new bytes(0), 2);` is a trick to assess how many calls to expect for that address, in this case 2; then, we assert the actual calls. It's kept it as is so to match the `RiscZeroVerifierRouter` on the `risc0-ethereum` repo.

PR #1320, opened 2025-11-14
Drop risc0-ethereum patch

capossele:


crates/boundless-market/src/selector.rs
```diff
@@ -15,16 +15,120 @@
...
+
+#[repr(u32)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+/// Extended selectors for various proof types.
+pub enum SelectorExt {
+    /// A fake receipt selector used for testing and development.
+    FakeReceipt = 0xFFFFFFFF,
+    /// Groth16 proof selector version 3.0.
+    Groth16V3_0 = 0x73c457ba,
```
ec2: Instead of hard coding it, should we reimport from r0-ethereum for the selectors that they manage?

crates/boundless-market/src/selector.rs
```diff
@@ -55,7 +159,7 @@ impl Default for SupportedSelectors {
             .with_selector(UNSPECIFIED_SELECTOR, ProofType::Any)
             .with_selector(FixedBytes::from(Selector::groth16_latest() as u32), ProofType::Groth16)
             .with_selector(
-                FixedBytes::from(Selector::blake3_groth16_latest() as u32),
+                FixedBytes::from(SelectorExt::blake3_groth16_latest() as u32),
```
ec2, line 165: Should the one above use SelectorExt::groth16_latest?

ec2 (approved):
Should we also go through the whole codebase and change every usage of Selector with SelectorExt? Otherwise lgtm

PR #1324, opened 2025-11-14
BM-1908: Add support for FallbackLayer to boundless sdk

willpote:
Support FallbackLayer in boundless SDK and in broker. This allows configuring multiple RPCs and trying all of them.

Enables it for one of the two single instance provers we operate, so we can monitor and compare perf to the other one

### DIFF
--- a/examples/composition/Cargo.lock
+++ b/examples/composition/Cargo.lock
@@ -2129,7 +2129,6 @@ dependencies = [
  "alloy",
  "alloy-chains",
  "alloy-primitives",
- "alloy-rpc-client",
  "alloy-sol-types",
  "anyhow",
  "async-stream",
--- a/examples/counter-with-callback/Cargo.lock
+++ b/examples/counter-with-callback/Cargo.lock
@@ -2129,7 +2129,6 @@ dependencies = [
  "alloy",
  "alloy-chains",
  "alloy-primitives",
- "alloy-rpc-client",
  "alloy-sol-types",
  "anyhow",
  "async-stream",
--- a/examples/counter/Cargo.lock
+++ b/examples/counter/Cargo.lock
@@ -2129,7 +2129,6 @@ dependencies = [
  "alloy",
  "alloy-chains",
  "alloy-primitives",
- "alloy-rpc-client",
  "alloy-sol-types",
  "anyhow",
  "async-stream",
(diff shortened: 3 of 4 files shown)


crates/boundless-market/src/client.rs
```diff
@@ -100,24 +105,64 @@ pub trait ClientProviderBuilder {
...
+    }
+
+    /// Build a custom RPC client transport with fallback support for multiple URLs.
+    fn build_fallback_transport(&self, urls: &[Url]) -> Result<RpcClient, anyhow::Error> {
+        // Create HTTP transports for each URL
+        let transports: Vec<Http<_>> = urls.iter().map(|url| Http::new(url.clone())).collect();
+
+        // Configure FallbackLayer with all transports active
+        let active_count =
+            std::num::NonZeroUsize::new(transports.len()).unwrap_or(std::num::NonZeroUsize::MIN);
```
austinabell: I don't think we should just ignore the case of no transports being selected? This feels like we would be opening up to issues where there are 0 transports, but the fallback layer assumes there is one? Given that we are only calling this if there are values, should this not just pass back an error in this case?

crates/boundless-market/src/client.rs
```diff
@@ -100,24 +105,64 @@ pub trait ClientProviderBuilder {
...
+
+        // Add the primary RPC URL if set
+        if let Some(ref rpc_url) = self.rpc_url {
+            all_urls.push(rpc_url.clone());
+        }
+
+        // Add any additional URLs from rpc_urls
+        all_urls.extend(self.rpc_urls.clone());
+
+        Ok(all_urls)
```
austinabell: Seems like erroring if `all_urls` is empty here might be helpful with a message that indicates you need to set either or both variables?

crates/boundless-market/src/client.rs
```diff
@@ -100,24 +105,66 @@ pub trait ClientProviderBuilder {
...
+    }
+
+    /// Build a custom RPC client transport with fallback support for multiple URLs.
+    fn build_fallback_transport(&self, urls: &[Url]) -> Result<RpcClient, anyhow::Error> {
+        // Create HTTP transports for each URL
+        let transports: Vec<Http<_>> = urls.iter().map(|url| Http::new(url.clone())).collect();
+
+        // Configure FallbackLayer with all transports active
+        let active_count = std::num::NonZeroUsize::new(transports.len())
+            .expect("at least one transport is required");
```
austinabell:
```suggestion
            .context("at least one transport is required")?;
```

Nit: just to future proof avoiding a panic if refactor leads this to be possible

austinabell (approved):

Issue #1304, opened 2025-11-07
Broker attempts to fulfill already fulfilled order after lookback failure


mintybasil:
Upon updating to the latest version (v1.1.1, from v1.0.2) and starting the broker this error occurred:
```
{"timestamp":"2025-11-07T21:27:07.768411Z","level":"WARN","fields":{"message":"[B-SUP-RECOVER] Recoverable error: [B-MM-500] Unexpected error: Failed to get logs: Max retries exceeded server returned an error response: error code -32005: please limit the query to at most 1000 blocks, data: null: Max retries exceeded server returned an error response: error code -32005: please limit the query to at most 1000 blocks, data: null, spawning replacement (retry 1)"},"target":"broker::task"}
```
My rpc provider limits log queries to 1000 blocks. I have 999 blocks configured (`lookback_blocks`), which worked previously but now causes the lookback to fail. (Reducing this value to `500` worked, I haven't tested other values nor looked into why the limit was hit as it may just be an issue with the provider.)

After failing to do the lookback the broker proceeded to fulfill a request:
```
{"timestamp":"2025-11-07T21:31:10.348453Z","level":"INFO","fields":{"message":"Request 0xc197ebe12c7bcf1d9f3b415342bdbc795425335cfce90c21 was locked by another prover but expired unfulfilled, setting status to pending proving"},"target":"broker::order_monitor"}
```
https://explorer.boundless.network/orders/0xc197ebe12c7bcf1d9f3b415342bdbc795425335cfce90c21?from=orders

This request was fulfilled 30minutes prior by another prover.

mintybasil:
My broker is continuing to fulfill already fulfilled orders, despite the lookback error being resolved. 

https://explorer.boundless.network/orders/0xc197ebe12c7bcf1d9f3b415342bdbc795425335c274a7ccf

mintybasil:
I've just tried running the pre-built docker image (v1.1.1) with this config:
```
[market]
min_mcycle_price = "0.00000002"
min_mcycle_price_collateral_token = "0.0005"
peak_prove_khz = 2800
priority_requestor_lists = [
	"https://requestors.boundless.network/boundless-recommended-priority-list.standard.json",
]
max_collateral = "20"
max_concurrent_preflights = 10
max_concurrent_proofs = 1
balance_warn_threshold = "0.025"
balance_error_threshold = "0.01"
collateral_balance_warn_threshold = "10"
collateral_balance_error_threshold = "5"
[...]

PR #1329, opened 2025-11-17
BM-1933: Additional Order Fulfillment Checks

austinabell:
Replaces #1323 (CI)

Draft because I'd also like to add a sanity check poll on proving to ensure if an event is missed that it still doesn't continue to prove

### DIFF
--- a/infra/prover-cluster/components/WorkerClusterComponent.ts
+++ b/infra/prover-cluster/components/WorkerClusterComponent.ts
@@ -71,7 +71,7 @@ export class WorkerClusterComponent extends BaseComponent {
             ...config,
             launchTemplateId: launchTemplate.launchTemplate.id,
             launchTemplateUserData: pulumi.output(launchTemplate.launchTemplate.userData).apply(u => u || ""),
-            minSize: 15,
+            minSize: config.proverCount < 15 ? config.proverCount : 15,
             maxSize: 100,
             desiredCapacity: config.proverCount,
             componentType: "prover",


crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -209,6 +218,18 @@ fn extract_tx_log<E: SolEvent + Debug + Clone>(
     }
 }
 
+fn validate_fulfill_receipt(receipt: TransactionReceipt) -> Result<(), MarketError> {
+    if let Some(log) = receipt.decoded_log::<IBoundlessMarket::PaymentRequirementsFailed>() {
+        let raw_error = Bytes::copy_from_slice(&log.error);
+        match IBoundlessMarketErrors::abi_decode(&raw_error) {
+            Ok(err) => Err(MarketError::PaymentRequirementsFailed(err)),
+            Err(_) => Err(MarketError::PaymentRequirementsFailedUnknownError(raw_error)),
+        }
```
austinabell: Second guessing this change. Is perhaps a bit misleading to throw an error if only one of the fulfillments failed, especially since the tx didn't fail. Perhaps the sanity checks to avoid submitting if the order is fulfilled is sufficient? @mintybasil was there a strong reason why you wanted this, or just for a better error message? Wondering if okay to just warn log in these cases?

mintybasil: Definitely agree this could be handled more gracefully if there are multiple fulfillments. I am a bit biased since I rarely submit batches > 1. 

Im on mobile right now, but IIRC this was also to address the successful submission log that specifies the reward amounts, since that is no longer accurate.

Warn level log seems very reasonable, that makes the issue visible to operators which is all that really matters.

mintybasil:
~@austinabell Running my commits on top of v1.1.2 it appears that the broker still fulfills already fulfilled orders when catching up after a restart, perhaps an edge case I missed.~ May have jumped the gun. Need to investigate further.

Also - this can close #1304.

PR #1337, opened 2025-11-18
BM-1949: Upgrade Rust toolchain to 1.91.1

ec2:
Rust 1.90 introduces the new faster linker, so updating from 1.89 to 1.91.1, the latest rust toolchain.


ec2:
Weird... my git autosigning stopped working at some point and didnt notice...

PR #1338, opened 2025-11-18
BM-1950: Update log lambda to query alarm tags for debugging via chatbot

ncocchiaro:
This PR adds an additional approach for the log lambda to find the information it needs either to output debug data via chatbot or to create a log query. If an incoming alarm has the relevant information in tags then the lambda won't need to rely on parsing the alarm names, offering an option for debug configuration that can be more robust and easier to expand on in the future.

In order to achieve this, the lambda needs to assume a role in the target account that allows listing tags on the alarm resources. This is set up here via new roles in `bootstrap` and a change to the lambda execution role, plus new lambda code.

The original name parsing logic remains and is still used in all cases in which alarms are not tagged (currently all cases except for prover-cluster alarms, which add tags in this PR as well).

### DIFF
--- a/infra/prover-cluster/components/MetricAlarmComponent.ts
+++ b/infra/prover-cluster/components/MetricAlarmComponent.ts
@@ -4,11 +4,12 @@ import { ChainId, Severity } from "../../util";
 import { BaseComponent, BaseComponentConfig } from "./BaseComponent";
 
 export interface MetricAlarmConfig extends BaseComponentConfig {
-    serviceName: string,
-    logGroupName: string,
-    alertsTopicArns: string[],
-    alarmDimensions: { [key: string]: pulumi.Input<string> },
+    serviceName: string;
+    logGroupName: string;
+    alertsTopicArns: string[];
+    alarmDimensions: { [key: string]: pulumi.Input<string> };
     chainId: string,
+    minAsgSize: pulumi.Output<number>;
 }
 
 // Creates and manages general metric filters and alarms that can be common to multiple components
@@ -211,17 +212,21 @@ export class WorkerClusterAlarmComponent extends MetricAlarmComponent {
     }
 
     private createAutoScalingGroupAlarms = (config: MetricAlarmConfig): void => {
-        this.createMetricAlarm(config, 'asg-in-service-instances', Severity.SEV2, {
-            metricName: 'GroupInServiceInstances',
-            namespace: `AWS/AutoScaling`,
-            period: 60,
-            dimensions: config.alarmDimensions,
-            evaluationPeriods: 20,
-            datapointsToAlarm: 20,
-            statistic: 'Maximum',
-            threshold: 0,
-            comparisonOperator: 'LessThanOrEqualToThreshold'
-        }, 'Number of in service instances is 0 for 20 consecutive minutes.')
+        config.minAsgSize.apply((size => {
+            if (size > 0) {
+                this.createMetricAlarm(config, 'asg-in-service-instances', Severity.SEV2, {
+                    metricName: 'GroupInServiceInstances',
+                    namespace: `AWS/AutoScaling`,
+                    period: 60,
+                    dimensions: config.alarmDimensions,
+                    evaluationPeriods: 20,
+                    datapointsToAlarm: 20,
+                    statistic: 'Maximum',
+                    threshold: 0,
+                    comparisonOperator: 'LessThanOrEqualToThreshold'
+                }, 'Number of in service instances is 0 for 20 consecutive minutes.')
+            }
+        }));
 
         // Errors from Bento logs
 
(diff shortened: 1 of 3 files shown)


infra/pipelines/log-lambda/src/logGroups.ts
```diff
@@ -5,7 +5,7 @@ const logGroupMapping = (service: string): string => {
     case 'bonsai-prover':
       return 'l-prod-11155111-bonsai-prover-11155111';
     case 'monitor':
-      return 'l-prod-11155111-monitor-11155111-monitor';
+      return 'l-prod-11155111-monitor-11155111-mon-monitor';
```
ncocchiaro, line 8: This was renamed in the wild, so needs to be renamed here

infra/pipelines/log-lambda/src/logQueries.ts
```diff
@@ -1,9 +1,7 @@
-export const SERVICE_TO_QUERY_STRING_MAPPING = (service: string, logGroup: string, metricAlarmName: string) => {
-  switch (service) {
-    case 'bento-prover-1':
-      // Note we have to escape two backslashes in the query string. In
-      // CW it should look like: regexp_replace(log, '\\x1b\\[[0-9;]*[mK]', '') AS msg
-      return `
+function proverQuery(logGroup: string): string {
```
ncocchiaro, line 1: Same query as before, just refactored into a function

infra/prover-cluster/components/MetricAlarmComponent.ts
```diff
@@ -321,20 +326,16 @@ export class ProverMetricAlarmComponent extends WorkerClusterAlarmComponent {
...
 
-export interface ManagerMetricAlarmConfig extends MetricAlarmConfig {
-    chainId: string
-}
-
 // Creates and manages metric filters and alarms that are specific to the manager component
 export class ManagerMetricAlarmComponent extends MetricAlarmComponent {
 
-    constructor(config: ManagerMetricAlarmConfig) {
+    constructor(config: MetricAlarmConfig) {
```
ncocchiaro, line 337: This no longer needs a specific subclass since `chainId` is moving to the base `MetricAlarmConfig` interface

infra/prover-cluster/components/MetricAlarmComponent.ts
```diff
@@ -111,6 +112,10 @@ export class MetricAlarmComponent extends BaseComponent {
                 Name: `${config.stackName}-${config.serviceName}-${metricName}-${severity}`,
                 Environment: this.config.environment,
                 Project: "boundless-bento-cluster",
+                ServiceName: config.serviceName,
+                LogGroupName: config.logGroupName,
+                StackName: config.stackName,
+                ChainId: config.chainId,
```
ncocchiaro, line 119: These new tags on the alarms allow the new lambda logic to work

ncocchiaro (commented):
Some comments to help reviewers

infra/pipelines/components/notifications.ts
```diff
@@ -383,6 +383,23 @@ export class Notifications extends pulumi.ComponentResource {
...
+          name: 'cloudwatchAlarmListTagsRolePolicy',
+          policy: pulumi.jsonStringify({
+            Version: "2012-10-17",
+            Statement: [{
+              Sid: "AllowAlarmListTagsAssumeRole",
+              Effect: 'Allow',
+              Action: 'sts:AssumeRole',
+              Resource: [
+                'arn:aws:iam::245178712747:role/alarmListTagsRole',
+                'arn:aws:iam::632745187633:role/alarmListTagsRole'
```
willpote: Nit: import the account numbers from `accountConstants.ts`

infra/bootstrap/index.ts
```diff
@@ -54,6 +55,38 @@ export = async () => {
     }],
   });
 
+  // Create a role that the log fetcher Lambda can assume to list tags on Cloudwatch alarms in the current account
+  const alarmTagsRole = new aws.iam.Role(`${prefix}alarmListTagsRole`, {
+    name: `${prefix}alarmListTagsRole`,
```
willpote, line 60: Why don't we just grant the lambda execution role the permissions to read tags, rather than creating a new role and assuming it in the lambda?

willpote (approved):

infra/bootstrap/index.ts
ncocchiaro: Does that work cross-account though? I had the same thought initially but from experience didn't think it would, I double checked in testing anyway but that confirmed that I couldn't get it to access the resources without assuming a role in the other accounts. Maybe you have a detailed suggestion?

infra/pipelines/components/notifications.ts
ncocchiaro: Good point, I missed that. Fixed in https://github.com/boundless-xyz/boundless/pull/1338/commits/3eb768f3c924b8898cd8278cb56f76a39d9e8be6

infra/bootstrap/index.ts
willpote: Ah forgot this was cross account, and I would assume tags doesn't support that. nevermind!

PR #1339, opened 2025-11-18
BM-1951: refactor DynamicGasFiller and update defaults

austinabell:
After a few iterations, decided to try to keep it as simple as possible, but has tradeoffs like not being able to configure priority fee separately from base fee. Deprecated the ability to set an additional gas to the lock transaction because it previously conflicted with the dynamic gas filler and static amounts don't make too much sense.

- Refactors dynamic gas filler to simplify into three main modes: low, medium, high where medium is 10% gas estimates scaled up, to give more breathing room by default
- Swap `lockin_priority_gas` for `gas_priority_mode` config
- Remove static gas optional from lock/fulfill (fulfill wasn't being used)

### DIFF
--- a/crates/broker/src/bin/broker.rs
+++ b/crates/broker/src/bin/broker.rs
@@ -118,19 +118,6 @@ async fn main() -> Result<()> {
         tracing::info!("Configuring broker with single RPC URL: {}", args.rpc_url);
         RpcClient::builder().layer(retry_layer).http(args.rpc_url.clone())
     };
-    let balance_alerts_layer = BalanceAlertLayer::new(BalanceAlertConfig {
-        watch_address: wallet.default_signer().address(),
-        warn_threshold: config
-            .market
-            .balance_warn_threshold
-            .map(|s| parse_ether(&s))
-            .transpose()?,
-        error_threshold: config
-            .market
-            .balance_error_threshold
-            .map(|s| parse_ether(&s))
-            .transpose()?,
-    });
 
     // Read config for balance alerts (scope the guard so we can move config_watcher later)
     let balance_alerts_config = {


crates/boundless-market/src/dynamic_gas_filler.rs
```diff
@@ -82,20 +157,30 @@ impl<N: Network> TxFiller<N> for DynamicGasFiller {
     where
         P: Provider<N>,
     {
+        const _: () = {
+            // Assumptions about the default GasFiller params.
+            // Using default, as it is requires a lot of duplication to override.
+            assert!(alloy::providers::utils::EIP1559_FEE_ESTIMATION_REWARD_PERCENTILE == 20.0);
+            assert!(alloy::providers::utils::EIP1559_BASE_FEE_MULTIPLIER == 2);
```
willpote: This is very unfortunate that Alloy doesn't enable customizing these easily (and `EIP1559_FEE_ESTIMATION_PAST_BLOCKS`), as this would give us the clearest way to create these priority modes.

As written we are applying a % multiplier increase to the median of the 20th percentile of priority fees from the past 10 blocks, which is extremely kind of hard to reason about... 

I think the ideal design would be for the priority modes to specify the percentile (and possibly num blocks to consider) when generating the values.

Agree too much duplication and code needed for a quick patch fix, but for 1.2 something to consider

crates/boundless-market/src/dynamic_gas_filler.rs
```diff
@@ -21,48 +23,121 @@ use alloy::{
...
 };
+use serde::{Deserialize, Serialize};
+use tokio::sync::RwLock;
+
+/// Priority mode for transaction gas pricing.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "snake_case")]
+#[non_exhaustive]
+pub enum PriorityMode {
+    /// Low gas price/priority fee.
```
willpote: I think its worth explaining what the logic does today a bit more, so that provers understand what customizing it does, focusing in on priority_gas_fee, as given we set max_fee_per_gas with a 2x multiplier on base fee, i don't see how it can ever get hit.

Something like, "By default transactions are sent using the median of the 20th percentile priority_gas_fee from the past 10 blocks. The various priority modes then apply a multiplier on top of this value to compute the final priority_gas_fee for sending the transaction."

"gas price" is basically deprecated at this point, so I think it makes sense to just talk about the impact on max_fee_per_gas and priority_fee_per_gas these modes have.

crates/boundless-market/src/dynamic_gas_filler.rs
```diff
@@ -21,48 +23,121 @@ use alloy::{
...
+                dynamic_multiplier_percentage: 5,
+            },
+        }
+    }
+}
+
+/// Configuration for a priority mode.
+#[derive(Clone, Debug)]
+struct PriorityModeConfig {
+    /// The base percentage increase applied to every transaction (e.g., 0 = no change, 10 = 10% increase).
```
willpote: Specially, this is applied to `max_priority_fee_per_gas` and `max_fee_per_gas`

austinabell: https://github.com/boundless-xyz/boundless/pull/1339/commits/de01e2186cf9e5acb38f41d2812ba9d4a6a27150

crates/boundless-market/src/dynamic_gas_filler.rs
austinabell: https://github.com/boundless-xyz/boundless/pull/1339/commits/de01e2186cf9e5acb38f41d2812ba9d4a6a27150

crates/boundless-market/src/dynamic_gas_filler.rs
```diff
@@ -12,57 +12,228 @@
...
+    (utils::EIP1559_BASE_FEE_MULTIPLIER as u64) * 100;
+
+/// Priority mode for transaction gas pricing.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "snake_case")]
+#[non_exhaustive]
+pub enum PriorityMode {
+    /// Use the eth_feeHistory as is and only add `+3%` per pending tx.
+    Low,
+    /// Add a `+20%` increase to the base/priority fee and `+5%` per pending transaction.
```
willpote: Comment looks incorrect here and below, think you mean use 20th percentile etc

crates/boundless-market/src/dynamic_gas_filler.rs
```diff
@@ -12,57 +12,228 @@
...
-    transports::TransportResult,
+    transports::{RpcError, TransportResult},
 };
+use serde::{Deserialize, Serialize};
+use tokio::sync::RwLock;
+
+const DEFAULT_FEE_HISTORY_PERCENTILES: [f64; 1] = [utils::EIP1559_FEE_ESTIMATION_REWARD_PERCENTILE];
+const LOW_PRIORITY_PERCENTILES: [f64; 1] = [10.0];
+const MEDIUM_PRIORITY_PERCENTILES: [f64; 1] = [20.0];
+const HIGH_PRIORITY_PERCENTILES: [f64; 1] = [30.0];
```
willpote: I think we should go with: 10 30 50

Compared to Metamask, the impact of a tx not confirming is far larger, as it causes slashes and holds up other transactions the broker needs to send. And in Metamask the user is able to manually retry, but the broker runs indefinitely.

crates/boundless-market/src/dynamic_gas_filler.rs
```diff
@@ -12,57 +12,228 @@
...
+        match self {
+            PriorityMode::Low => PriorityModeConfig {
+                base_fee_multiplier_percentage: DEFAULT_BASE_FEE_MULTIPLIER_PERCENTAGE,
+                priority_fee_multiplier_percentage: 100,
+                reward_percentiles: LOW_PRIORITY_PERCENTILES.to_vec(),
+                dynamic_multiplier_percentage: 3,
+            },
+            PriorityMode::Medium => PriorityModeConfig {
+                base_fee_multiplier_percentage: DEFAULT_BASE_FEE_MULTIPLIER_PERCENTAGE,
+                priority_fee_multiplier_percentage: 105,
```
willpote: I don't think we need to do a multiplier on priority fee given the new design. The way its different between Medium and High makes it difficult to reason about also. Fine to leave it for the custom one, but I think by default  we should just set to 100 for Low/medium/high

crates/boundless-market/src/dynamic_gas_filler.rs
```diff
@@ -12,57 +12,228 @@
...
+        base_fee_multiplier_percentage: u64,
+        /// Percentage multiplier applied to the computed priority fee (100 = unchanged, 120 = +20%).
+        #[serde(
+            default = "default_custom_priority_fee_multiplier_percentage",
+            alias = "multiplier_percentage"
+        )]
+        priority_fee_multiplier_percentage: u64,
+        /// Which percentiles to request via `eth_feeHistory`.
+        #[serde(default = "default_custom_reward_percentiles")]
+        reward_percentiles: Vec<f64>,
```
willpote: I don't think this should be a Vec- i dont see why you'd want to request multiple reward percentiles per block then take the median

Also how about `priority_fee_percentile`

crates/boundless-market/src/dynamic_gas_filler.rs
austinabell: Honestly meant it to be 20, 30, 50 and forgot to update. Good catch! (If you think low should stay at 10 I can make that change, but figured in that edge case someone can just set a custom value.

crates/boundless-market/src/dynamic_gas_filler.rs
austinabell: I was defaulting to allowing it because the upstream alloy APIs have it that way (as well as obviously the underlying eth call), so didn't want to assume it wouldn't be used as such (https://github.com/boundless-xyz/boundless/pull/1339/commits/d2d888e8457fa453e9ab94b1abbb72b59b00f6ae)

crates/boundless-market/src/dynamic_gas_filler.rs
austinabell: fair enough, was just giving a bit of buffer to avoid issues. Set all to 100 https://github.com/boundless-xyz/boundless/pull/1339/commits/d2d888e8457fa453e9ab94b1abbb72b59b00f6ae

crates/boundless-market/src/dynamic_gas_filler.rs
austinabell: yes, was from previous version. https://github.com/boundless-xyz/boundless/pull/1339/commits/d2d888e8457fa453e9ab94b1abbb72b59b00f6ae fixed, thanks

PR #1342, opened 2025-11-20
BM-1961: Default to 3s poll to save RPC costs

willpote:
Currently we poll every second. This is unnecessary as instant-locking doesn't occur at the moment, and leads to higher RPC costs from `eth_getLogs` calls. Increasing to 3s by default.

### DIFF
--- a/infra/prover-cluster/components/AutoScalingGroupComponent.ts
+++ b/infra/prover-cluster/components/AutoScalingGroupComponent.ts
@@ -9,7 +9,7 @@ export interface AutoScalingGroupConfig extends BaseComponentConfig {
     minSize: number;
     maxSize: number;
     desiredCapacity: number;
-    componentType: "prover" | "execution" | "aux";
+    componentType: "manager" | "prover" | "execution" | "aux";
 }
 
 export class AutoScalingGroupComponent extends BaseComponent {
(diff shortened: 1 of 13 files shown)


crates/broker/src/market_monitor.rs
```diff
@@ -295,7 +299,7 @@ where
 
                 while from_block <= to_block {
                     let chunk_end = std::cmp::min(
-                        from_block.saturating_add(lookback_blocks - 1),
+                        from_block.saturating_add(events_poll_blocks.saturating_sub(1)),
```
austinabell, line 302: lookback_blocks was what this new `events_poll_blocks` was. If you want to rename it, why not just alias under this new name? Did I misread that the `lookback_blocks` is intended to be used for something else?

willpote: `lookback_blocks` was the number of blocks the broker looks back on startup to find open orders. It just got re-used for this purpose at some point. I think it makes sense to separate them

austinabell: Fair enough, I suppose there could be a reason for setting a smaller window to request each call -- just noting it was a bit confusing what each was representing

documentation/site/pages/provers/broker.mdx
```diff
@@ -66,6 +66,8 @@ Below are all `broker.toml` settings organized by section:
 | max\_journal\_bytes | Max journal size in bytes. Orders that produce a journal larger than this size in preflight will be skipped. Since journals must be posted onchain to complete an order, an excessively large journal may prevent completion of a request. |
 | min\_deadline | Min seconds left before the deadline to consider bidding on a request. If there is not enough time left before the deadline, the prover may not be able to complete proving of the request and finalize the batch for publishing before expiration. |
 | lookback\_blocks | On startup, the number of blocks to look back for possible open orders. |
+| events\_poll\_blocks | Number of blocks to query per batch when polling for new market events. After the initial lookback completes, the market monitor will query this many blocks at a time when polling for new events. A higher value reduces RPC calls but may increase response latency. A lower value provides more real-time updates but increases RPC load. Default: 50. |
```
austinabell: This comment seems misleading, as the way the logic is written now, it does not wait until `event_poll_blocks` behind before making a call, this just sets the max amount of blocks in one query. Also, the default is 150 in the PR, so not sure which of the two is intended

PR #1343, opened 2025-11-20
BM-1971: autoscaler for proves

zeroecco:
elasticache is too expensive
auto scaling for bento


zeroecco:
tested on my dev stack

ec2 (approved):

PR #1371, opened 2025-11-26
BM-2003: remove all documentation files + related CI

sasha-computer:
the docs are now served from [boundless-xyz/docs](https://github.com/boundless-xyz/docs)


sasha-computer:
i don't expect this to work first time, there was a lot of doc CI everywhere

PR #1375, opened 2025-11-26
BM-2008: prune claim journal before finalize proof

austinabell:


willpote (approved):
@nategraf mind giving this a quick look?

austinabell:
> @nategraf mind giving this a quick look?

he already has, so just going to merge. Can change if anyone has opinions after

PR #1377, opened 2025-12-01
BM-2010: Add random requestor that runs daily

willpote:
Random requestor is intended to confirm the permissionless-ness nature of the network by sending requests from an address that is not in any requestor lists periodically.

This is an MVP that uses the same address daily, and just sends the looper program. Next iteration will generate a new address for each run, and prove a block to make the workload more accurate

### DIFF
--- a/infra/order-generator/components/order-generator.ts
+++ b/infra/order-generator/components/order-generator.ts
@@ -284,7 +284,7 @@ export class OrderGenerator extends pulumi.ComponentResource {
 
       logDependency = fargateTask;
     } else {
-      // Service mode: Create FargateService (existing behavior)
+      // Service mode, runs indefinitely
       const service = new awsx.ecs.FargateService(
         `${serviceName}-service`,
         {


infra/distributor/Pulumi.l-prod-8453.yaml
```diff
@@ -30,4 +30,4 @@ config:
   distributor:ETH_RPC_URL:
     secure: v1:E1WuDwMATaOQ6GEE:y8KpEhqaQwjYWkPUmoYNGjuJZTMXmoXcnGyZcECvwkoHjbtA5og5ShcIRcEw2InPOagemm79RIvUAKIWlRthzqBlPXY6gL9eRxwQ
   distributor:DISTRIBUTOR_ETH_ALERT_THRESHOLD: "0.5"
-  distributor:DISTRIBUTOR_STAKE_ALERT_THRESHOLD: "200.0"
\ No newline at end of file
+  distributor:DISTRIBUTOR_STAKE_ALERT_THRESHOLD: "200.0"
```
capossele, line 33: not urgent but eventually will need to change this to use COLLATERAL instead of STAKE for consistency

PR #1386, opened 2025-12-04
BM-2031: Change skip reasons to INFO level

willpote:
You should be able to search an order ID and see the full request lifecycle on INFO mode

Currently we are missing this info so it is hard to debug why our provers are skipping requests

Also adding info to the max concurrent / peak mhz logic to make it clearer what orders are blocking other orders

### DIFF
--- a/crates/broker/src/order_monitor.rs
+++ b/crates/broker/src/order_monitor.rs
@@ -620,8 +620,6 @@ where
         prev_orders_by_status: &mut String,
     ) -> Result<Vec<Arc<OrderRequest>>> {
         let num_orders = orders.len();
-        let order_ids_for_logging =
-            orders.iter().map(|order| order.id()).collect::<Vec<_>>().join(",");
         // Get our current capacity for proving orders given our config and the number of orders that are currently committed to be proven + fulfilled.
         let capacity = self
             .get_proving_order_capacity(config.max_concurrent_proofs, prev_orders_by_status)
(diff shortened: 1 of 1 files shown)


crates/broker/src/order_monitor.rs
```diff
@@ -417,10 +417,10 @@ where
         ) -> bool {
             let expiration = order.expiry();
             if expiration < current_block_timestamp {
-                tracing::debug!("Request {:x} has now expired. Skipping.", order.request.id);
+                tracing::info!("Request {:x} has now expired. Skipping.", order.request.id);
                 false
             } else if expiration.saturating_sub(now_timestamp()) < min_deadline {
-                tracing::debug!("Request {:x} deadline at {} is less than the minimum deadline {} seconds required to prove an order. Skipping.", order.request.id, expiration, min_deadline);
+                tracing::info!("Request {:x} deadline at {} is less than the minimum deadline {} seconds required to prove an order. Skipping.", order.request.id, expiration, min_deadline);
```
austinabell, line 423: nit: this should probably be printing the order ID rather than the request ID to not confuse the lock expiry event from the order expiry. Can fix in another PR though

crates/broker/src/order_monitor.rs
```diff
@@ -803,9 +807,11 @@ where
         }
 
         tracing::info!(
-            "Started with {} orders ready to be locked and/or proven. Already commited to {} orders. After applying capacity limits of {} max concurrent proofs and {} peak khz, filtered to {} orders: {:?}",
+            "Started with {} orders ready to be locked and/or proven ({}). Already commited to {} orders ({}). After applying capacity limits of {} max concurrent proofs and {} peak khz, filtered to {} orders: {:?}",
```
austinabell: This will be an extremely spammy and large log to print out all possible candidate orders on info.

Preference if listing these to put on a lower log level

austinabell (approved):

PR #1387, opened 2025-12-04
BM-2032: move to valkey

zeroecco:
🍱 added to vscode for linting etc
🎉 valkey

### DIFF
--- a/.env.broker-template
+++ b/.env.broker-template
@@ -2,8 +2,8 @@
 
 RUST_LOG=info
 
-REDIS_HOST=redis
-REDIS_IMG=redis:7.2.5-alpine3.19
+VALKEY_HOST=valkey
+VALKEY_IMG=valkey/valkey:9.0.0
 
 POSTGRES_HOST=postgres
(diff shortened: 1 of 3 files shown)


austinabell:
why duplicate of https://github.com/boundless-xyz/boundless/pull/1383 ? Also, so many unnecessary breaking changes with this, why? At least would have to make the env vars backwards compat. This also breaks if you try to upgrade with an in-progress proof in bento, when it shouldn't be necessary

zeroecco:
> why duplicate of #1383 ? Also, so many unnecessary breaking changes with this, why? At least would have to make the env vars backwards compat. This also breaks if you try to upgrade with an in-progress proof in bento, when it shouldn't be necessary

On your first point, I didn't see that PR, my bad! will close this.
whether it is now or later the naming should change to match the service. Might as well call it "cache" or "queue" or something semi app agnostic instead of "redis" or "valkey". 

In the next release I want to push the default PostgreSQL version as well to major version 18 which has async IO improvements.

In progress proof failed for me on 9.0.0 and latest when redis was 7.2.x

PR #1399, opened 2025-12-05
BM-2053: Increase MAX_BLOCK_RANGE from 500 to 10000

zeroecco:


willpote:
> seems like a large range for each request, and untested, but lgtm
> 
> Also, should this be applied to testnets also? Same with #1398

10k is what we use for our other indexing services, I _think_ should be good, but we can revert later if need be

PR #1401, opened 2025-12-08
Backport steel fix to release branch

austinabell:
Just this commit to fix new boundless-cli installs


capossele:
https://github.com/boundless-xyz/boundless/pull/1394 needs also to be back ported to fix the CI

PR #1409, opened 2025-12-10
BM-2054: Update order-generator-random-requestor param

capossele:


infra/order-generator/Pulumi.l-prod-8453.yaml
```diff
@@ -58,6 +58,8 @@ config:
   order-generator-random-requestor:COUNT: "10"
   order-generator-random-requestor:LOCK_TIMEOUT: "1800"
   order-generator-random-requestor:TIMEOUT: "1800"
+  order-generator-random-requestor:RAMP_UP_SECONDS_PER_MCYCLE: "2"
+  order-generator-random-requestor:SECONDS_PER_MCYCLE: "7"
```
capossele, line 62: Alternatively we could be more conservative and use something like 5 and 20

PR #1410, opened 2025-12-10
BM-2096: Cluster log level adjust, fix staging order stream url

willpote:
* Adjusting log level to avoid debug logs from sqlx etc.
* Fixing order stream url to use staging

### DIFF
--- a/infra/prover-cluster/Pulumi.yaml
+++ b/infra/prover-cluster/Pulumi.yaml
@@ -8,3 +8,5 @@ config:
   pulumi:tags:
     value:
       pulumi:template: aws-typescript
+backend:
+  url: s3://boundless-pulumi-state?region=us-west-2&awssdk=v2


austinabell (approved):

austinabell:
> Fixing order stream url to use staging

Was this not committed yet?

infra/prover-cluster/index.ts
```diff
@@ -27,7 +27,7 @@ const environment: string = config.get("environment") || "custom";
 const privateKey: pulumi.Output<string> = config.requireSecret("privateKey");
 const ethRpcUrl: pulumi.Output<string> = config.requireSecret("ethRpcUrl");
 const managerInstanceType: string = config.require("managerInstanceType");
-const orderStreamUrl: string = config.require("orderStreamUrl");
+const orderStreamUrl: pulumi.Output<string> = config.requireSecret("orderStreamUrl"); // Use secret to avoid exposing staging urls
```
austinabell, line 30: I'm maybe just overlooking something, but won't this break release as those are in plaintext? Or does it allow either?

willpote: It works for either, we use this trick for the single instance provers already

PR #1414, opened 2025-12-10
BM-2042: Add FundingMode

capossele:
- `FundingMode::Always` that always send `max_price` as value [**Default**].
- `FundingMode::Never` can be used to never send value with the request.
- `FundingMode::AvailableBalance` can be used to only send value if the current balance is insufficient.
- `FundingMode::BelowThreshold` can be used to send value only if the balance is below a configurable threshold.
- `FundingMode::MinMaxBalance` can be used to maintain a minimum balance by funding requests accordingly.

### DIFF
--- a/crates/boundless-market/src/client.rs
+++ b/crates/boundless-market/src/client.rs
@@ -1045,14 +1045,10 @@ where
         if client_address != signer.address() {
             return Err(MarketError::AddressMismatch(client_address, signer.address()))?;
         };
-        // Ensure address' balance is sufficient to cover the request
-        let balance = self.boundless_market.balance_of(client_address).await?;
-        if balance < U256::from(request.offer.maxPrice) {
-            return Err(ClientError::Error(anyhow!(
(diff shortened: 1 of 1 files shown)


crates/boundless-market/src/client.rs
```diff
@@ -781,6 +849,70 @@ where
...
+                        format_ether(balance),
+                        format_ether(min_balance),
+                        format_ether(to_send),
+                        format_ether(max_balance),
+                    );
+                    to_send
+                } else {
+                    U256::ZERO
+                }
+            }
```
willpote, line 913: What would happen if my request has max price > max balance in this mode?

willpote (approved):

crates/boundless-market/src/client.rs
capossele: Yeah bad things would happen. Maybe I'll check for that and make sure to use the max

PR #1423, opened 2025-12-11
BM-2115: Remove Rsyslog from packer build

zeroecco:
This is to allow journald to handle logs.
also cleanup old builds on nightly packer builds.

### DIFF
--- a/dockerfiles/distributor.dockerfile
+++ b/dockerfiles/distributor.dockerfile
@@ -1,5 +1,5 @@
 # Build stage
-FROM rust:1.88.0-bookworm AS init
+FROM rust:1.89.0-bookworm AS init
 
 RUN apt-get -qq update && \
     apt-get install -y -q clang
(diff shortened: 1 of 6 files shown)


bento/crates/workflow/src/bin/agent.rs
```diff
@@ -14,21 +14,87 @@ async fn main() -> Result<()> {
...
+                    }
+                };
+
+                tracing::info!("Agent {} started", agent_id);
+                agent
+                    .poll_work()
+                    .await
+                    .context(format!("[BENTO-AGENT-003] Agent {} exiting", agent_id))
+            });
+            handles.push(handle);
```
austinabell, line 69: This seems very problematic because currently there is a non-trivial amount of sync code that is not put on threads where blocking is acceptable and is blocking the executor. This isn't a big deal now, but multiplying this by the number of agents seems like it would cause issues.

I suppose the assumption here is that this agents num would only be used for agent variants where this isn't the case? Executor agent for example doesn't look that bad for example as the core execution is put on a blocking thread. This does seem like a footgun though where if you apply the wrong agent type, you will run into issues

bento/crates/workflow/src/bin/agent.rs
```diff
@@ -14,21 +14,87 @@ async fn main() -> Result<()> {
...
+                Ok(Err(e)) => {
+                    tracing::error!("Agent {} exited with error: {:#}", idx + 1, e);
+                    has_error = true;
+                }
+                Err(e) => {
+                    tracing::error!("Agent {} task panicked: {:#}", idx + 1, e);
+                    has_error = true;
+                }
+            }
+        }
```
austinabell, line 88: This seems unideal, because it means that if an agent fails, it will not be restarted until all others fail, where then they will be all restarted (assuming the execution has restarts). Seems like you would want to have each individually restart?

austinabell (commented):
only looked at the Rust code and commented on that (unsure if that was meant to be committed, as it seems unrelated to the PR)

bento/crates/workflow/src/bin/agent.rs
zeroecco: thi is a bad push. will fix

PR #1432, opened 2025-12-12
BM-2126: Ansible Support

zeroecco:
adds ansible for cluster deployment. This is a jump off point and will be improved with additional pr's

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -212,7 +212,8 @@ services:
   aux_agent:
     <<: *agent-common
     mem_limit: 256M
-    replicas: ${BENTO_AUX_WORKER_COUNT:-2}
+    deploy:
+      replicas: ${BENTO_AUX_WORKER_COUNT:-2}
     cpus: 1
     environment:
       <<: *base-environment


ansible/roles/broker/templates/broker.toml.j2
```diff
@@ -0,0 +1,104 @@
...
+# broker being slashed.
+#
+# We strongly recommend using the Boundless CLI's `boundless prover
+# generate-config` command to generate a broker.toml file that is tailored to
+# your specific setup.
+#
+# Additionally, refer to documentation:
+# https://docs.boundless.network/provers/broker#settings-in-brokertoml
+### [End] ###
+
```
willpote: Re toml, I think would be a lot easier if we could just have a tomls folder with "real" tomls, e.g. what we have for the existing single instance provers

https://github.com/boundless-xyz/boundless/tree/main/infra/prover/tomls/bento

ansible/roles/awscli/tasks/main.yml
```diff
@@ -0,0 +1,147 @@
...
+    state: present
+    update_cache: true
+  when:
+    - awscli_install_method == "installer"
+    - awscli_which_check.rc != 0 or awscli_update | bool
+  tags:
+    - awscli
+    - awscli-install
+
+- name: Download AWS CLI installer
```
ncocchiaro, line 45: For this external installer it would be good to add a step to verify the package signature (step 2 in [guided installation steps](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html)).

ansible/roles/broker/templates/broker.toml.j2
ncocchiaro: +1 plus it may be easier to automatically validate pre-generated tomls and protect against type errors.

.github/workflows/prover-release.yml
```diff
@@ -0,0 +1,143 @@
+name: Prover Deployment
+
+on:
+  # Nightly deployment at 1am UTC
+  schedule:
+    - cron: '0 1 * * *'
```
willpote: Seems like as written we will redeploy the release cluster also every night? Not a huge deal, but shutting it down and rebooting it will create a bit of downtime, so if we can avoid would be good

ansible/roles/prover/configs/broker/8453.toml
```diff
@@ -0,0 +1,29 @@
+[market]
```
willpote, line 1: We should have two tomls right? One for release (which sets allow client) and one for nightly (which doesn't and uses more of a vanilla setup). 

We should also have separate tomls for staging/prod, as well as each chain, as we need to configure pricing on staging differently

ansible/roles/bento/templates/docker-compose.env.j2
```diff
@@ -0,0 +1,42 @@
...
+
+# Broker Configuration (if enabled)
+{% if bento_prover_private_key is defined and bento_prover_private_key | string | length > 0 %}
+PROVER_PRIVATE_KEY={{ bento_prover_private_key }}
+{% endif %}
+{% if bento_povw_log_id is defined and bento_povw_log_id | string | length > 0 %}
+POVW_LOG_ID={{ bento_povw_log_id }}
+{% endif %}
+{% if bento_prover_rpc_url is defined and bento_prover_rpc_url | string | length > 0 %}
+PROVER_RPC_URL={{ bento_prover_rpc_url }}
```
willpote: Didn't we migrate to using multiple RPC urls for the release cluster?

ansible/roles/bento/README.md
```diff
@@ -0,0 +1,125 @@
...
+  REDIS_TTL: "57600"
+  BENTO_BINARY_URL: "https://github.com/boundless-xyz/boundless/releases/download/bento-v1.2.0/bento-bundle-linux-amd64.tar.gz"
+```
+
+## Usage
+
+### Basic Deployment
+
+```bash
+ansible-playbook -i inventory.yml bento-standalone.yml
```
willpote: Looks out of date? I dont see a file bento-standalone

.github/workflows/ansible.yml
```diff
@@ -0,0 +1,172 @@
...
+  workflow_dispatch:
+    inputs:
+      playbook:
+        description: 'Playbook to run'
+        required: true
+        type: choice
+        options:
+          - cluster.yml
+          - bento-worker.yml
+          - broker.yml
```
willpote: I don't see these files in the PR

.github/workflows/prover-deploy.yml
```diff
@@ -0,0 +1,143 @@
+name: Prover Deployment
```
willpote, line 1: Can we rename file to `prover-deploy.yml`

ansible/roles/bento/README.md
```diff
@@ -0,0 +1,125 @@
+# Bento Role
```
willpote: This also deploys and runs the broker right? I think a better name for this would just be `prover` then, since its not just running Bento?

compose.yml
```diff
@@ -58,12 +72,33 @@ x-broker-environment: &broker-environment
...
+  environment:
+    <<: *base-environment
+    REDIS_TTL: ${REDIS_TTL:-57600}
+  # Detect GPUs and spawn one agent per GPU, then wait for all
+  entrypoint:
+    [
+      "/bin/sh",
+      "-c",
+      "for gpu in $(nvidia-smi --query-gpu=index --format=csv,noheader,nounits); do CUDA_VISIBLE_DEVICES=$$gpu /app/agent -t prove --redis-ttl $$REDIS_TTL & done; wait",
+    ]
```
willpote, line 88: If we're automating this here, can you add a follow up task to the board to update the wizard to not try and do this

ansible/roles/prover/configs/broker/8453.toml
willpote: I'm also unclear on where this file is used- i don't see it referenced anywhere else

.github/workflows/prover-deploy.yml
```diff
@@ -0,0 +1,143 @@
...
+permissions:
+  contents: read
+
+jobs:
+  deploy-nightly:
+    name: Deploy Nightly
+    runs-on: ubuntu-latest
+    # Run on schedule OR when manually triggered with nightly/all target
+    if: |
+      github.event_name == 'schedule' ||
```
willpote, line 42: I don't see in the PR where we configure the chains and stages (staging/prod) that we are deploying to. Where is it configured that running this job will redeploy "staging cluster on chain 84532 with toml `staging-84532.toml`", or running release deploys two clusters, using tomls?

ansible/README.md
```diff
@@ -0,0 +1,394 @@
...
+├── bento-standalone.yml  # Standalone Bento deployment with Docker Compose
+├── broker.yml             # Broker deployment
+├── inventory.yml          # Inventory file
+├── ansible.cfg            # Ansible configuration
+├── README.md              # This file
+├── ENV_VARS.md            # Environment variable documentation
+└── roles/
+    ├── awscli/             # AWS CLI v2 installation
+    ├── bento/              # Bento agent service (uses launcher scripts)
+    ├── bento/              # Bento Docker Compose deployment
```
willpote: Dup

ansible/README.md
```diff
@@ -0,0 +1,394 @@
...
+- Sudo/root privileges on target hosts
+- Ansible collections (install with `ansible-galaxy collection install community.postgresql`)
+
+## Directory Structure
+
+```
+ansible/
+├── cluster.yml            # Full cluster deployment (bento + broker + dependencies)
+├── bento-worker.yml      # Bento worker deployment (no dependencies)
+├── bento-standalone.yml  # Standalone Bento deployment with Docker Compose
```
willpote: Out of date?

ansible/roles/bento/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+This Ansible role deploys Bento services using Docker Compose, providing a standalone deployment option that's ideal for:
```
willpote: I'm confused about exactly what we are running in latitude- are we deploying clusters? Or large single instance machines? (I thought the former, but seems from the PR may be the latter?)

ansible/roles/prover/configs/broker/8453.toml
zeroecco: the file url is referenced in the inventory yaml. it is pulled that way

zeroecco: was wondering about nightly. I wasn't sure if we wanted a fully simplified. Will update.

.github/workflows/prover-deploy.yml
zeroecco: sure

compose.yml
zeroecco: done

ansible/roles/bento/README.md
zeroecco: We are deploying large single nodes. technically they are "clusters" but they exist on a single server.

.github/workflows/prover-deploy.yml
zeroecco: This is configured in the inventory file and not a part of this pr. it is stored with secrets.

ansible/roles/bento/templates/docker-compose.env.j2
zeroecco: 3db4bbbf

PR #1438, opened 2025-12-12
BM-2134: update keccak input key to fix issue with keeping cleanup

austinabell:
A more practical workaround than #1431, but with the tradeoff that it will not work when using a previous version in the prove agent and this change for the executor version -- so it probably can't be included in a patch release.


austinabell:
marked as draft so it doesn't get pulled into next release, feel free to review anytime though

zeroecco (approved):

PR #1441, opened 2025-12-15
BM-2146: indexer: add cycle counts via bento API

ncocchiaro:
This PR extends the indexer to preflight requests via the bento API to add cycle counts for all requests.

### DIFF
--- a/crates/indexer/src/db/market.rs
+++ b/crates/indexer/src/db/market.rs
@@ -16,6 +16,7 @@ use std::{
     collections::{HashMap, HashSet},
     str::FromStr,
     sync::Arc,
+    time::Duration,
 };
 
 use super::DbError;
@@ -25,9 +26,10 @@ use boundless_market::contracts::{
     AssessorReceipt, Fulfillment, FulfillmentDataType, Predicate, PredicateType, ProofRequest,
     RequestInputType,
 };
+use log::LevelFilter;
 use sqlx::{
     any::{install_default_drivers, AnyConnectOptions, AnyPoolOptions},
-    AnyPool, Row,
+    AnyPool, ConnectOptions, Row,
 };
 
 const SQL_BLOCK_KEY: i64 = 0;
@@ -836,6 +838,19 @@ pub trait IndexerDb {
         cursor: Option<B256>,
         limit: i64,
     ) -> Result<Vec<B256>, DbError>;
+
+    /// Gets all request digests from request_status table up to the given end_timestamp.
+    /// Returns Vec of (request_digest, created_at) tuples.
+    async fn get_all_request_digests(
+        &self,
+        cursor: Option<(u64, B256)>,
+        end_timestamp: u64,
+        limit: i64,
+    ) -> Result<Vec<(B256, u64)>, DbError>;
+
+    /// Gets the count of request digests in request_status table filtered by end_timestamp.
+    /// Used for logging total digests to process during backfill.
+    async fn count_request_digests_by_timestamp(&self, end_timestamp: u64) -> Result<i64, DbError>;
 }
 
 pub type DbObj = Arc<MarketDb>;
@@ -858,10 +873,11 @@ impl MarketDb {
         pool_options: Option<AnyPoolOptions>,
         skip_migrations: bool,
     ) -> Result<Self, DbError> {
-        use std::time::Duration;
-
         install_default_drivers();
-        let opts = AnyConnectOptions::from_str(conn_str)?;
+        let mut opts = AnyConnectOptions::from_str(conn_str)?;
+
+        // Configure slow query logging: only log queries that take over 2 seconds
+        opts = opts.log_slow_statements(LevelFilter::Warn, Duration::from_secs(2)); // Only warn for queries > 2s
 
         let pool = if let Some(pool_opts) = pool_options {
             pool_opts.connect_with(opts).await?
@@ -3426,6 +3442,65 @@ impl IndexerDb for MarketDb {
 
         Ok(digests)
     }
+
+    async fn get_all_request_digests(
+        &self,
+        cursor: Option<(u64, B256)>,
+        end_timestamp: u64,
+        limit: i64,
+    ) -> Result<Vec<(B256, u64)>, DbError> {
+        let rows = if let Some((cursor_ts, cursor_digest)) = cursor {
+            // Format without 0x prefix to match how digests are stored in the database
+            let cursor_digest_hex = format!("{:x}", cursor_digest);
+            sqlx::query(
+                "SELECT request_digest, created_at 
+                 FROM request_status
+                 WHERE created_at <= $1
+                   AND (created_at > $2 OR (created_at = $2 AND request_digest > $3))
+                 ORDER BY created_at ASC, request_digest ASC
+                 LIMIT $4",
+            )
+            .bind(end_timestamp as i64)
+            .bind(cursor_ts as i64)
+            .bind(cursor_digest_hex)
+            .bind(limit)
+            .fetch_all(&self.pool)
+            .await?
+        } else {
+            sqlx::query(
+                "SELECT request_digest, created_at 
+                 FROM request_status
+                 WHERE created_at <= $1
+                 ORDER BY created_at ASC, request_digest ASC
+                 LIMIT $2",
+            )
+            .bind(end_timestamp as i64)
+            .bind(limit)
+            .fetch_all(&self.pool)
+            .await?
+        };
+
+        let mut results = Vec::new();
+        for row in rows {
+            let digest_hex: String = row.try_get("request_digest")?;
+            let digest = B256::from_str(&digest_hex)
+                .map_err(|e| DbError::BadTransaction(format!("Invalid digest: {}", e)))?;
+            let created_at = row.get::<i64, _>("created_at") as u64;
+            results.push((digest, created_at));
+        }
+
+        Ok(results)
+    }
+
+    async fn count_request_digests_by_timestamp(&self, end_timestamp: u64) -> Result<i64, DbError> {
+        let count: i64 =
+            sqlx::query_scalar("SELECT COUNT(*) FROM request_status WHERE created_at <= $1")
+                .bind(end_timestamp as i64)
+                .fetch_one(&self.pool)
+                .await?;
+
+        Ok(count)
+    }
 }
 
 impl MarketDb {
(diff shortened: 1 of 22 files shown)


infra/indexer/components/market-indexer.ts
```diff
@@ -22,6 +22,8 @@ export interface MarketIndexerArgs {
   dockerRemoteBuilder?: string;
   orderStreamUrl?: pulumi.Output<string>;
   orderStreamApiKey?: pulumi.Output<string>;
+  bentoApiUrl: pulumi.Output<string>;
+  bentoApiKey: pulumi.Output<string>;
```
willpote: Lets make these optional so its possible to run an indexer without doing execution (e.g. if wanting to do a dev deployment to test)

infra/indexer/Pulumi.l-prod-1.yaml
```diff
@@ -18,3 +18,7 @@ config:
     secure: v1:Bm3FUPbytchzaS94:uO0iW9AUegk+SkbPOOpmnwfn3rRZjmSznZ4/NM2l0o6Bqon5qKkxcuTffdicUDHeE7sA9D9UrFGpXkHGgHIINhBkCaRw+m7wu9Y=
   indexer:RDS_PASSWORD:
     secure: v1:Dgy+uSuiBgyEcb13:XtMKpSotMXWM3UnaY0RehuUKvF2IliQRxDQH0RgN4w==
+  indexer:BENTO_API_URL:
+    secure: v1:241smsVYdJrp3x97:4Xphhu28RYX3EYc4s2JaNU9R7tROCcQjbcrugpCVYgrC+NM5n505nR8OFfzg0Bu2LGuoTKRsBSAxtPJNpq/EpjZcf9A+gvqcPloXCWRi/Ia5iWObnXU=
+  indexer:BENTO_API_KEY:
+    secure: v1:6NfJyGHR0KZ5ILgG:WGwjEGCZMDw9aJjBhcVSbaiP4W23yg/8p27V6I5a6ol56QVlb70d2p7JaJ4yv9/eyYskkpZm4/EeR9M=
```
willpote: We shouldn't need this on chain 1, since we don't run a market there, so the market indexer never starts

infra/indexer/index.ts
```diff
@@ -68,6 +68,8 @@ export = () => {
     const logsEthRpcUrl = isDev ? pulumi.output(getEnvVar("LOGS_ETH_RPC_URL")) : config.requireSecret('LOGS_ETH_RPC_URL');
     const orderStreamApiKey = isDev ? pulumi.output(getEnvVar("ORDER_STREAM_API_KEY")) : config.requireSecret('ORDER_STREAM_API_KEY');
     const orderStreamUrl = isDev ? pulumi.output(getEnvVar("ORDER_STREAM_URL")) : config.getSecret('ORDER_STREAM_URL');
+    const bentoApiUrl = isDev ? pulumi.output(getEnvVar("BENTO_API_URL")) : config.requireSecret('BENTO_API_URL');
+    const bentoApiKey = isDev ? pulumi.output(getEnvVar("BENTO_API_KEY")) : config.requireSecret('BENTO_API_KEY');
```
willpote: Same here, should be optional

infra/indexer/Pulumi.l-staging-11155111.yaml
```diff
@@ -14,5 +14,9 @@ config:
     secure: v1:aDACh+FhCJte0D8a:x1Qz6hBhdgjHUAhu3m5NBvipn0MFctp5PRSNttCgV+7mPVgUU7+mvkRDwEa+K8Ll1IUToGX/eKBkwBeE+XYrWbuv+5EFsE9SHiY=
   indexer:RDS_PASSWORD:
     secure: v1:w4gH1O7RvkOIKDd+:QSjcdDB9Fz4L45pDn/KqazFQ7f3jmVqk6huQozjQ
+  indexer:BENTO_API_URL:
+    secure: v1:3StxfNxj3frkUJNs:KM7FbnTvZPYRzMYlI51pV6s95QNH/VL7a3cTYROdhyC6Aq2oCmYy+LpoYCRKJQ1wloin2CIIKy2IDWFbkjhqvpmMmQ20FgxsIH8Uk+BiTJVta+5RUOMOfZg=
+  indexer:BENTO_API_KEY:
+    secure: v1:Sk50eYSjW0shSH1l:dHJXvw4nRjWYKvwpaiLtLJz0FrH1sCrxWd9Z/76RPCSSVNLbPuaxKY2GzqgkfC/ne9k9PehrjC+Ngn0=
```
willpote: Same here, no staging market on eth sep

crates/indexer/src/db/market.rs
```diff
@@ -2230,6 +2273,138 @@ impl IndexerDb for MarketDb {
         Ok(request_digests)
     }
 
+    async fn get_cycle_counts_pending(&self, limit: u32) -> Result<HashSet<B256>, DbError> {
+        let query =
+            "SELECT request_digest FROM cycle_counts WHERE cycle_status = 'PENDING' LIMIT $1";
+
+        let rows = sqlx::query(query).bind(limit as i64).fetch_all(self.pool()).await?;
```
willpote, line 2299: I feel like we should have some sort of ordering here, probably makes sense to process the oldest ones first?

crates/indexer/src/market/service/execution.rs
```diff
@@ -0,0 +1,383 @@
...
+        config.bento_api_key.clone().unwrap(),
+        risc0_zkvm::VERSION,
+    )
+    .unwrap();
+
+    loop {
+        interval.tick().await;
+
+        // Find cycle counts still in state PENDING
+        tracing::debug!("Querying DB for cycle counts in status PENDING...");
```
willpote: It would be useful to log the size of the queue here so we can easily see how far behind we are, e.g. add a `SELECT COUNT(*) ... where PENDING` query on the table and have a log like `Currently X orders pending cycle counts, Y orders currently executing`

crates/indexer/src/db/market.rs
```diff
@@ -2230,6 +2273,138 @@ impl IndexerDb for MarketDb {
...
+        }
+
+        Ok(request_digests)
+    }
+
+    async fn get_cycle_counts_executing(
+        &self,
+        limit: u32,
+    ) -> Result<HashSet<CycleCountExecution>, DbError> {
+        let query = "SELECT request_digest, session_uuid FROM cycle_counts WHERE cycle_status = 'EXECUTING' LIMIT $1";
```
willpote: Same here re: ordering

crates/indexer/src/market/service/execution.rs
```diff
@@ -0,0 +1,383 @@
...
+use boundless_market::storage::fetch_url;
+use broker::futures_retry::retry;
+use bytes::Bytes;
+use std::collections::HashSet;
+
+pub async fn execute_requests(db: DbObj, config: IndexerServiceExecutionConfig) {
+    tracing::info!("Started execution task");
+
+    const PENDING_LIMIT: u32 = 10;
+    const EXECUTING_LIMIT: u32 = 40;
```
willpote: It seems that the default config means that every 3 seconds we will submit 40 execution requests.

Do we need some sort of limit on the amount of requests we have 'EXECUTING'? During the initial backfill especially we will send a lot of requests, and its not clear to me how this will impact the clusters ability to do other work (e.g. prove market orders)

ncocchiaro: Not exactly, every 3 seconds we'll submit up to 10, then check status on up to 40 to update their status relatively quickly, but I agree we can limit this more strictly. I'll do that and try to make the wording clearer

infra/indexer/components/market-indexer.ts
ncocchiaro: Should be fixed in https://github.com/boundless-xyz/boundless/pull/1441/commits/311671185d41421effdcf5004e4f97a01453ba52

infra/indexer/Pulumi.l-prod-1.yaml
ncocchiaro: Should be fixed in https://github.com/boundless-xyz/boundless/pull/1441/commits/311671185d41421effdcf5004e4f97a01453ba52

infra/indexer/index.ts
ncocchiaro: Optional now in https://github.com/boundless-xyz/boundless/pull/1441/commits/311671185d41421effdcf5004e4f97a01453ba52

infra/indexer/Pulumi.l-staging-11155111.yaml
ncocchiaro: https://github.com/boundless-xyz/boundless/pull/1441/commits/311671185d41421effdcf5004e4f97a01453ba52

crates/indexer/src/db/market.rs
ncocchiaro: Ordered by updated_at in https://github.com/boundless-xyz/boundless/pull/1441/commits/311671185d41421effdcf5004e4f97a01453ba52

crates/indexer/src/market/service/execution.rs
ncocchiaro: Added a function to count rows in status PENDING, EXECUTING and FAILED, used for logging currently but could be useful later for other logic. https://github.com/boundless-xyz/boundless/pull/1441/commits/311671185d41421effdcf5004e4f97a01453ba52

crates/indexer/src/db/market.rs
ncocchiaro: Ordered by updated_at in https://github.com/boundless-xyz/boundless/pull/1441/commits/311671185d41421effdcf5004e4f97a01453ba52

crates/indexer/src/market/service/execution.rs
ncocchiaro: In https://github.com/boundless-xyz/boundless/pull/1441/commits/311671185d41421effdcf5004e4f97a01453ba52 I've lowered the limits and made it so we're only submitting execution requests up to the specified limit, based on how many we know we have already in flight.

nahoc: FWIW I had 20 running orders at a time MAX in my implementation. Me and Rich thought about making it 50, which seemed to be fine. I would use that as a starting point and increase after some testing.

ncocchiaro: So I made that max default to 20 now, but also removed the constants and made these limits configurable via command line args.

crates/indexer/src/db/market.rs
nahoc: Actually makes more sense to process newest one first imo, because if you get a lot of orders it's gonna be hard to catch up, especially in the beginning.

I understand wanting to go from oldest first, but only after backfill is complete.

crates/indexer/src/db/market.rs
```diff
@@ -2353,22 +2353,30 @@ impl IndexerDb for MarketDb {
...
+                    SET cycle_status = 'COMPLETED', program_cycles = $1, total_cycles = $2
+                    WHERE request_digest = $3";
+            sqlx::query(query)
                 .bind(u256_to_padded_string(update_data.program_cycles))
                 .bind(u256_to_padded_string(update_data.total_cycles))
-                .bind(format!("{:x}", update_data.request_digest));
-            query_builder.execute(&mut *tx).await?;
+                .bind(format!("{:x}", update_data.request_digest))
+                .execute(&mut *tx)
+                .await?;
```
willpote: We shouldn't need to do this here

When we call `get_requests_comprehensive` in core.rs, we should join the cycles table and get the cycles. That is the object that gets provided to `status.rs`, and `status.rs` should be the only place responsible for updating the status table

crates/indexer/src/db/market.rs
ncocchiaro: In https://github.com/boundless-xyz/boundless/pull/1441/commits/8dfb084861a0024e474f985580211fae39a80822 we're sorting by newest order first

crates/indexer/src/db/market.rs
ncocchiaro: Turned out the inconsistencies here were due to a bug with update timestamps fixed in https://github.com/boundless-xyz/boundless/pull/1441/commits/8dfb084861a0024e474f985580211fae39a80822, and my local test indexer not being caught up enough to pick up the updated cycle counts. I've removed this table update

crates/indexer/src/db/market.rs
```diff
@@ -2246,6 +2292,162 @@ impl IndexerDb for MarketDb {
...
+    }
+
+    async fn set_cycle_counts_executing(
+        &self,
+        execution_info: &HashSet<CycleCountExecution>,
+    ) -> Result<(), DbError> {
+        let execution_vec: Vec<&CycleCountExecution> = execution_info.iter().collect();
+
+        let current_timestamp =
+            std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH).unwrap().as_secs();
```
willpote, line 2349: We should use CURRENT_TIMESTAMP here and elsewhere. e.g.: https://github.com/boundless-xyz/boundless/blob/willpote/cluster-prio/crates/indexer/src/db/market.rs#L3238

willpote (approved):
Missing tests, so approving assuming manual testing. We should add in a follow up

crates/indexer/src/db/market.rs
ncocchiaro: Currently done this way due to BIGINT type but we can revisit

PR #1442, opened 2025-12-15
BM-2148: Improved logging around proof generation and submission

willpote:
Currently its hard to tie proving errors to the orders that caused them. This makes it super clear when proofs are run for each order id, and when they fail/retry

This also should help debug an issue with fulfillment where we see "RequestIsNotPricedOrLocked" reverts. Its unclear which contract function we are trying to execute when we see this, which makes debugging difficult.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,5 +3,4 @@
 # Note: Order is important; the last matching pattern takes the most
 # precedence.
 
-*                                      @boundless-xyz/maintainers @capossele @Wollac @zeroecco @austinabell @willemolding @zeroecco @ec2
-/docs/                                 @boundless-xyz/maintainers @capossele @Wollac @zeroecco @austinabell @willemolding @zeroecco @ec2 @neutronmoderator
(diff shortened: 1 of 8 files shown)


crates/broker/src/aggregator.rs
```diff
@@ -134,19 +135,23 @@ impl AggregatorService {
...
+                        "Error fetching proof from batch: {e:?} containing orders {:x?}, excluding",
+                        all_orders
+                    );
                 }
             }
         }
 
         if claims.is_empty() {
-            anyhow::bail!("No valid proofs found in batch");
+            anyhow::bail!(format!("No valid proofs found in batch with orders {:x?}", all_orders));
```
austinabell: nit: Curious why the `:x?` formatting for strings, assuming this is already hex formatted?

Tested in Rust playground that it seems the hex formatting just gets ignored, so fine to just leave

austinabell (approved):

crates/broker/src/aggregator.rs
willpote: Oops fixed

PR #1445, opened 2025-12-16
BM-2154: Support retries for event queries in SDK

willpote:
Due to RPC data inconsistency issues, its possible to query for event logs looking for things like Fulfillment data, and miss it as the RPC provider has fallen behind on eth_blockNumber or eth_getLogs. We've had customers hit this issue when using our default SDK.

* Changes the SDK to by default include a retry with a 5s delay, and to query up to around 1 week of events
* Aims to not change current behavior in broker (no retries) + using the same block_range / iterations as currently
* Doing this involved creating our own EventQueryConfig rather than using the one from r0. It makes sense for us to be able to control the config struct anyway
* Also improved the error message when you fail to find the event

### DIFF
--- a/crates/boundless-market/src/contracts/boundless_market.rs
+++ b/crates/boundless-market/src/contracts/boundless_market.rs
@@ -44,7 +44,6 @@ use super::{
 };
 
 /// Retry configuration for query operations.
-
 #[derive(Clone, Debug)]
 #[non_exhaustive]
 pub struct EventRetryConfig {


crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -44,6 +43,63 @@ use super::{
...
+    /// where calls to fetch events or current block numbers return stale data. For example,
+    /// if an event was just emitted but the RPC provider hasn't indexed it yet, a retry
+    /// after a short delay may succeed.
+    pub retries: u32,
+    /// Backoff time in seconds between retry attempts.
+    ///
+    /// This delay helps handle transient RPC provider issues where data may be temporarily
+    /// inconsistent or unavailable. A short backoff (e.g., 5 seconds) is typically sufficient
+    /// to allow the RPC provider to catch up with the latest blockchain state.
+    pub retries_backoff_s: u64,
```
austinabell: this being ms seems like a safer path to future proof if this needs to be tuned more tightly for specific cases. Optional.

Also, backoff seems very misleading, since the retry is not a backoff at all?

austinabell: If we are exposing all fields, we should probably add the `#[non_exhaustive]` tag to make this able to be modified in the future without being breaking. I think it makes sense to have these be public, because the `new` constructor would be very error prone just being all integers (to the point where we might not want to include), but this would be very restrictive to make changes in the future without breaking this.

crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -1023,55 +1150,63 @@ impl<P: Provider> BoundlessMarketService<P> {
...
-        let start_block = lower_bound.unwrap_or(upper_block.saturating_sub(
-            self.event_query_config.block_range * self.event_query_config.max_iterations,
-        ));
+        let config = self.event_query_config.clone();
+        self.retry_query(&config, || async {
+            let mut upper_block = upper_bound.unwrap_or(self.get_latest_block_number().await?);
+            let initial_upper_block = upper_block;
+            let start_block = lower_bound.unwrap_or(upper_block.saturating_sub(
+                self.event_query_config.block_range * self.event_query_config.max_iterations,
+            ));
```
austinabell, line 1173: So just a thought, but I wonder if the better long-term path for these is to do an `eth_call` for `requestIsFulfilled(..)` over the block range, and then just query logs through `eth_getLogs` once we know the range. Given the logs call is >2x more expensive from a CU perspective than an `eth_call` for a provider like alchemy and you can rule out ranges larger than the event query range, it seems like it would be cheaper and quicker.

The other main benefit to doing this is that for the problem we are solving for, we would be able to more easily detect, and then just search the range we missed (instead of backsearching two weeks of events per retry). Would also allow us to retry more frequently without worrying about costs or delays.

willpote: I had a similar thought around using the request info to restrict the search range. Generally trying to not make major changes before the break, so I captured the idea in https://linear.app/boundlessnetwork/issue/BM-2152/fix-sdk-footgun-around-event-query-block-length

I'll add your  idea there too

crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -44,6 +43,71 @@ use super::{
     Offer, ProofRequest, RequestError, RequestId, RequestStatus, TxnErr, TXN_CONFIRM_TIMEOUT,
 };
 
+/// Retry configuration for query operations.
+
```
austinabell:
```suggestion
```

austinabell (approved):

PR #1455, opened 2025-12-17
BM-2166: Additional tests for indexer cycle count logic

ncocchiaro:
Refactors and adds a number of unit tests and integration tests to the cycle count execution portions of the indexer.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,4 +3,4 @@
 # Note: Order is important; the last matching pattern takes the most
 # precedence.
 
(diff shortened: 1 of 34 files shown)


crates/indexer/src/db/market.rs
```diff
@@ -6138,62 +6152,20 @@ mod tests {
     }
 
     // Helper to create test proof requests and cycle counts
-    async fn setup_test_requests_and_cycles(
```
ncocchiaro, line 6141: Moved this to test_utils to use it in other tests

willpote (approved):

PR #1467, opened 2025-12-19
BM-2186: Restart cycle count execution task on failure. Handle digests without entry in proof request table

willpote:
- Restart execution task on failure. 
- CLI arg clean-up
- Handle requests without request ids

### DIFF
--- a/crates/indexer/src/market/service/core.rs
+++ b/crates/indexer/src/market/service/core.rs
@@ -349,7 +349,7 @@ async fn run_execution_task_supervisor(db: DbObj, config: IndexerServiceExecutio
 
         match handle.await {
             Ok(()) => {
-                tracing::warn!(
+                tracing::error!(
                     "Cycle count execution task returned unexpectedly, restarting in {} seconds",
                     RESTART_DELAY_SECS
                 );


crates/indexer/src/db/market.rs
```diff
@@ -449,13 +449,19 @@ pub struct CycleCountExecution {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct RequestWithId {
-    pub request_id: U256,
+    /// Some requests may not have a corresponding entry in the proof_requests table. So we can fail to find
+    /// the request id. This can happen if the request was fulfilled without it being submitted on-chain,
+    /// or via our known off-chain order stream.
```
austinabell: I'm confused, if it's fulfilled then it will for sure have a request ID?

crates/indexer/src/market/service/core.rs
```diff
@@ -58,26 +61,15 @@ where
...
-                } else {
-                    tracing::info!("Bento API URL or key not provided, not starting executor task");
-                }
-            }
-            None => {
-                tracing::info!("Execution configuration not found, not starting executor task");
-            }
-        };
+        if let Some(execution_config) = config_clone.execution_config {
+            tokio::spawn(run_execution_task_supervisor(db_clone, execution_config));
```
austinabell, line 70: Maybe add a comment that is is expected not to keep the handle to this task and check errors because it's assumed this `run_execution_task_supervisor` won't panic/error. I'm assuming that's why it's like this?

crates/indexer/src/market/service/core.rs
```diff
@@ -340,6 +332,64 @@ fn find_starting_block(
...
+        tokio::time::sleep(tokio::time::Duration::from_secs(RESTART_DELAY_SECS)).await;
+
+        // Respawn the task
+        tracing::info!("Respawning cycle count execution task");
+        let db_clone = db.clone();
+        let config_clone = config.clone();
+        tasks.spawn(async move {
+            execute_requests(db_clone, config_clone).await;
+        });
+    }
```
austinabell: Optional: this seems like it could be simplified rather than keeping a `JoinSet` for just one task. Sth like:

```
loop {
    tracing::info!("Starting cycle count execution task");
    let db_clone = db.clone();
    let config_clone = config.clone();

    let handle = tokio::spawn(async move {
        execute_requests(db_clone, config_clone).await;
    });

    match handle.await {
        Ok(()) => {
            tracing::warn!(
                "Cycle count execution task returned unexpectedly, restarting in {} seconds",
                RESTART_DELAY_SECS
            );
        }
		// TODO: Shouldn't need to differentiate panic vs cancel for this to simplify if you want also
        Err(e) if e.is_panic() => {
            tracing::error!(
                "Cycle count execution task panicked, restarting in {} seconds",
                RESTART_DELAY_SECS
            );
        }
        Err(e) => {
            tracing::error!(
                "Cycle count execution task cancelled ({e}), restarting in {} seconds",
                RESTART_DELAY_SECS
            );
        }
    }

    sleep(Duration::from_secs(RESTART_DELAY_SECS)).await;
}
```

willpote: Thanks, thats much better

crates/indexer/src/db/market.rs
willpote: That is true, the comment is a bit misleading. This is really more a quirk of the indexer. If we only see fulfill, but no request submission we don't populate the proof_request table at all because we would be missing so much other data  e.g. submission time, image url, etc.

Will update comment

PR #1475, opened 2026-01-05
BM-2206: Indexer- simplify cloudfront caching

willpote:
We are seeing intermittent Cloudfront errors. We are using the legacy method for defining Cloudfront caching, which supposedly can cause issues. Migrating to the latest recommend setup.

### DIFF
--- a/infra/indexer/components/indexer-api.ts
+++ b/infra/indexer/components/indexer-api.ts
@@ -355,7 +355,7 @@ export class IndexerApi extends pulumi.ComponentResource {
             priority: 1,
             statement: {
               rateBasedStatement: {
-                limit: 200, // 150 requests per 5 minutes per IP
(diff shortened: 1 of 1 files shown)


infra/indexer/components/indexer-api.ts
```diff
@@ -355,12 +355,8 @@ export class IndexerApi extends pulumi.ComponentResource {
             priority: 1,
             statement: {
               rateBasedStatement: {
-                limit: 150, // 150 requests per 5 minutes per IP
+                limit: 200, // 150 requests per 5 minutes per IP
```
ncocchiaro:
```suggestion
                limit: 200, // 200 requests per 5 minutes per IP
```

ncocchiaro (approved):

PR #1477, opened 2026-01-06
BM-1976: add dynamic allowed requestor list

capossele:


crates/broker/src/config.rs
```diff
@@ -105,6 +105,10 @@ mod defaults {
         ])
     }
 
+    pub fn allowed_requestor_lists() -> Option<Vec<String>> {
```
willpote: Lets call it `allow_requestor_lists` for consistency since the other field is `allow_client_addresses`

requestor-lists/boundless-allowed-list.json
```diff
@@ -0,0 +1,134 @@
+{
+  "name": "Boundless Allowed List",
+  "description": "List of requestors that are allowed to submit requests. If this list is configured, only requestors in this list will be accepted.",
```
willpote: Can we add a note that this is just for internal usage- just to avoid potential confusion for external provers as its in the same directly as the priority lists

PR #1495, opened 2026-01-12
BM-2235: Add prover aggs, split aggs to own task

willpote:
Adds prover aggregations, similar to how we currently have requestor aggregates.

I also move the aggregations logic to its own task. This is so that computing aggregates, which is slow, does not block the processing of new events and updating of statuses.This allows the orders list page and order details pages to be close to realtime.


ncocchiaro (approved):
Ideally smaller chunks would be easier to review but seems reasonable to me

PR #1496, opened 2026-01-12
BM-2237: Add trait-based price provider support

capossele:
Introduces a `PriceProvider` trait for market price fetching and integrates it into `ClientBuilder` and `OfferLayer`.

- Add PriceProvider trait and PriceProviderArc type alias
- Add ClientBuilder::with_price_provider() for custom implementations
- Use Deployment.indexer_url automatically when no price provider is set
- Add test_helpers.rs module for IndexerClient test utilities
- Update e2e tests to use the new API

### DIFF
--- a/crates/boundless-market/src/indexer_client.rs
+++ b/crates/boundless-market/src/indexer_client.rs
@@ -231,41 +231,49 @@ impl IndexerClient {
     ) -> Result<MarketAggregatesResponse> {
         let mut url = self.base_url.join("v1/market/aggregates").context("Failed to build URL")?;
 
-        // Build query parameters
-        let mut query_pairs = url.query_pairs_mut();
-        query_pairs.append_pair("aggregation", &params.aggregation.to_string());
-        if let Some(ref cursor) = params.cursor {
-            query_pairs.append_pair("cursor", cursor);
+        // Build query parameters manually to avoid non-Send closures from query_pairs_mut()
+        // Collect all query parameters as owned strings first
+        let mut query_parts = Vec::new();
+        query_parts.push(format!("aggregation={}", params.aggregation));
+        if let Some(cursor) = params.cursor {
+            // URL encode cursor manually
+            let encoded_cursor: String =
+                url::form_urlencoded::byte_serialize(cursor.as_bytes()).collect();
+            query_parts.push(format!("cursor={}", encoded_cursor));
         }
         if let Some(limit) = params.limit {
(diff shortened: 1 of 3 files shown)


crates/boundless-market/src/indexer_client.rs
```diff
@@ -0,0 +1,568 @@
...
+}
+
+/// Price range with minimum and maximum values as U256.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PriceRange {
+    /// Minimum price (p10 percentile) in wei.
+    pub min: U256,
+    /// Maximum price (p99 percentile) in wei.
+    pub max: U256,
+}
```
willpote: The usage of p10 and p99 as min and max is an opinionated default for the SDK that I think we should keep out of the indexer client, as it reduces its re-usability. 

So essentially proposing we just have generic methods in the client that return _all_ the percentiles we compute, and then in the default price provider in the SDK, we choose p10 and p99 there

crates/boundless-market/src/indexer_client.rs
```diff
@@ -0,0 +1,568 @@
...
+
+impl TryFrom<PricePercentiles> for PriceRange {
```
willpote: Similarly here price range feels like a concept specific to the SDK and our opinionated p10/p99 methodology, so would move it there

crates/boundless-market/src/client.rs
```diff
@@ -346,13 +351,34 @@ impl<St, Si> ClientBuilder<St, Si> {
...
+            let url = Url::parse(url_str.as_ref()).with_context(|| {
+                format!("Failed to parse indexer URL from deployment: {}", url_str)
+            })?;
+            let indexer_client = IndexerClient::new(url).with_context(|| {
+                format!("Failed to create indexer client from deployment indexer URL: {}", url_str)
+            })?;
+            Some(Arc::new(indexer_client))
+        } else {
+            None
+        };
```
willpote, line 385: I think our default price provider should have an on-chain fallback, i worry about clients using this and then some issue with the indexer API taking down their requestors

We have a sample of fetching market prices from on-chain in the wizard that coudl be refactored for re-use here

capossele: Addressed all of your feedback in https://github.com/boundless-xyz/boundless/pull/1496/commits/58947d2012b35ebd04a604cb1a3c5e24385955b5
It's a big change though

PR #1497, opened 2026-01-12
WEB3-607: Unified Download/Upload Handlers

Wollac:
## Description

Consolidated all storage handling into a unified system in `boundless-market`:

**New Architecture:**

- StorageDownloader / StorageUploader traits defining the interface
- DefaultDownloader that auto-selects provider based on URL scheme (s3://, https://, gcs://, file://)
- Consistent provider implementations: HTTP, S3, GCS, File, Pinata
- Shared credential and configuration handling across upload and download paths

**Broker Changes:**
- Replaced custom handler code with ConfigurableDownloader wrapper
- Downloads now use the same credential chain as uploads
- Configuration changes (max size, retries, cache dir) are picked up dynamically

### 1. Renamed Types & Traits
  - `StorageProvider` → `StorageUploader`
  - `StorageProviderConfig` → `StorageUploaderConfig`
  - `StorageProviderType` → `StorageUploaderType`
  - `StandardStorageProvider` → `StandardUploader`
  - Individual providers renamed (e.g., `S3StorageProvider` → `S3StorageUploader`, `PinataStorageProvider` → `PinataStorageUploader`)

  ### 2. New Downloader Abstraction
  - Added `StorageDownloader` trait and `StandardDownloader` type
  - `Client` now takes separate `U` (uploader) and `D` (downloader) generics instead of single `St` (storage)

  ### 3. API Changes
  - `ClientBuilder::with_storage_provider()` → `with_uploader()`
  - `ClientBuilder::with_storage_provider_config()` → `with_uploader_config()` (now `async`)
  - `StandardUploader::from_config()` and `from_env()` are now `async`
  - `Client.storage_provider` field → split into `uploader` and `downloader` fields

  ### 4. Removed Exports
  - `fetch_url()` function removed (use `StorageDownloader::download()` instead)
  - `override_gateway()` removed from public API
  - `MockStorageProvider` → `MockStorageUploader` (now behind `test-utils` feature)
  - `storage_provider_from_env()` and `storage_provider_from_config()` standalone functions removed (use `StandardUploader::from_config()` instead)

  ### 5. Feature Flags
  - S3 support requires `s3` feature flag
  - GCS support added (requires `gcs` feature flag)
  - Tests requiring Dev mode (`test-r0vm`) and tests requiring RPC (`test-rpc`) now use feature instead of `ignore`

  ### 6. S3 Changes
  - `S3_ACCESS_KEY` → `AWS_ACCESS_KEY_ID`
  - `S3_SECRET_KEY` → `AWS_SECRET_ACCESS_KEY`
  - Credentials now auto-resolve via AWS SDK default chain (env vars, `~/.aws/credentials`, IAM roles)
  - Presigned URL expiry increased from 1 hour to 7 days
  - Removed automatic bucket creation on upload

### 7. Broker Config Changes
- `MarketConf` now also contains `ipfs_gateway_fallback`. This is optional so previous config files continue to work. Previously setting the fallback was possible by env var this has been changed to a config option.

fixes WEB3-607
fixes BM-2029

### DIFF
--- a/infra/indexer/components/indexer-infra.ts
+++ b/infra/indexer/components/indexer-infra.ts
@@ -138,7 +138,7 @@ export class IndexerShared extends pulumi.ComponentResource {
     }, { parent: this /* protect: true */ });
 
     // IAM role for RDS Enhanced Monitoring
-    const enhancedMonitoringRole = new aws.iam.Role(`${serviceName}-rds-monitoring-role`, {
+    const enhancedMonitoringRole = new aws.iam.Role(`${serviceName}-rds-mon`, {
       assumeRolePolicy: JSON.stringify({
         Version: '2012-10-17',
         Statement: [{
@@ -149,7 +149,7 @@ export class IndexerShared extends pulumi.ComponentResource {
       }),
     }, { parent: this });
 
-    new aws.iam.RolePolicyAttachment(`${serviceName}-rds-monitoring-policy`, {
+    new aws.iam.RolePolicyAttachment(`${serviceName}-rds-mon-pol`, {
       role: enhancedMonitoringRole.name,
       policyArn: 'arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole',
     }, { parent: this });


crates/boundless-market/src/client.rs
```diff
@@ -119,14 +119,15 @@ pub struct ClientBuilder<St = NotProvided, Si = NotProvided> {
...
-impl<St, Si> Default for ClientBuilder<St, Si> {
+impl<U, D, S> Default for ClientBuilder<U, D, S> {
     fn default() -> Self {
         Self {
             deployment: None,
             rpc_url: None,
             rpc_urls: Vec::new(),
             signer: None,
             storage_provider: None,
+            downloader: None,
```
willpote, line 139: I think we should just rename `storage_provider` to `downloader` at this point. No need to worry about breaking changes, and seems much clearer

crates/boundless-market/src/storage/providers/s3.rs
```diff
@@ -0,0 +1,387 @@
...
+                return Err(StorageError::SizeLimitExceeded { size: buffer.len(), limit });
+            }
+        }
+
+        Ok(buffer)
+    }
+}
+
+/*
+#[cfg(test)]
```
willpote: Plan for this?

crates/boundless-market/src/storage/default.rs
```diff
@@ -0,0 +1,178 @@
...
+
+/// A downloader that can fetch data from various URL schemes.
+///
+/// This struct dispatches to the appropriate backend based on URL scheme:
+/// - `http://`, `https://` → HTTP downloader
+/// - `file://` → File downloader
+/// - `s3://` → S3 downloader (requires `s3` feature)
+/// - `gs://` → GCS downloader (requires `gcs` feature)
+#[derive(Clone, Debug)]
+pub struct DefaultDownloader {
```
willpote: Maybe StandardDownloader to match StandardUploader?

crates/boundless-market/src/storage/traits.rs
```diff
@@ -0,0 +1,80 @@
...
+use super::StorageError;
+use async_trait::async_trait;
+use auto_impl::auto_impl;
+use sha2::{Digest as _, Sha256};
+use url::Url;
+
+/// A trait for uploading risc0-zkvm programs and input files to a storage provider.
+#[async_trait]
+#[auto_impl(Arc)]
+pub trait StorageUploader: Send + Sync {
```
willpote, line 26: So to do encrypted inputs, it seems we will recommend:
1/  create a dedicated StorageUploader initialized with no_presigned just for uploading inputs
2/ use it to manually upload the inputs
3/ set the input in the client request builder using `.with_input_url`, 
4/ use the Client's default uploader for uploading program + submitting request

One alternative would be for the StorageUploader to be an object like 
```
{ program_uploader: Uploader, inputs_uploader: Uploader }
```

Then to use encrypted inputs you set a custom uploader more like

```Client::new().with_uploader(uploader)```

where uploader is: 

```
{ program_uploader: S3Uploader::default(), inputs_uploader: S3Uploader::from_config(s3_use_presigned=false, client=) }
```

Advantage would be to keep more of the logic inside the SDK, and make it a bit more explicit what is being used for program upload vs input upload

Lmk what you think

**Review by willpote — COMMENTED** [review]
Overall looks great, mostly naming/interface questions.

One question I have is how can we test this? To start with just having a test for the S3 encrypted inputs flow would be good. I assume to do this we will need some real AWS creds in CI, which I'm not against if we scope them down. Fine to merge this without that test, but we should add a follow up task

We definitely need to update the docs as well

File: crates/boundless-market/src/storage/providers/s3.rs (hunk 2 above)
Comment (Wollac) [reply crates/boundless-market/src/storage/providers/s3.rs]: Yeah, it is very unfortunate they we have abused #[ignore] to mark tests that require proof generation. And thus, all ignored tests will be run in the CI breaking these particular tests.
I only see two options:
a) remove the story tests that require auth
b) introduce a cfg option for all currently ignored tests, so that we could enable only those when the CI runs with dev mode.

File: crates/boundless-market/src/client.rs (hunk 1 above)
Comment (Wollac) [reply crates/boundless-market/src/client.rs:139]: There are two things now an. uploader as well as a downloader. And there could be situations where lets say the program is already on GCS and the inputs are supposed to go on S3 so must be different

File: crates/boundless-market/src/storage/traits.rs (hunk 4 above)
Comment (Wollac) [reply crates/boundless-market/src/storage/traits.rs:26]: Yes, having heterogeneous storage for the program and the input is tricky. In theory, the program could be on IPFS, and the input could be on S3.
I thought you would upload the program manually using the AWS CLI, web interface, etc., and then use the with_program_url function, since you probably have multiple inputs for a single program.
If this is a use case that we want to support, we'd have to add uploaders to the client and make them separately configurable. This would however add code and API complexity, unsure whether we want to support this usecase.

File: crates/boundless-market/src/client.rs (hunk 1 above)
Comment (willpote) [reply crates/boundless-market/src/client.rs:139]: I mean still keep the two things, just rename the storage_provider to uploader, so now we have two fields in ClientBuilder, one named uploader and one downloader

File: crates/boundless-market/src/storage/traits.rs (hunk 4 above)
Comment (willpote) [reply crates/boundless-market/src/storage/traits.rs:26]: We discussed offline and we can drop this request

File: crates/boundless-market/src/storage/providers/s3.rs (hunk 2 above)
Comment (willpote) [reply crates/boundless-market/src/storage/providers/s3.rs]: b) sounds good, but out of scope for this PR. Will add a task

File: crates/boundless-market/src/client.rs [hunk 5]
```diff
@@ -484,18 +515,39 @@ impl<St, Si> ClientBuilder<St, Si> {
 
     /// Set the storage provider.
     ///
-    /// The returned [ClientBuilder] will be generic over the provider [StorageProvider] type.
-    pub fn with_storage_provider<Z: StorageProvider>(
+    /// The returned [ClientBuilder] will be generic over the provider [StorageUploader] type.
+    pub fn with_storage_provider<Z: StorageUploader>(
```
willpote: @Wollac I meant that I think we should rename it in the client interface also, so you do with_uploader(..) etc.

Its totally fine to make a big breaking change here

crates/boundless-market/src/client.rs
```diff
@@ -93,12 +92,13 @@ pub enum FundingMode {
...
 #[derive(Clone)]
-pub struct ClientBuilder<St = NotProvided, Si = NotProvided> {
+pub struct ClientBuilder<U, D, S> {
     deployment: Option<Deployment>,
     rpc_url: Option<Url>,
     rpc_urls: Vec<Url>,
-    signer: Option<Si>,
-    storage_provider: Option<St>,
+    signer: Option<S>,
+    storage_provider: Option<U>,
```
willpote: Rename to `uploader` here too

crates/boundless-market/src/storage/default.rs
Wollac: done: 233119b7aae98233844cfdd4825a174cebde2473

crates/boundless-market/src/storage/providers/s3.rs
Wollac: done: b95d78fbb28be15ac64228456bac98cbe7d00ce4

crates/boundless-market/src/client.rs
Wollac: We now consistently use `downloader` and `uploader` for all the fields etc. The types are still called `StorageUploader` and `StorageDownloader` which I think is fine.

Wollac:
The only change worth documenting is that S3 and GCS authentication can now be configured in any way the toolchain supports, such as via environment, config, or IAM role. Previously, it was inconsistent, and certain things required boundless-specific ENVs. These boundless-specific variables are no longer supported.

koxu1996:
I couldn't find any information about Boundless following SemVer, but this PR is clearly breaking change (storage API modified). Got surprised with failing build after `cargo update`.

austinabell:
> I couldn't find any information about Boundless following SemVer, but this PR is clearly breaking change (storage API modified). Got surprised with failing build after `cargo update`.

Sorry about that, unfortunately we have needed to make some breaking changes, but aim to support semver for patch releases, so would recommend using tilde version constraints with the minor version you use (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#tilde-requirements) for now until we support semver more strictly.

PR #1499, opened 2026-01-12
BM-2239: add submit and submit_request

capossele:
Automatically uses offchain submission via the order stream service if available, otherwise falls back to onchain submission.

### DIFF
--- a/crates/boundless-market/src/client.rs
+++ b/crates/boundless-market/src/client.rs
@@ -1027,17 +1027,6 @@ where
         self.submit_request_with_signer(request, signer).await
     }
 
-    /// Submit a proof request onchain with a signature.
-    ///
-    /// Requires a signature to be provided to sign the request.
-    pub async fn submit_request_with_signature(
-        &self,
-        request: &ProofRequest,
-        signature: impl Into<Bytes>,
-    ) -> Result<(U256, u64), ClientError> {
-        self.submit_request_onchain_with_signature(request, signature).await
-    }
-
     /// Submit a proof request with a provided signer.
     ///
     /// Automatically uses offchain submission via the order stream service if available,
--- a/examples/smart-contract-requestor/apps/src/main.rs
+++ b/examples/smart-contract-requestor/apps/src/main.rs
@@ -114,7 +114,7 @@ async fn run(args: Args) -> Result<()> {
     let request = client.build_request(request).await?;
     let signature: Bytes = request.abi_encode().into();
     let (request_id, expires_at) =
-        client.submit_request_with_signature(&request, signature).await?;
+        client.submit_request_onchain_with_signature(&request, signature).await?;
     tracing::info!("Request {:x} submitted", request_id);
 
     // Wait for the request to be fulfilled by the market. The market will return the fulfillment data and seal.


crates/boundless-market/src/client.rs
```diff
@@ -993,6 +993,104 @@ where
...
+    /// Submit a proof request onchain with a signature.
+    ///
+    /// Requires a signature to be provided to sign the request.
+    pub async fn submit_request_with_signature(
+        &self,
+        request: &ProofRequest,
+        signature: impl Into<Bytes>,
+    ) -> Result<(U256, u64), ClientError> {
+        self.submit_request_onchain_with_signature(request, signature).await
+    }
```
austinabell: I'm not sure I understand the use of this new method, if not submitting off-chain first?

crates/boundless-market/src/client.rs
```diff
@@ -993,6 +993,104 @@ where
...
+                    );
+                    // Fall through to onchain submission
+                }
+            }
+        }
+
+        // Fallback to onchain submission (or use directly if offchain_client is None)
+        let request_id =
+            self.boundless_market.submit_request_with_value(&request, signer, value).await?;
+        Ok((request_id, request.expires_at()))
```
austinabell, line 1080: Is the reason that the submit logic is duplicated to avoid warning twice for the same error if not failing specifically on submit?

At first glance it seemed it might be because of the deposit, but seems that isn't deduplicated (if only submit fails and deposit succeeds, this deposits twice)

willpote (approved):
Nice, we should update docs also

crates/boundless-market/src/client.rs
capossele: I added to keep consistency with the "on_chain" methods. This in particular is only supported in the on_chain case though. Can be removed if we don't like it

crates/boundless-market/src/client.rs
capossele: The main reason was to avoid depositing twice yeah. I can't see though why this should deposit twice

austinabell: oh oops when I skimmed I missed that you are setting value to zero after the deposit, ignore me :)

crates/boundless-market/src/client.rs
austinabell: yeah I guess what I'm asking is why not have submitting with signature supported for off-chain when there is `..offchain_with_signer`?

As is, seems to me to be a bit unintuitive between the two methods `submit_request_with_signer` and `submit_request_with_signature` that the latter only submits on-chain where signer tries both?

capossele: It's a bit tricky because the order stream currently does not support EIP1271 signatures so we only implemented that method for the onchain part. But I understand as is it creates confusion so I'll drop the method from the generalized one and only leave the explicit one. We can revisit once we add support for that on the order stream

capossele: https://github.com/boundless-xyz/boundless/pull/1499/commits/61733e1b470f31a9c2fa707d544477ca4169c771

PR #1500, opened 2026-01-13
BM-2247: add warnings

capossele:
Warn if 

- Order requires less than 4 Khz and collateral requirement < 20 ZKC
- Orders requires greater than 4Khz , and less than 10Khz, and has collateral < 50 ZKC
- Order requires greater than 10Khz and has collateral < 100 ZKC

“Warning: the collateral requirement of your request is low. This means the incentives for secondary provers to fulfill the order if the primary provers is slashed may be too low. Consider raising your lock collateral to Y.”

Warn if timeouts require > 10 Mhz to be fulfilled

“Warning: your request requires a proving Khz of X to be fulfilled before timeout. This limits the number of provers in the network that will be able to fulfill your order. Consider setting a longer timeout of Y.”


capossele:
Superseeded by https://github.com/boundless-xyz/boundless/pull/1505

PR #1502, opened 2026-01-13
BM-2249: Add Parameterization Mode

capossele:
Set rampUpPeriod, lockTimeout, timeout based on a configurable “mode”. Two options: FULFILLMENT/LATENCY

- Fulfillment: assume 750Khz of proving power + 30 Mhz execution power
- Latency: assume 3000kHz of proving power + 50 Mhz execution power

Include documentation that “latency” can result in fewer provers taking your requests, and thus higher prices / lower fulfillment guarantees.

### DIFF
--- a/crates/boundless-market/src/request_builder/mod.rs
+++ b/crates/boundless-market/src/request_builder/mod.rs
@@ -725,8 +725,8 @@ impl ParameterizationMode {
     ///
     /// Sets the ramp up period as 10x the executor time assuming the executor speed is 30 MHz.
     /// Sets the lock timeout as the sum of the ramp up period and the proving and executor times
-    /// assuming the proving speed is 500 kHz and the executor speed is 30 MHz.
-    /// Sets the timeout as 2 times the lock timeout.
+    /// assuming the proving speed is 500 kHz and the executor speed is 30 MHz and capped at 4 hours.
+    /// Sets the timeout as 2 times the lock timeout and capped at 8 hours.
     pub fn fulfillment() -> Self {
         Self {
             proving_speed: Self::DEFAULT_PROVING_SPEED_HZ,
@@ -745,8 +745,8 @@ impl ParameterizationMode {
     ///
     /// Sets the ramp up period as 5x the executor time assuming the executor speed is 50 MHz.
     /// Sets the lock timeout as the sum of the ramp up period and the proving and executor times
-    /// assuming the proving speed is 3 MHz and the executor speed is 50 MHz.
-    /// Sets the timeout as 2 times the lock timeout.
+    /// assuming the proving speed is 3 MHz and the executor speed is 50 MHz and capped at 4 hours.
+    /// Sets the timeout as 2 times the lock timeout and capped at 8 hours.
     pub fn latency() -> Self {
         Self {
             proving_speed: Self::FAST_PROVING_SPEED_HZ,
--- a/crates/boundless-market/src/request_builder/offer_layer.rs
+++ b/crates/boundless-market/src/request_builder/offer_layer.rs
@@ -35,6 +35,7 @@ use derive_builder::Builder;
 
 pub(crate) const DEFAULT_TIMEOUT: u32 = 600;
 pub(crate) const DEFAULT_RAMP_UP_PERIOD: u32 = 60;
+pub(crate) const DEFAULT_MAX_TIMEOUT: u32 = 14400; // 4 hours
 
 struct CollateralRecommendation {
     default: U256,
@@ -539,10 +540,15 @@ where
                     .config
                     .ramp_up_period
                     .unwrap_or(parameterization_mode.recommended_ramp_up_period(cycle_count));
-                let lock_timeout = self.config.lock_timeout.unwrap_or(
-                    parameterization_mode.recommended_timeout(cycle_count) + ramp_up_period,
+                let recommended_timeout =
+                    parameterization_mode.recommended_timeout(cycle_count) + ramp_up_period;
+                let lock_timeout = self
+                    .config
+                    .lock_timeout
+                    .unwrap_or(recommended_timeout.min(DEFAULT_MAX_TIMEOUT));
+                let timeout = self.config.timeout.unwrap_or(
+                    (recommended_timeout * 2 - ramp_up_period).min(DEFAULT_MAX_TIMEOUT * 2),
                 );
-                let timeout = self.config.timeout.unwrap_or(lock_timeout * 2 - ramp_up_period);
                 (lock_timeout, timeout, ramp_up_period, ramp_up_start)
             } else {
                 let lock_timeout = DEFAULT_TIMEOUT + DEFAULT_RAMP_UP_PERIOD;


crates/boundless-market/src/request_builder/mod.rs
```diff
@@ -635,6 +636,292 @@ impl MissingFieldError {
...
+///
+/// Defines the proving and executor speeds used to calculate recommended timeouts.
+/// Note: setting faster speeds may result in fewer provers being able to fulfill the request,
+/// higher prices, and lower fulfillment guarantees.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct DeliverySpeed {
+    /// Proving speed in kHz.
+    pub proving_speed: f64,
+    /// Executor speed in kHz.
+    pub executor_speed: f64,
```
austinabell: These are both used (and casted as u64). Seems much safer for these to never be floats? Seems unlikely one would want to configure <1khz precision?

crates/boundless-market/src/request_builder/mod.rs
```diff
@@ -635,6 +636,292 @@ impl MissingFieldError {
...
+    ///
+    /// This is used to ensure that the ramp up period is long enough
+    /// to allow provers to execute and evaluate the request.
+    pub const DEFAULT_RAMP_UP_PERIOD_MULTIPLIER: u32 = 10; // 10x the executor time
+
+    /// Fast multiplier for the ramp up period.
+    ///
+    /// This is used to ensure that the ramp up period is long enough
+    /// to allow provers to execute and evaluate the request.
+    pub const FAST_RAMP_UP_PERIOD_MULTIPLIER: u32 = 5; // 5x the executor time
```
austinabell: This is a large API surface to maintain without breaking changes, specifically these consts. Just want to confirm that you don't forsee anything changing with these

crates/boundless-market/src/request_builder/mod.rs
```diff
@@ -635,6 +636,292 @@ impl MissingFieldError {
...
+    /// Executor speed in kHz.
+    pub executor_speed: f64,
+    /// Minimum timeout in seconds.
+    pub min_timeout: u32,
+    /// Ramp up period multiplier.
+    pub ramp_up_period_multiplier: u32,
+}
+
+impl DeliverySpeed {
+    /// Default proving speed in kHz.
```
austinabell: Just a comment that khz/mhz conversions have been a bit annoying, and was hoping we would move away from them. Especially since the const or field names doesn't indicate KHZ, perhaps this should just be HZ?

crates/boundless-market/src/client.rs
```diff
@@ -395,6 +396,28 @@ impl<St, Si> ClientBuilder<St, Si> {
...
+    /// # use boundless_market::Client;
+    /// use boundless_market::DeliverySpeed;
+    ///
+    /// Client::builder().with_delivery_speed(DeliverySpeed::fast());
+    /// ```
+    ///
+    pub fn with_delivery_speed(self, delivery_speed: DeliverySpeed) -> Self {
+        self.config_offer_layer(|config| config.delivery_speed(delivery_speed))
+    }
+
```
willpote: As written I think this means all OfferLayers (e.g. custom ones people write) need to support this concept of DeliverySpeed (or least have it as part of their params)

I'm not actually sure if this is possible in the current framework, but I was thinking this should just be part of the default OfferLayer we provide only. And other offer layers don't need to think about it.

Maybe only way to do this is to have two separate OfferLayer, one OfferLayerStandard and OfferLayerFast, or something like that?

We could still provide a convenience function at the Client layer like `with_default_offer_layer_fast_mode`.

crates/boundless-market/src/client.rs
```diff
@@ -395,6 +396,28 @@ impl<St, Si> ClientBuilder<St, Si> {
         Self { funding_mode, ..self }
     }
 
+    /// Set the delivery speed for the offer layer.
```
willpote: Thinking about naming wise, I think we should probably steer away from focusing it too much on delivery speed, as I think the reliability tradeoffs are equally important. 

Maybe we do something like: 

* Instead of Delivery Speed call it Parameterization Mode
* FULFILLMENT / LATENCY as the two names

crates/boundless-market/src/request_builder/mod.rs
willpote: I don't see an issue with making breaking changes here, as these are just our recommended defaults intended to get you sending requests that get fulfilled on the market- more important is that we are conservative with the numbers here and ensure the defaults don't cause requests to fail/expire.

I actually fully expect us to make breaking changes 
1/ as the market evolves, and 2/ as we come up with better methodologies (these hardcoded values can surely be improved upon, but are already a huge improvement on what we have today)

austinabell: I meant more in the sense of semver breaking changes from a release perspective, the API surface being a bit limiting. Very minor point, just noting that if changed/improved as you mention, then it would require major version bump to not be semver breaking. Seems like we haven't been very strict about this so far, but just commenting on it anyway

capossele: my bad, I should have not made those const public, they are supposed to stay internal

crates/boundless-market/src/client.rs
capossele: https://github.com/boundless-xyz/boundless/pull/1502/commits/89d252909eef1f546b564bd1738f29e062b19eda

crates/boundless-market/src/request_builder/mod.rs
capossele: fixed in https://github.com/boundless-xyz/boundless/pull/1502/commits/89d252909eef1f546b564bd1738f29e062b19eda

crates/boundless-market/src/client.rs
capossele: Typically though we do not implement more offerLayers, rather we implement different request builders. So I've made the new param mode an option, and it defaults to none, so we need to be explicit to select the mode

crates/boundless-market/src/request_builder/mod.rs
```diff
@@ -635,6 +636,316 @@ impl MissingFieldError {
...
+    ///
+    /// This is to prevent the timeout from being too short and causing the request
+    /// to expire before the prover can execute and evaluate the request.
+    const FAST_MIN_TIMEOUT: u32 = 30;
+
+    /// Default multiplier for the ramp up period.
+    ///
+    /// This is used to ensure that the ramp up period is long enough
+    /// to allow provers to execute and evaluate the request.
+    const DEFAULT_RAMP_UP_PERIOD_MULTIPLIER: u32 = 10; // 10x the executor time
```
capossele, line 696: I'm still pondering if these values as multiplier make sense

crates/boundless-market/src/request_builder/offer_layer.rs
```diff
@@ -346,15 +370,38 @@ where
             .bidding_start
             .unwrap_or_else(|| now_timestamp() + self.config.bidding_start_delay);
 
-        Ok(Offer {
+        let (lock_timeout, timeout, ramp_up_period) =
+            if let Some(parameterization_mode) = self.config.parameterization_mode {
+                (
+                    self.config
+                        .lock_timeout
+                        .unwrap_or_else(|| parameterization_mode.recommended_timeout(cycle_count)),
```
capossele: Also here we could increase the timeout by adding also the rampUpPeriod

crates/boundless-market/src/request_builder/offer_layer.rs
```diff
@@ -346,15 +370,39 @@ where
...
+                let lock_timeout = self.config.lock_timeout.unwrap_or(
+                    parameterization_mode.recommended_timeout(cycle_count) + ramp_up_period,
+                );
+                let timeout = self
+                    .config
+                    .timeout
+                    .unwrap_or(parameterization_mode.recommended_timeout(cycle_count) * 2);
+                (lock_timeout, timeout, ramp_up_period)
+            } else {
+                (
```
willpote: Can we log a warning in the case where we end up setting the hardcoded 600/1200/60, something like "no parameterization mode configured, using defaults of x y z"

examples/counter/apps/src/main.rs
```diff
@@ -90,6 +90,9 @@ async fn run(args: Args) -> Result<()> {
         .with_deployment(args.deployment)
         .with_storage_provider_config(&args.storage_config)?
         .with_private_key(args.private_key)
+        // Use ParameterizationMode::fulfillment() for more conservative offering parameters.
+        // Or use ParameterizationMode::latency() for lower latency but higher price and lower fulfillment guarantees.
+        .with_parameterization_mode(ParameterizationMode::fulfillment())
```
capossele: This is not really needed anymore since it's on by default. I'll remove it

willpote (approved):

PR #1505, opened 2026-01-13
BM-2252: Lock collateral settings

capossele:
Add warnings and recommended values based on chain id


crates/boundless-market/src/request_builder/offer_layer.rs
```diff
@@ -30,6 +30,94 @@ use anyhow::{ensure, Context};
...
+/// Returns `true` if a warning was logged.
+fn check_primary_performance_warning(cycle_count: u64, primary_performance: f64) -> bool {
+    if primary_performance > XL_REQUESTOR_LIST_THRESHOLD_KHZ {
+        let recommended_lock_timeout =
+            cycle_count.div_ceil(XL_REQUESTOR_LIST_THRESHOLD_KHZ as u64) as u32;
+        tracing::warn!(
+            "Warning: your request requires a proving Khz of {primary_performance} to be \
+             fulfilled before the lock timeout. This limits the number of provers in the \
+             network that will be able to fulfill your order. Consider setting a longer \
+             lock timeout of at most {recommended_lock_timeout} seconds."
```
austinabell:
```suggestion
             lock timeout of at least {recommended_lock_timeout} seconds."
```

crates/boundless-market/src/request_builder/offer_layer.rs
```diff
@@ -30,6 +30,94 @@ use anyhow::{ensure, Context};
...
+            && lock_collateral < self.large
+        {
+            Some(self.large)
+        } else if secondary_performance >= XL_REQUESTOR_LIST_THRESHOLD_KHZ
+            && lock_collateral < self.xl
+        {
+            Some(self.xl)
+        } else {
+            None
+        };
```
austinabell, line 69: These seem like they should be based on the cycle count? Fixed collateral for dynamically sized orders seems weird. Maybe we want it dynamic with a min value being the ones we recommend?

willpote: Dynamic could make sense, but we need to make sure that we don't by default set collateral that is so high that everyone in the market will skip the order.

I don't think collateral needs to be super accurate so I'm fine with static as a first iteration- it just needs to be "high" such that the incentive for secondary proving is there. The footgun we're trying to fix here is customers setting collateral of like 1 ZKC, which has happened a few times. I'm fine with either though

crates/boundless-market/src/request_builder/offer_layer.rs
```diff
@@ -30,6 +30,94 @@ use anyhow::{ensure, Context};
 use clap::Args;
 use derive_builder::Builder;
 
+// Peak performance thresholds for enabling requestor lists (kHz)
+const LARGE_REQUESTOR_LIST_THRESHOLD_KHZ: f64 = 4000.0;
+const XL_REQUESTOR_LIST_THRESHOLD_KHZ: f64 = 10000.0;
+
```
willpote: Can we use the existing consts from the wizard to keep in sync?

capossele: https://github.com/boundless-xyz/boundless/pull/1505/commits/aed3e6479eac7011323dd5022b24674600affdaa move them under the boundless market crate

PR #1508, opened 2026-01-14
BM-2256: Zeroecco/84532 support

zeroecco:
add support for sepolia
fix bug in shell vs bash
Docker should not bypass iptables
2 concurrent proofs is better
with dynamic gpu allocation, memory and cpu should not locked
clean up scripts need to be removed

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -52,7 +52,7 @@ x-exec-agent-common: &exec-agent-common
     [
       "/bin/sh",
       "-c",
-      "/app/agent -t exec --segment-po2 $SEGMENT_SIZE --redis-ttl $REDIS_TTL",
+      "/app/agent -t exec --segment-po2 $$SEGMENT_SIZE --redis-ttl $$REDIS_TTL",
     ]
 
 x-broker-environment: &broker-environment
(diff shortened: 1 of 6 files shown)


compose.yml
```diff
@@ -219,10 +217,6 @@ services:
       <<: *base-environment
       # Default to 4 max client connections if not set (work, poll for requeue, completed cleanup, stuck tasks cleanup)
       DB_MAX_CONNECTIONS: ${DB_MAX_CONNECTIONS:-4}
-      CLEANUP_POLL_INTERVAL: ${CLEANUP_POLL_INTERVAL:-}
-      STUCK_TASKS_POLL_INTERVAL: ${STUCK_TASKS_POLL_INTERVAL:-}
-      BENTO_DISABLE_COMPLETED_CLEANUP: ${BENTO_DISABLE_COMPLETED_CLEANUP:-}
-      BENTO_DISABLE_STUCK_TASK_CLEANUP: ${BENTO_DISABLE_STUCK_TASK_CLEANUP:-}
```
austinabell: why delete these env var forwardings? This would be breaking for some setups

zeroecco: this breaks deployments because aux bombs out when considering. the feature is also considered unsafe. removing it was the option I went with

austinabell (approved):

PR #1509, opened 2026-01-14
BM-2152: BM2152: Support on-chain querying of old proof requests 

jonastheis:
This PR enables the ability to query old proof requests/fulfillments by adding configurable block range parameters across all CLI commands that search for events. 

Specifically it introduces following changes:
- improve error message for users when a on-chain query fails
- refactor and deduplicate backward querying logic
- introduce parameters in CLI commands

**Updated commands:** 
1. boundless prover lock                                                                                                                             
2. boundless prover fulfill                                                                                                                          
3. boundless prover execute                                                                                                                          
4. boundless prover benchmark                                                                                                                        
5. boundless requestor status                                                                                                                        
6. boundless requestor get-proof
7. boundless requestor verify-proof
 
**Options added:**
```bash
--search-to-block <SEARCH_TO_BLOCK>
    Lower bound: search events backwards down to this block

--search-from-block <SEARCH_FROM_BLOCK>
    Upper bound: search events backwards from this block (defaults to latest). Set this for old requests to reduce RPC calls and cost

```

### Known issues
- `boundless requestor verify-proof` fetches fulfillment (`get_request_fulfillment`) and request (`get_submitted_request`) data. If request and fulfillment transaction are too far apart specifying `--search-from-block` alone might not work and the entire block range for both needs to be specified. Here it would be better to introduce `--search-from-block` for each of the requests within the command. 

### Further work
I think it would be valuable to utilize the indexer more to make the life of SDK users easier and cheaper. We could nicely abstract this away by having multiple clients (indexer, order stream, on-chain) implementing the same interface. It should always be possible to opt out of centralized components usage but if used these could be queried first with the on-chain query as a fallback. 

### How to test
We can test the functionality manually by querying a very old order (3 months): https://explorer.boundless.network/orders/0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2

```bash
# improved error message for user
$ ./target/debug/boundless prover execute --request-id 0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2 

Executing Proof Request [Base Mainnet]
  Request ID:      0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2
  Status:          Fetching request from blockchain
Error: Market error Request not found for request 0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2 in event logs after searching from block 40880900 to block 40880811. Consider increasing EventQueryConfig.block_range or EventQueryConfig.max_iterations to increase search depth or specify start/end blocks

# specifying `--search-from-block`
$ ./target/debug/boundless prover execute --request-id 0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2 --search-from-block 36990812 

Executing Proof Request [Base Mainnet]
  Request ID:      0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2
  Status:          Fetching request from blockchain
  Status:          Starting execution
  Status:          Fetching program
[...]
```

```bash
# improved error message for user
$ ./target/debug/boundless requestor get-proof 0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2

Fetching Proof [Base Mainnet]
  Request ID:      0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2
Error: Failed to retrieve proof

Caused by:
    Proof not found for request 0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2 in events logs after searching from block 40880824 to block 40880735. Consider increasing EventQueryConfig.block_range or EventQueryConfig.max_iterations to increase search depth or specify start/end blocks

# specifying `--search-from-block`
$ ./target/debug/boundless requestor get-proof 0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2 --search-from-block 36991649

Fetching Proof [Base Mainnet]
  Request ID:      0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2

✓ Successfully retrieved proof
```

```bash
# improved error message for user
$ ./target/debug/boundless requestor verify-proof 0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2 0x22a074286e1409a043a02a1e9e1064c13ab853d296da7abdca78a8180b933c7e                            

Verifying Proof [Base Mainnet]
  Request ID:      0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2
Error: Proof not found for request 0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2 in events logs after searching from block 40881032 to block 40880943. Consider increasing EventQueryConfig.block_range or EventQueryConfig.max_iterations to increase search depth or specify start/end blocks

# specifying `--search-from-block`
$ ./target/debug/boundless requestor verify-proof 0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2 0x22a074286e1409a043a02a1e9e1064c13ab853d296da7abdca78a8180b933c7e --search-from-block 36991649

Verifying Proof [Base Mainnet]
  Request ID:      0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2

✓ Successfully verified proof for request 0xb686f080c2c7045d13d196787c99bc9972480fd32a4170c2
```

### DIFF
--- a/crates/boundless-market/src/client.rs
+++ b/crates/boundless-market/src/client.rs
@@ -1249,6 +1249,24 @@ where
 
     /// Get the [SetInclusionReceipt] for a request.
     ///
+    /// This method fetches the fulfillment data for a request and constructs the set inclusion receipt.
+    ///
+    /// # Parameters
+    ///
+    /// * `request_id` - The unique identifier of the proof request
+    /// * `image_id` - The image ID for the receipt claim
+    /// * `search_to_block` - Optional lower bound for the block search range. The search will go backwards
+    ///   down to this block number. Combined with `search_from_block` to define a specific range.
+    /// * `search_from_block` - Optional upper bound for the block search range. The search starts backwards
+    ///   from this block. Defaults to the latest block if not specified. Set this to a block number near
+    ///   when the request was fulfilled to reduce RPC calls and cost when querying old fulfillments.
+    ///
+    /// # Default Search Behavior
+    ///
+    /// Without explicit block bounds, the onchain search covers blocks according to
+    /// EventQueryConfig.block_range and EventQueryConfig.max_iterations.
+    /// Fulfillment events older than this default range will not be found unless you provide explicit `search_to_block` and `search_from_block` parameters.
+    ///
     /// # Examples
     ///
     /// ```rust
@@ -1260,7 +1278,23 @@ where
     ///
     /// async fn fetch_set_inclusion_receipt(request_id: U256, image_id: B256) -> Result<(Bytes, SetInclusionReceipt<ReceiptClaim>)> {
     ///     let client = ClientBuilder::new().build().await?;
-    ///     let (journal, receipt) = client.fetch_set_inclusion_receipt(request_id, image_id).await?;
+    ///
+    ///     // For recent requests
+    ///     let (journal, receipt) = client.fetch_set_inclusion_receipt(
+    ///         request_id,
+    ///         image_id,
+    ///         None,
+    ///         None,
+    ///     ).await?;
+    ///
+    ///     // For old requests with explicit block range
+    ///     let (journal, receipt) = client.fetch_set_inclusion_receipt(
+    ///         request_id,
+    ///         image_id,
+    ///         Some(1000000),  // search_to_block
+    ///         Some(1500000),  // search_from_block
+    ///     ).await?;
+    ///
     ///     Ok((journal, receipt))
     /// }
     /// ```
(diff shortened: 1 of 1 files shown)


crates/boundless-market/src/contracts/boundless_market.rs
```diff
@@ -1697,6 +1756,8 @@ impl<P: Provider> BoundlessMarketService<P> {
         &self,
         request_id: U256,
         tx_hash: Option<B256>,
+        lower_bound: Option<u64>,
+        upper_bound: Option<u64>,
```
capossele, line 1627: Just an idea, since we write most of the time none for all the 3 of these, maybe we could create a Config structure that defaults to none for all of them and then you can use something like .with_tx_hash(tx_hash) to add that and so on?

jonastheis: I get your point but I'm not sure if it's worth it as this function has only 4 call sites. Might just increase complexity unnecessarily.

capossele (approved):
Awesome!

crates/boundless-market/src/client.rs
```diff
@@ -1183,7 +1183,8 @@ where
     ) -> Result<(Bytes, SetInclusionReceipt<ReceiptClaim>), ClientError> {
         // TODO(#646): This logic is only correct under the assumption there is a single set
         // verifier.
-        let fulfillment = self.boundless_market.get_request_fulfillment(request_id).await?;
+        let fulfillment =
+            self.boundless_market.get_request_fulfillment(request_id, None, None).await?;
```
willpote: Would be worth updating this one too while we're at it, seems as is theres no way to get set inclusion receipts older than x days

crates/boundless-market/src/client.rs
```diff
@@ -1218,6 +1219,8 @@ where
         request_id: U256,
         tx_hash: Option<B256>,
         request_digest: Option<B256>,
+        lower_bound: Option<u64>,
+        upper_bound: Option<u64>,
```
willpote: This is the function customers will be calling so its the most important one to be clearly documented/named. 

Perhaps `search_from_block` `search_to_block`? 

And would be good to update the comment to explain why it may be important to set this value, and if you don't set it what the default search range is

PR #1531, opened 2026-01-26
BM-2288: Improve request builder params

capossele:
- increase timeout
- decrease default exec speed
- default to market pricing if no indexer is provided
- add indexer urls to order generator infra
- use FundingMode:BelowThreshold on order generators


capossele:
Fixes BM-2246

PR #1535, opened 2026-01-26
BM-2292: Add NO_COLOR environment variable to compose.yml

zeroecco:
improve logging for users who use log aggregations


compose.yml
```diff
@@ -14,6 +14,7 @@ x-base-environment: &base-environment
   S3_BUCKET: ${MINIO_BUCKET:-workflow}
   S3_ACCESS_KEY: ${MINIO_ROOT_USER:-admin}
   S3_SECRET_KEY: ${MINIO_ROOT_PASS:-password}
+  NO_COLOR: 1
```
austinabell, line 17: Should these not change the default? Seems like what you want is just to set the env var for when ansi colours are not handled well? Opinionated so not blocking this change

```suggestion
  NO_COLOR:
```

zeroecco: This pr is def opinionated for me. I hate colors in my logs weirdly enough

PR #1536, opened 2026-01-26
BM-2293: chore: add market fees

unizarr:
## Description

Added market fee data to the epoch endpoints:

- `/v1/povw/epochs/{epoch}`
- `/v1/povw/epochs`


crates/indexer/src/db/market.rs
```diff
@@ -3652,6 +3660,39 @@ impl IndexerDb for MarketDb {
...
+              AND max_price IS NOT NULL",
+        )
+        .bind(period_start as i64)
+        .bind(period_end as i64)
+        .fetch_one(&self.pool)
+        .await?;
+
+        let market_fee_str: String = row.try_get("market_fee")?;
+        let market_fee =
+            U256::from_str(market_fee_str.split('.').next().unwrap_or("0")).unwrap_or(U256::ZERO);
```
willpote: Are you sure we need to split on '.' here. Lock price and max price should be stored as U256 strings in the DB, so should just be integer arithmetic?

crates/indexer/src/db/market.rs
```diff
@@ -3652,6 +3660,39 @@ impl IndexerDb for MarketDb {
         Ok(total)
     }
 
+    async fn get_period_market_fees(
```
willpote: Lets add a test for this fn

crates/lambdas/indexer-api/src/routes/povw.rs
```diff
@@ -525,6 +548,29 @@ async fn get_povw_address_history_impl(
     // Fetch aggregate summary for this address
     let address_aggregate = state.rewards_db.get_povw_rewards_aggregate_by_address(address).await?;
 
+    // Get unique epochs and calculate market fees per epoch
```
willpote: It looks like you are adding overall market fees to the `/addresses/{address}` endpoint? 

See my comment below but I think we can remove all this

crates/lambdas/indexer-api/src/models.rs
```diff
@@ -344,6 +354,8 @@ pub struct PoVWAddressSummary {
     pub total_uncapped_rewards: String,
     pub total_uncapped_rewards_formatted: String,
     pub epochs_participated: u64,
+    pub market_fees: String,
+    pub market_fees_formatted: String,
```
willpote: Market fees is an overall market metric, not specific to a particular address, so I don't think it needs to be added to `PoVWAddressSummary`, or any of the `/address` endpoints. It should really just live on `/epochs` and `/epochs/:epoch.   Which will simplify the PR quite a bit

crates/lambdas/indexer-api/src/routes/povw.rs
```diff
@@ -146,30 +147,39 @@ async fn get_povw_all_epochs_summary_impl(
...
-        })
-        .collect();
+    let mut entries: Vec<EpochPoVWSummary> = Vec::with_capacity(summaries.len());
+
+    for summary in summaries {
+        let market_fees = state
+            .market_db
+            .get_period_market_fees(summary.epoch_start_time, summary.epoch_end_time)
+            .await
+            .unwrap_or_default();
```
willpote: You can parallelize this with `tokio::join` to make this quite a bit faster, though doing 60+ in parallel might be a bit much. Maybe we can start with doing chunks of 10 in parallel?

crates/indexer/src/db/market.rs
unizarr: yep we should remove it

PR #1540, opened 2026-01-28
BM-1842: Make ETH and ZKC dynamically priced by USD

jonastheis:
Addresses [BM-1842](https://linear.app/boundlessnetwork/issue/BM-1842/make-eth-and-zkc-dynamically-priced-by-usd) in 2 parts:
1. Adds a price oracle infrastructure to fetch ETH/USD and ZKC/USD prices from multiple sources with aggregation and caching.
2. Enables provers to configure pricing thresholds in USD, which are automatically converted to ETH/ZKC at runtime using live oracle prices.

- This is a backward compatible change, config values without currency symbol (`ETH`, `ZKC`, `USD`) will be interpreted with their default `Asset` based on the corresponding config property. 
- By [default](https://github.com/boundless-xyz/boundless/blob/dc28a0bbfeb43fd9ffb34fb78b028864835c231b/crates/boundless-market/src/price_oracle/config.rs#L104) the price oracle is enabled (median price of Coingecko(offchain) and Chainlink(onchain)), refreshing prices every 60s and logging an error message if any price could not be refreshed for more than 12h to alert the prover about potentially stale prices. A prover can then manually set a static price if desired. More context to this [here](https://github.com/boundless-xyz/boundless/pull/1540#issuecomment-3876798217).
- Default prices in `generate config` command and in the `broker-template.toml` as well as in the default implementation of the config have been set to be in USD. 

# Part 1: Price Oracle implementation
Adds a `price_oracle` module to `boundless-market` with a trait-based, composable architecture designed for reliability and testability. The crate provides extensive unit and integration tests.

### Architecture
```
PriceOracleManager (manages ETH/USD + ZKC/USD pairs)
  ├── eth_usd: CachedPriceOracle (background refresh)
  │     └── CompositeOracle (aggregation + fallback)
  │           ├── ChainlinkSource (on-chain)
  │           ├── CoinGeckoSource (off-chain)
  │           └── CoinMarketCapSource (off-chain)
  └── zkc_usd: CachedPriceOracle
        └── CompositeOracle
              ├── CoinGeckoSource (off-chain)
              └── CoinMarketCapSource (off-chain)
```

### Key Components
- **PriceOracle trait**: Async interface for price fetching
- **PriceSource trait**: Named price sources
- **CompositeOracle**: Aggregates multiple sources with configurable modes
- **CachedPriceOracle**: Non-blocking reads with background refresh
- **PriceOracleManager**: Top-level manager with background refresh task and asset conversion

### Aggregation Modes
- **Priority**: First successful source wins (sequential)
- **Median**: Median of all successful sources (parallel)
- **Average**: Average of all successful sources (parallel)

### Implemented Price Sources
| Source | Type | Pairs | Notes |
|--------|------|-------|-------|
| Chainlink | On-chain | ETH/USD | Uses AggregatorV3Interface |
| CoinGecko | Off-chain | ETH/USD, ZKC/USD | Free API, no key required |
| CoinMarketCap | Off-chain | ETH/USD, ZKC/USD | Requires API key |
| Static | Fallback | Any | Returns configured fixed price |

### Asset Conversion
`Amount` allows for monetary values to be passed together with their `Asset`. Through this we can accept inputs in different currencies and convert (using the price oracle) at the time of resolution when we need the target currency (usually ETH or ZKC).
- **Asset enum**: USD (6 decimals), ETH (18 decimals), ZKC (18 decimals)
- **Amount struct**: Value + asset with human-readable parsing (`"1.5 ETH"`, `"100 USD"`)
- **PriceOracleManager.convert()**: Convert amounts between USD/ETH/ZKC using live prices

### Testing
- **Unit tests** (no network): aggregation modes, caching, staleness, error handling
- **Integration tests** (`--ignored`): live Chainlink/CoinGecko/CMC calls

```bash
cargo test -p boundless-market price_oracle --lib
cargo test -p boundless-market price_oracle::integration_tests -- --ignored
```
# Part 2: USD-Based Pricing

### USD-Enabled Config Fields
| Field | Allowed Assets | Example | Purpose |
|-------|---------------|---------|---------|
| `min_mcycle_price` | USD, ETH | `"0.02 USD"` | Minimum price per mega-cycle for ETH-paid orders |
| `max_collateral` | USD, ZKC | `"100 USD"` | Maximum collateral to stake for any order |
| `min_mcycle_price_collateral_token` | USD, ZKC | `"0.001 USD"` | Minimum price per mega-cycle for collateral-paid orders |

### Configuration Examples
```toml
[market]
# USD-based (recommended for stable pricing)
min_mcycle_price = "0.02 USD"
max_collateral = "100 USD"
min_mcycle_price_collateral_token = "0.001 USD"

# Or native token pricing (existing behavior)
min_mcycle_price = "0.00001 ETH"
max_collateral = "200 ZKC"
min_mcycle_price_collateral_token = "0.00005 ZKC"
```

### `Amount` and `Asset` type
These types are central to the entire design of USD-based pricing. Instead of simply representing assets as U256, we represent them together with their `Asset`-type and `decimals`. This way we read values in any `Asset`-type and convert to the target value whenever we need it using the price oracle.

### Runtime Conversion
- `OrderPricingContext` trait provides `convert_to_eth()` and `convert_to_zkc()` methods
- Conversion happens at order pricing time using live prices from `PriceOracleManager`
- `scale_decimals()` handles decimal precision differences between Asset types (18 decimals) and contract tokens (6 or 18 depending on chain)

### Implementation Details
- **Amount parsing**: `Amount::parse_with_allowed()` validates that only permitted assets are used per field
- **Custom deserializers**: Each config field has a deserializer that validates allowed assets at config load time
- **Logging**: Debug logs show both configured amount (e.g., `"0.02 USD"`) and converted value for transparency

### Backwards Compatibility
- Existing configs using ETH/ZKC values continue to work unchanged
- The asset suffix is now required (e.g., `"0.00001 ETH"` instead of `"0.00001"`)

### DIFF
--- a/crates/boundless-cli/src/commands/prover/generate_config.rs
+++ b/crates/boundless-cli/src/commands/prover/generate_config.rs
@@ -502,7 +502,7 @@ impl ProverGenerateConfig {
             Ok(price_percentiles) => {
                 display.item_colored(
                     "Median price",
-                    &format!(
+                    format!(
                         "{} / Mcycle",
                         format_amount_with_conversion(
                             &format!(
@@ -521,7 +521,7 @@ impl ProverGenerateConfig {
                 );
                 display.item_colored(
                     "25th percentile",
-                    &format!(
+                    format!(
                         "{} / Mcycle",
                         format_amount_with_conversion(
                             &format!(


capossele:
Thinking out loud about what could brake after merging this; I think the nightly-prover might fail at parsing the broker.toml file as files like `ansible/roles/prover/configs/broker/8453.toml` ad similarly `nightly_8453.toml`, `prod_11155111.toml`, `staging_84532.toml`, `prod_84532.toml` use something like:

```
min_mcycle_price = "0.000000025" (no asset)
min_mcycle_price_collateral_token = "0" (no asset)
max_collateral = "200" (no asset)
```

Would that fail to parse?

crates/boundless-market/src/price_oracle/asset.rs
```diff
@@ -0,0 +1,567 @@
...
+/// Parse an amount string into a value and asset
+///
+/// Format: "\<value\> \<asset\>" (e.g., "1.5 ETH", "100 USD")
+/// Plain numbers without asset specification are NOT supported and will return an error.
+pub fn parse_amount_str(s: &str) -> Result<(U256, Asset), ParseAmountError> {
+    let s = s.trim();
+
+    // Split by whitespace to find asset
+    let (value_str, asset_str) =
+        s.rsplit_once(char::is_whitespace).ok_or(ParseAmountError::InvalidFormat)?;
```
capossele: Super minor comment but would be more reliable and less error prone: 
if instead of assuming that only a single space between the value and the symbol is allowed, we could use something like `split_whitespace().collect::<Vec<_>>()` and validating the two parts

crates/boundless-market/src/price_oracle/asset.rs
```diff
@@ -0,0 +1,567 @@
...
+    if to_decimals > from_decimals {
+        // Converting from lower to higher decimals (USD → ETH/ZKC)
+        // scale = 10^(to_decimals - from_decimals + price_decimals)
+        let exponent = to_decimals - from_decimals + price_decimals;
+        let scale = U256::from(10u128).pow(U256::from(exponent));
+        value
+            .checked_mul(scale)
+            .ok_or(ConversionError::Overflow)?
+            .checked_div(price)
+            .ok_or(ConversionError::Overflow)
```
capossele: If price is zero, this isn't technically speaking an overflow

crates/boundless-market/src/price_oracle/cached_oracle.rs
```diff
@@ -0,0 +1,235 @@
...
+        // Try to get from cache first
+        if let Some(rate) = self.get_cached_rate().await {
+            return Ok(rate);
+        }
+
+        // If not in cache, fetch from oracle and cache it.
+        // In practice this should never happen, as we fetch when we initialize the oracle.
+        tracing::warn!("Cache miss, fetching from oracle");
+
+        let rate = self.oracle.get_rate().await?;
```
capossele, line 74: I think this is fine as is. However if we imagine the case where multiple concurrent calls are made to `get_rate`, in case of a cache miss, multiple fetches to the oracle might happen. But again I believe this to be very rare given how we use the price oracle

crates/boundless-market/src/price_oracle/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// Get the current exchange rate
+    async fn get_rate(&self) -> Result<ExchangeRate, PriceOracleError>;
+}
+
+/// Trait for price sources - each instance is dedicated to one trading pair
+pub trait PriceSource: PriceOracle + Send + Sync {
+    /// Returns the name of this price source
+    fn name(&self) -> &'static str;
+}
```
capossele: Is there a specific reason why not include the name method directly as part of the PriceOracle trait instead of adding this additional trait?

crates/boundless-market/src/price_oracle/manager.rs
```diff
@@ -0,0 +1,480 @@
...
+            tokio::select! {
+                // Make sure to handle cancellation promptly
+                biased;
+
+                _ = cancel_token.cancelled() => {
+                    tracing::info!("Price oracle refresh task shutting down");
+                    break;
+                }
+
+                // Periodic tick to refresh rates and check staleness
```
capossele, line 115: Maybe here we could add before the tick something like
`self.refresh_all_rates().await` to minimize the impact of waiting until the first tick?
Or something equivalent if that makes sense

crates/boundless-market/src/price_oracle/config.rs
```diff
@@ -0,0 +1,348 @@
...
+    /// Off-chain source configuration
+    pub offchain: Option<OffChainConfig>,
+}
+
+impl Default for PriceOracleConfig {
+    fn default() -> Self {
+        Self {
+            eth_usd: PriceValue::Auto,
+            zkc_usd: PriceValue::Auto,
+            max_secs_without_price_update: 7200,
```
willpote: Imo default here can be much larger, especially as the current experience is completely static. Id even go as large as 12 hour old, as even 12 hour old price should be better than static values provers are using today.

crates/boundless-market/src/price_oracle/config.rs
```diff
@@ -0,0 +1,348 @@
...
+            refresh_interval_secs: 60,
+            timeout_secs: 10,
+            aggregation_mode: AggregationMode::Priority,
+            min_sources: 1,
+            // Enable both Chainlink and CoinGecko by default
+            onchain: Some(OnChainConfig { chainlink: Some(ChainlinkConfig { enabled: true }) }),
+            offchain: Some(OffChainConfig {
+                coingecko: Some(CoinGeckoConfig { enabled: true, api_key: None }),
+                cmc: None,
+            }),
```
willpote, line 118: Any reason we don't just enable CMC also? The more the better imo

willpote, line 118: Any reason we don't just enable CMC also? The more the better

crates/boundless-market/src/price_oracle/error.rs
```diff
@@ -0,0 +1,73 @@
...
+// limitations under the License.
+
+use super::TradingPair;
+use thiserror::Error;
+
+/// Price oracle error types
+#[derive(Error, Debug)]
+pub enum PriceOracleError {
+    /// All price sources failed for a trading pair
+    #[error("All price sources failed for {pair:?}")]
```
willpote: Could you add the formatting such that the errors will get logged along with an error code of [B-ORA-XX], similar to how other errors get logged? We use these error codes to set up automated alerting for pagerduty/Slack etc.

See errors in `market_monitor.rs` for an example of how we set it up

crates/boundless-market/src/prover_utils/requestor_pricing.rs
```diff
@@ -287,6 +288,24 @@ where
...
+
+    async fn convert_to_zkc(&self, amount: &Amount) -> Result<Amount, OrderPricingError> {
+        if amount.asset == Asset::ZKC {
+            return Ok(amount.clone());
+        }
+        Err(OrderPricingError::UnexpectedErr(Arc::new(anyhow::anyhow!(
+            "Price conversion not available (USD pricing requires price oracle)"
+        ))))
+    }
+
```
willpote, line 333: I'm confused about these functions- it seems we don't try to do any conversion here nor check if the oracle exists?

crates/boundless-market/src/price_oracle/error.rs
willpote: Could you add the formatting such that the errors will get logged along with an error code of [B-ORA-XX], similar to how other errors get logged? We use these error codes to set up automated alerting for pagerduty/Slack etc.

See errors in `market_monitor.rs` for an example of how we set it up

crates/broker/src/task.rs
```diff
@@ -212,6 +212,38 @@ where
...
+            PriceOracleError::HttpError(_) => "[B-PO-004]",
+            PriceOracleError::InvalidPrice(_) => "[B-PO-005]",
+            PriceOracleError::StalePrice { .. } => "[B-PO-006]",
+            PriceOracleError::ConfigError(_) => "[B-PO-007]",
+            PriceOracleError::Internal(_) => "[B-PO-008]",
+            PriceOracleError::UpdateTimeout() => "[B-PO-009]",
+        }
+    }
+}
+
```
willpote: Feels like these shouldn't live in this file, maybe just create a dedicated pricing_oracle.rs in src/?

crates/boundless-market/src/price_oracle/config.rs
jonastheis: CMC requires a API key for every request.  It needs to be specifically configured. That's why I disabled it by default.

jonastheis:
> Thinking out loud about what could brake after merging this; I think the nightly-prover might fail at parsing the broker.toml file as files like `ansible/roles/prover/configs/broker/8453.toml` ad similarly `nightly_8453.toml`, `prod_11155111.toml`, `staging_84532.toml`, `prod_84532.toml` use something like:
> 
> ```
> min_mcycle_price = "0.000000025" (no asset)
> min_mcycle_price_collateral_token = "0" (no asset)
> max_collateral = "200" (no asset)
> ```
> 
> Would that fail to parse?


Yes that would fail to parse. We could fallback to ETH/ZKC if no currency is specified to retain backward compatibility though. My initial thought was to make it breaking so that it becomes explicit and there's no surprises for a prover. After the upgrade a prover would need to re-configure the prices (potentially in USD) to keep operating with the new software.

jonastheis:
> One thing I'm wondering is what is the behavior when we breach the max stale price threshold. Does the broker completely shut down? Or just the pricing oracle shuts down, and we fall back to static? Perhaps its in the PR and I just didn't see it.
> 
> I think the behavior we want is likely fall back to static-> if not configured shutdown, but open to debate

Currently it is either:
1. `eth_usd = "auto"`: the price oracle is fetching prices and if the price couldn't be refreshed in `max_secs_without_price_update` (see [here)](https://github.com/boundless-xyz/boundless/blob/2eba91faa4aa2f6b2158e060092ed5332a403886/crates/boundless-market/src/price_oracle/config.rs#L88) then the entire broker shuts down.
2. static (ie `eth_usd = "2500.00"`): the price oracle always just returns the static value and thus the broker will never be shut down as the static value is always "fresh".

iiuc what you're saying you'd like to introduce a 3rd option:
- `eth_usd = "auto,2500.00"` (or similar, open for better suggestions to configure this intuitively): use 1, fallback to 2 if prices are too old instead of shutting down broker if static price is configured. if no static price configured, shut down broker.

We can do that but the question then is: Is a *very stale* statically configured price better than a *12h stale* price from the price oracle? Once we have successfully fetched a price in the price oracle this should technically always be better than the static price (if we assume the sources are trustworthy). 
So as you suggested above @willpote to increase `max_secs_without_price_update` to 12h, I think we could increase further to 24h. Then it should be *very unlikely* for this to ever become stale and the broker to shut down. I get the concern about worst case scenario and basically all provers crash at the same time. So this is probably not a satisfying option. 

Another thought: Perhaps we shouldn't even fail the node after `max_secs_without_price_update` but instead just log errors. 
- essentially we assume stale price is better than static price
- broker always keeps running
- we only keep options 1 (without crashing the node) and 2 from above 

Wdyt?

crates/boundless-market/src/price_oracle/manager.rs
jonastheis: added `// with ::interval the first tick completes immediately so we refresh right away.` as a comment so it is explicit in the code

crates/boundless-market/src/price_oracle/cached_oracle.rs
jonastheis: You're right in every aspect. This might be an issue but the way we use the oracle this should never happen as we always have a value in cache due to refreshing of prices on startup

crates/boundless-market/src/price_oracle/mod.rs
jonastheis: You're right, combined them in b94a9ccc96e2bdf73542f9edba6648719c915ce6

crates/boundless-market/src/price_oracle/asset.rs
jonastheis: changed in 42ee835f72191fd89a69eaa90bc7e7b10f63bf9e

crates/boundless-market/src/price_oracle/asset.rs
jonastheis: changed in 42ee835f72191fd89a69eaa90bc7e7b10f63bf9e

willpote:
> > One thing I'm wondering is what is the behavior when we breach the max stale price threshold. Does the broker completely shut down? Or just the pricing oracle shuts down, and we fall back to static? Perhaps its in the PR and I just didn't see it.
> > I think the behavior we want is likely fall back to static-> if not configured shutdown, but open to debate
> 
> Currently it is either:
> 
> 1. `eth_usd = "auto"`: the price oracle is fetching prices and if the price couldn't be refreshed in `max_secs_without_price_update` (see [here)](https://github.com/boundless-xyz/boundless/blob/2eba91faa4aa2f6b2158e060092ed5332a403886/crates/boundless-market/src/price_oracle/config.rs#L88) then the entire broker shuts down.
> 2. static (ie `eth_usd = "2500.00"`): the price oracle always just returns the static value and thus the broker will never be shut down as the static value is always "fresh".
> 
> iiuc what you're saying you'd like to introduce a 3rd option:
> 
> * `eth_usd = "auto,2500.00"` (or similar, open for better suggestions to configure this intuitively): use 1, fallback to 2 if prices are too old instead of shutting down broker if static price is configured. if no static price configured, shut down broker.
> 
> We can do that but the question then is: Is a _very stale_ statically configured price better than a _12h stale_ price from the price oracle? Once we have successfully fetched a price in the price oracle this should technically always be better than the static price (if we assume the sources are trustworthy). So as you suggested above @willpote to increase `max_secs_without_price_update` to 12h, I think we could increase further to 24h. Then it should be _very unlikely_ for this to ever become stale and the broker to shut down. I get the concern about worst case scenario and basically all provers crash at the same time. So this is probably not a satisfying option.
> 
> Another thought: Perhaps we shouldn't even fail the node after `max_secs_without_price_update` but instead just log errors.
> 
> * essentially we assume stale price is better than static price
> * broker always keeps running
> * we only keep options 1 (without crashing the node) and 2 from above
> 
> Wdyt?

Hmm, tricky. I think given as a prover you only get 2 pricing sources by default (and one is offchain), I'm a little concerned about the worst case scenario where all provers crash. 

So I think I'm aligned with what you propose. Provers can always just alert on the logs and decide if they should shut down or not.

crates/boundless-market/src/price_oracle/config.rs
willpote: Hmm I wonder if we should offer an oracle ourselves that essentially proxies CMC. Could just add it to the order stream API. I'm a little worried about just having 2 sources.

Out of scope for this PR but maybe lets discuss on Thurs

crates/boundless-market/src/price_oracle/error.rs
jonastheis: done in 95818d75035303b8252828ecbf089c8e599892ba

crates/broker/src/task.rs
jonastheis: done in 93aa78543bdcfed5220a47dd9f06eb2014d80d5d

crates/boundless-market/src/prover_utils/requestor_pricing.rs
jonastheis: Good catch. Initially I was thinking that due to the `OfferLayer` producing an offer in ETH in the `request_builder` (we convert there using the price oracle already), we don't need to convert here. However, after double checking I realized that we're actually converting the config values in `price_order` (which uses both `convert_to_eth` and `convert_to_zkc`). As such we need the price oracle here as well. Added in 32a1a9d50006be737ae8c9803fd9d0365e31cf1b

crates/boundless-market/src/price_oracle/config.rs
jonastheis: done in dc28a0bbfeb43fd9ffb34fb78b028864835c231b

willpote (approved):

austinabell (approved):

jonastheis:
> > Thinking out loud about what could brake after merging this; I think the nightly-prover might fail at parsing the broker.toml file as files like `ansible/roles/prover/configs/broker/8453.toml` ad similarly `nightly_8453.toml`, `prod_11155111.toml`, `staging_84532.toml`, `prod_84532.toml` use something like:
> > ```
> > min_mcycle_price = "0.000000025" (no asset)
> > min_mcycle_price_collateral_token = "0" (no asset)
> > max_collateral = "200" (no asset)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Would that fail to parse?
> 
> Yes that would fail to parse. We could fallback to ETH/ZKC if no currency is specified to retain backward compatibility though. My initial thought was to make it breaking so that it becomes explicit and there's no surprises for a prover. After the upgrade a prover would need to re-configure the prices (potentially in USD) to keep operating with the new software.

@capossele Forget what I said, I've changed the behavior to be backward compatible in 385e15b. So now it works with plain values without asset given and implicitly using a default asset just as before.

PR #1566, opened 2026-01-30
BM-2339: Zeroecco/broker config

zeroecco:


ansible/roles/prover/configs/broker/nightly_8453.toml
```diff
@@ -1,5 +1,7 @@
 [market]
-min_mcycle_price = "0.0000003"
+min_mcycle_price = "0.000000025"
+min_deadline = 300
```
willpote: Can drop

ansible/roles/prover/configs/broker/8453.toml
```diff
@@ -1,5 +1,7 @@
 [market]
-min_mcycle_price = "0.0000003"
+min_mcycle_price = "0.000000025"
+min_deadline = 300
```
willpote: Can drop

zeroecco: done

PR #1577, opened 2026-02-02
BM-2353: avoid requiring image ID to be specified for skipping requestor preflight

austinabell:
Fix from #1574 to not require image ID on request params to be filled by this point, diff to example tested with:

```diff
diff --git a/crates/boundless-market/examples/submit_echo.rs b/crates/boundless-market/examples/submit_echo.rs
index 5f84c655..b6a74f2a 100644
--- a/crates/boundless-market/examples/submit_echo.rs
+++ b/crates/boundless-market/examples/submit_echo.rs
@@ -53,8 +53,21 @@ async fn main() -> Result<()> {
         .build()
         .await?;
 
-    // Build and submit request using the builder pattern
-    let request = client.new_request().with_program(ECHO_ELF).with_stdin(b"Hello, Boundless!");
+    let input = b"Hello, Boundless!";
+
+    // For ECHO guest, journal == input
+    let journal = risc0_zkvm::Journal::new(input.to_vec());
+    // Static cycle count (doesn't need to be accurate for testing cache pre-fill)
+    let cycles = 1_000_000u64;
+
+    // Provide cycles, journal, and image_id to skip execution during preflight.
+    // The PreflightLayer will pre-fill the executor cache for pricing checks.
+    let request = client
+        .new_request()
+        .with_program(ECHO_ELF)
+        .with_stdin(input)
+        .with_cycles(cycles)
+        .with_journal(journal);
 
     let (request_id, expires_at) = client.submit(request).await?;
     println!("Submitted request: {:x}", request_id);

```

### DIFF
--- a/crates/boundless-market/src/request_builder/preflight_layer.rs
+++ b/crates/boundless-market/src/request_builder/preflight_layer.rs
@@ -66,6 +66,41 @@ impl PreflightLayer {
         };
         Ok(env)
     }
+
+    /// Ensures image_id is set, computing from program (inline or fetched) if needed.
+    async fn ensure_image_id(&self, params: RequestParams) -> anyhow::Result<RequestParams> {
+        if params.image_id.is_some() {
+            return Ok(params);
+        }
+        let program = match params.require_program() {
+            Ok(bytes) => bytes.to_vec(),
+            Err(_) => {
+                let url = params.require_program_url()?;
+                fetch_url(url.as_str()).await?
+            }
+        };
+        let image_id = risc0_zkvm::compute_image_id(&program)?;
+        Ok(params.with_image_id(image_id))
+    }
+
+    /// Best-effort: fills executor cache when we have all precomputed data.
+    async fn fill_executor_cache_if_ready(&self, params: &RequestParams) {
+        let (Some(image_id), Some(request_input), Some(cycles), Some(journal)) = (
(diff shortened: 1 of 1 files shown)


crates/boundless-market/src/request_builder/preflight_layer.rs
```diff
@@ -72,10 +72,19 @@ impl Adapt<PreflightLayer> for RequestParams {
...
     type Error = anyhow::Error;
 
-    async fn process_with(self, layer: &PreflightLayer) -> Result<Self::Output, Self::Error> {
+    async fn process_with(mut self, layer: &PreflightLayer) -> Result<Self::Output, Self::Error> {
         tracing::trace!("Processing {self:?} with PreflightLayer");
 
         if self.cycles.is_some() && self.journal.is_some() {
+            // Compute image_id from program if not yet provided.
+            if self.image_id.is_none() {
+                if let Ok(program_url) = self.require_program_url() {
```
capossele: Made a few suggestions directly here: https://github.com/boundless-xyz/boundless/pull/1578

capossele: but basically, it uses the program instead of fetching from url, Not 100% sure is the best as we might like this check, however it is sort of bad if this happens very frequently in car of high submission rate for the same program

PR #1579, opened 2026-02-02
BM-2354: improve gas price

capossele:
Switching from `eth_gasPrice` to `base_fee + max_priority_fee_per_gas` where `max_priority_fee_per_gas ` is the max 50th percentile across latest 10 blocks

### DIFF
--- a/crates/boundless-market/src/request_builder/preflight_layer.rs
+++ b/crates/boundless-market/src/request_builder/preflight_layer.rs
@@ -66,35 +66,53 @@ impl PreflightLayer {
         };
         Ok(env)
     }
+
+    /// Ensures image_id is set, computing from program (inline or fetched) if needed.
+    async fn ensure_image_id(&self, params: RequestParams) -> anyhow::Result<RequestParams> {
+        if params.image_id.is_some() {
+            return Ok(params);
+        }
+        let program = match params.require_program() {
+            Ok(bytes) => bytes.to_vec(),
+            Err(_) => {
+                let url = params.require_program_url()?;
+                fetch_url(url.as_str()).await?
+            }
+        };
+        let image_id = risc0_zkvm::compute_image_id(&program)?;
+        Ok(params.with_image_id(image_id))
+    }
+
+    /// Best-effort: fills executor cache when we have all precomputed data.
+    async fn fill_executor_cache_if_ready(&self, params: &RequestParams) {
+        let (Some(image_id), Some(request_input), Some(cycles), Some(journal)) = (
+            params.image_id,
+            params.request_input.as_ref(),
+            params.cycles,
+            params.journal.as_ref(),
+        ) else {
+            return;
+        };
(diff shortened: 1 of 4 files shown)


crates/boundless-market/src/request_builder/offer_layer.rs
```diff
@@ -464,16 +465,30 @@ where
...
+                BlockNumberOrTag::Latest,
+                &[50.0],
+            )
+            .await?;
+        let base_fee =
+            fee_history.latest_block_base_fee().unwrap_or(self.provider.get_gas_price().await?);
+        let priority_fee = fee_history
+            .reward
+            .as_ref()
+            .and_then(|r| r.iter().map(|b| b[0]).max()) // p50 across blocks
```
austinabell, line 481: Just want to be sure you want the max across those blocks, as the default is to use median in alloy

crates/boundless-market/src/request_builder/offer_layer.rs
```diff
@@ -464,16 +465,30 @@ where
...
+            .unwrap_or_default();
+        let gas_price = base_fee + priority_fee;
```
austinabell, line 483: Ideally this isn't duplicated and is consistently done by the provider, but I think that's too large of a change for this PR. Just commenting to link to in an issue

crates/boundless-market/src/request_builder/offer_layer.rs
capossele: Yeah the max is what we want as we should increase max_price high enough to also cover for increase gas cost by the time the broker will try to lock (can happen multiple hours later)

crates/boundless-market/src/request_builder/offer_layer.rs
capossele: Im still not sure we should couple this code with the provider, also because I'm still not convinced this approach is what we really need. Here we are attempting to grasp the current-ish gas price and then multiply by 2 and add that on top of the max_price. However this might still not be enough. Potentially we could change something at the contract level to reimburse the prover out of an additional gas deposit of the requestor but that seems an overcomplication

austinabell: yeah to be clear, I think this change is reasonable as is to pull in, I was just proposing if worth it to have it consistent across the board and more easily configurable, rather than diff percentiles and strategies used for different estimations/queries.

crates/boundless-market/src/request_builder/offer_layer.rs
austinabell: I suppose, intuitively to me seems more efficient to have a price set that can be completed within the timeout, rather than completed at any time during the timeout (assuming gas prices do not spike up dramatically for a long period). This definitely seems like a good strategy for short deadlines though

capossele:
Superseded by #1612

PR #1599, opened 2026-02-04
BM-2388: ansible pipelines

zeroecco:
initial kailua order generator role
ansible pipeline
clear out dead pulumi code for prover clusters

### DIFF
--- a/ansible/roles/kailua/tasks/main.yml
+++ b/ansible/roles/kailua/tasks/main.yml
@@ -1,5 +1,6 @@
 ---
 - name: Install Kailua binary from release (source)
+  when: kailua_install_method == "source"
   block:
     - name: Download Kailua binary
       ansible.builtin.get_url:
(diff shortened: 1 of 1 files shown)


crates/boundless-market/src/prover_utils/mod.rs
```diff
@@ -929,30 +929,41 @@ pub trait OrderPricingContext {
...
+                    format_ether(target_min_price)
+                );
 
-                    let target_time = order
-                        .request
-                        .offer
-                        .time_at_price(target_min_price)
-                        .context("Failed to get target price timestamp")?;
-                    (target_min_price, target_time)
-                };
```
austinabell: This feels like a weird strategy to have enabled, because in cases where an order has a ramp up time that is close to the lock timeout time, you will _never_ accept that order, so for the goals of backstopping or proving at a specific price, this seems unideal.

Unsure if this change was intended to be included in this pr, as it was referenced to be about the ansible pipelines (and no comment about this feature)

zeroecco: was not intended will remove

zeroecco: e98f3edd

ansible/roles/kailua/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: Remove Kailua archive
+      ansible.builtin.file:
+        path: "/tmp/{{ kailua_archive }}"
+        state: absent
+  when: kailua_install_method == "source"
+
+- name: Write Kailua launcher script
+  ansible.builtin.template:
+    src: kailua_launcher.sh.j2
```
austinabell: This path seems to whiff?

PR #1612, opened 2026-02-06
BM-2408: Use consistent base fee and priority fee estimations

austinabell:
removes all non-test usages of calls internally made to `eth_gasPrice` and adds a test to ensure the estimate is >= the actual gas usage.

### DIFF
--- a/crates/boundless-market/src/request_builder/offer_layer.rs
+++ b/crates/boundless-market/src/request_builder/offer_layer.rs
@@ -630,7 +630,9 @@ where
         max_price: U256,
         journal_bytes: Option<usize>,
     ) -> anyhow::Result<U256> {
-        // Use custom priority mode with a base fee multiplier of 250x, priority fee multiplier of 100x, priority fee percentile of 75.0, and dynamic multiplier percentage of 7 to give a buffer for gas price fluctuations.
+        // Intentionally more aggressive than PriorityMode::High (50th percentile, 2.5x base fee).
+        // Requestors use the 75th percentile so that `maxPrice` has enough headroom for gas
+        // volatility between request submission and fulfillment.
         let gas_price: u128 = PriorityMode::Custom {
             base_fee_multiplier_percentage: 250,
             priority_fee_multiplier_percentage: 100,
--- a/crates/broker/src/lib.rs
+++ b/crates/broker/src/lib.rs
@@ -1329,6 +1329,13 @@ mod unit_tests {
         // Deserialize and verify journal_bytes defaults to None.
         let deserialized: Order = serde_json::from_str(&json_without_field).unwrap();
         assert!(deserialized.journal_bytes.is_none());
+
+        // Verify that an explicit null also round-trips to None.
+        let mut value_null: serde_json::Value = serde_json::from_str(&json).unwrap();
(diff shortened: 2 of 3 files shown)


crates/boundless-market/src/request_builder/offer_layer.rs
```diff
@@ -464,10 +465,13 @@ where
         request_id: &RequestId,
         max_price: U256,
     ) -> anyhow::Result<U256> {
-        let gas_price: u128 = self.provider.get_gas_price().await?;
+        let gas_price: u128 = PriorityMode::default()
+            .estimate_max_fee_per_gas(&self.provider)
+            .await
+            .map_err(|err| anyhow::anyhow!("Failed to estimate gas price: {err:?}"))?;
```
austinabell: @capossele this could increase the gas estimation (but will be closer to what provers will price in). Wondering if worth to reduce the gas cost estimate. Right now 2x seems like it might be a lot in gas spikes, where this value should be roughly the gas price slippage tolerance. Do you think we should reduce that factor and if so do you have an intuition of what it should be?

capossele: Yeah setting that at 2x might indeed not be the right solution to account for potential future periods of higher gas prices. Maybe we should only add an additional 10 or 20% to that estimate. 

And trying to include some extra  buffer in the max_price instead. But this last comment should belong to a different PR

austinabell: Ah I thought https://github.com/boundless-xyz/boundless/pull/1579 had already come in. I think that strategy is reasonable, or we could have this use `PriorityMode::High` to be this estimation?

I'm quite unopinionated about this specific change, so let me know what you think an ideal strategy would be and I can change this PR to reflect that

capossele: High would use the 50th percentile, yeah that's ok

austinabell: [`2d22d86` (this PR)](https://github.com/boundless-xyz/boundless/pull/1612/commits/2d22d868f383e39e21b8997b2fd6847fed09229d)

Inferring that you meant to use this high value and remove the multiplier, if that's not what you meant, let me know!

austinabell:
draft because I think gas buffer might need to be increased here for requestors to handle gas price fluctuations better

crates/boundless-market/src/prover_utils/mod.rs
```diff
@@ -865,6 +870,23 @@ pub trait OrderPricingContext {
...
+        if order_predicate_type != PredicateType::ClaimDigestMatch {
+            let journal_gas = journal_len as u64 * config.fulfill_journal_gas_per_byte;
+            order_gas_cost += U256::from(gas_price) * U256::from(journal_gas);
+        }
+
+        // Re-check gas cost now that journal calldata gas is included.
+        if order_gas_cost > U256::from(order.request.offer.maxPrice) && !lock_expired {
+            return Ok(Skip {
+                reason: format!(
+                    "estimated gas cost to lock and fulfill order of {} exceeds max price of {}",
```
willpote: lets add "after accounting for journal costs" to distinguish the two cases

PR #1613, opened 2026-02-06
BM-2410: Order generator address rotation

capossele:
Adds key rotation support to the order generator, allowing it to cycle through a set of derived wallet addresses at a configurable interval. This prevents long-running order activity from being easily linked to a single address.

- Key derivation (`derivation.rs`): BIP-39/BIP-32 key derivation from a mnemonic, producing a funding key (index 0) and N rotation keys (indices 1..N)
- Indexer integration (`indexer.rs`, `indexer_client.rs`): Tracks open requests per address to determine when it's safe to rotate (i.e., no pending requests)
- Rotation logic (`rotation.rs`, `main.rs`): Orchestrates the rotation cycle, waits for outstanding requests to settle, sweeps funds back to the funding key, then activates the next key
- Infra (`order-generator.ts`, `index.ts`): Updates Pulumi infra to wire up the new mnemonic/rotation config;


crates/boundless-market/src/indexer_client.rs
capossele, line 1: TODO: in a follow up PR, these structs should be merged with the indexer one, to avoid code repetition and desync risks (already happening for the aggregation one)

capossele:

PR #1623, opened 2026-02-10
BM-2431: Add retry logic with exponential backoff for cycle count execution and redrive Lambda

willpote:
Adds automatic retry with exponential backoff for failed cycle count executions, and a new manually-invoked redrive Lambda for resetting FAILED (and optionally stuck EXECUTING) cycle counts back to PENDING.

Since the DB is in a VPC, its annoying/hard to access, hence the use of Lambda to do the DB query to reset the old rows back to PENDING. Alternative was setting up a bastion, but we use this lambda pattern elsewhere so I just re-used what we have. Its also nice that the logic for doing this will be in code and reviewed

Changes
* New `RETRY_PENDING` cycle count status with `retry_count`, `retry_after`, and `last_error` columns (migration 37). Failed executions are now retried with exponential backoff (configurable via `--max-retries`, `--retry-base-delay-secs`, `--retry-max-delay-secs`) before being marked FAILED.
* New `indexer-redrive` Lambda crate that resets FAILED and optionally stuck EXECUTING cycle counts back to PENDING, with dry-run support, requestor filtering, and lookback window. Deployed via Pulumi `RedriveLambda` component in the indexer VPC.
* Updated `get_cycle_counts_pending` to also pick up `RETRY_PENDING` rows whose `retry_after` has elapsed. Updated `count_cycle_counts_by_status` to include the new status. Added `set_cycle_counts_retry_pending` DB method.
* Updated tests to validate retry behavior: RETRY_PENDING pickup, retry exhaustion to FAILED, and adjusted existing failure tests to expect RETRY_PENDING instead of FAILED/PENDING.

### DIFF
--- a/crates/indexer/src/db/market.rs
+++ b/crates/indexer/src/db/market.rs
@@ -2790,22 +2790,53 @@ impl IndexerDb for MarketDb {
 
         let mut tx = self.pool.begin().await?;
 
-        for (request_digest, last_error, retry_after) in updates {
-            let query = "UPDATE cycle_counts
+        // 3 params per row (retry_after, last_error, request_digest) + 1 shared (updated_at)
+        const BATCH_SIZE: usize = 10;
+        for chunk in updates.chunks(BATCH_SIZE) {
+            let mut values_clauses = Vec::new();
+            let mut param_idx = 2; // $1 is reserved for updated_at
+
+            for (i, _) in chunk.iter().enumerate() {
+                if i == 0 {
+                    values_clauses.push(format!(
+                        "(${}::bigint, ${}::text, ${}::text)",
+                        param_idx,
+                        param_idx + 1,
+                        param_idx + 2,
+                    ));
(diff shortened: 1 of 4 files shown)


crates/indexer/src/db/market.rs
```diff
@@ -2751,20 +2777,62 @@ impl IndexerDb for MarketDb {
...
+
+        let mut tx = self.pool.begin().await?;
+
+        for (request_digest, last_error, retry_after) in updates {
+            let query = "UPDATE cycle_counts
+                 SET cycle_status = 'RETRY_PENDING',
+                     retry_count = COALESCE(retry_count, 0) + 1,
+                     retry_after = $1,
+                     last_error = $2,
+                     updated_at = $3,
```
unizarr: Is it worth doing batch of 10 here? Just in case if `updates` has 60+ results so we don't query 60+ times?

PR #1625, opened 2026-02-11
BM-2433: alerting, ansible roles, and code clean up

zeroecco:


infra/cw-monitoring/index.ts
```diff
@@ -0,0 +1,279 @@
...
+        name: `${prefix}-log-fatal`,
+        logGroupName: logGroup.name,
+        metricTransformation: {
+            namespace: alarmNamespace,
+            name: `${node.name}-log-fatal`,
+            value: "1",
+            defaultValue: "0",
+        },
+        pattern: "FATAL",
+    });
```
willpote: We should use the alarm setup we had previously, which tunes thresholds for errors based on which we care about vs not

zeroecco: fair enough. this way just made more sense to me at a glance. working on it.

zeroecco: [a01a5e1](https://github.com/boundless-xyz/boundless/pull/1625/commits/a01a5e11ce0dc7d62e8217ec986fbce28368de59)

PR #1630, opened 2026-02-11
BM-2442: Ensure indexer deploys before updating API lambda. Add max_price_cap to order generator + cleanups

willpote:
The Lambda which reads from the database was updated before the indexer ECS service was updated. The ECS service is responsible for applying migrations, which meant that the lambda started erroring as it was trying to read columns that weren't created yet. This fixes the Pulumi depends on relation so the lambda will only be updated after the indexer finishes updating.

Also various cleanups in the OG,
* adds a temporary "max price cap" and sets to 0.01 ETH on base (we should probably build this into the SDK itself, to be discussed)
* `max_price_per_mcycle` was left in the config, but being ignored. Also the param was not used in the OG code anymore. I I fixed the OG code to support it, then removed it from the config so we don't use it (since we want to rely on the automatic pricing)

Changes
* Adds new `max_price_cap` to OG to caps maxPrice with a warning if it exceeds the cap.
* Infra (order-generator): Wire `maxPriceCap` config through all requestor types, remove hardcoded `MAX_PRICE_PER_MCYCLE` values from several Pulumi stacks, add `MAX_PRICE_CAP` for evm-requestor on base prod.
* Infra (indexer): Expose `image` and `service` as public properties on `MarketIndexer` and `RewardsIndexer`, add `continueBeforeSteadyState: false` to Fargate services.

### DIFF
--- a/crates/indexer/src/market/service/log_processors.rs
+++ b/crates/indexer/src/market/service/log_processors.rs
@@ -59,7 +59,7 @@ async fn list_orders_v2_with_retry(
                         MAX_DELAY_MS,
                     );
                     tracing::warn!(
-                        "Failed to fetch orders (attempt {}/{}): {}. Retrying in {}ms",
+                        "Failed to fetch orders (attempt {}/{}): {:?}. Retrying in {}ms",
                         attempt + 1,
                         MAX_RETRIES + 1,
                         last_error.as_ref().unwrap(),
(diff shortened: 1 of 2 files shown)


crates/order-generator/src/main.rs
```diff
@@ -392,7 +421,31 @@ async fn handle_zeth_request(
         .with_env(env);
 
     // Build the request, including preflight, and assigned the remaining fields.
-    let request = client.build_request(request).await?;
+    let mut request = client.build_request(request).await?;
+
+    let m_cycles = max_cycles >> 20;
```
austinabell, line 426: Noting that this is micycles, not mcycles (it's not consistent, in certain places in r0/boundless, but I think it's been converging on actual mcycles rather than this). Especially confusing because we use micycles within the loop inputs

PR #1633, opened 2026-02-12
BM-1843: add expected probability for secondary fulfillment, calculate and convert total reward and prioritize

jonastheis:
- Adds a configurable `expected_probability_win_secondary_fulfillment parameter` (0-any), code default 50%) that scales the expected reward of secondary (lock-expired) orders by the probability of winning the proof race
- Removes the redundant `min_mcycle_price_collateral_token` config. `min_mcycle_price` is now dynamically converted to ZKC via the price oracle for secondary order evaluation
- Unifies primary (lock and fulfill) and secondary order prioritization into a single sort by converting all (discounted) rewards to ETH, so high-value secondary orders can rank above low-value primary orders
- Applies a random factor (0.2x–1.0x) to secondary order sort keys to spread provers across different orders and prevent all provers starting to compete in the same proof race if there's a secondary order with a high collateral value
- Internal/production provers use `expected_probability_win_secondary_fulfillment = 100` (no discount), since we want aggressive secondary fulfillment. The broker template and `generate-config` command recommends 50% for external provers.

### DIFF
--- a/crates/boundless-cli/src/commands/prover/generate_config.rs
+++ b/crates/boundless-cli/src/commands/prover/generate_config.rs
@@ -609,9 +609,7 @@ impl ProverGenerateConfig {
         let expected_probability_str =
             Text::new("Expected secondary fulfillment win multiplier (percentage, default 50):")
                 .with_default("50")
-                .with_help_message(
-                    "< 100 discounts reward, 100 = no change, > 100 boosts reward",
-                )
+                .with_help_message("< 100 discounts reward, 100 = no change, > 100 boosts reward")
                 .prompt()
                 .context("Failed to get expected probability")?;
 
--- a/crates/boundless-market/src/prover_utils/config.rs
+++ b/crates/boundless-market/src/prover_utils/config.rs
@@ -270,7 +270,6 @@ where
         .map_err(serde::de::Error::custom)
 }
 
-
 /// All configuration related to markets mechanics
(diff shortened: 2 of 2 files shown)


ansible/roles/prover/configs/broker/nightly_8453.toml
```diff
@@ -1,7 +1,7 @@
 [market]
 min_mcycle_price = "0.000000025"
+expected_probability_win_secondary_fulfillment = 100 # assume we win all secondary fulfillments -> min_mcycle_price is used to judge whether to accept a request or not
```
willpote: I actually think we should leave this at the default for 'nightly'. Intention with nightly is that it runs a very default/vanilla config, similar to what a new prover would run if just using the defaults, so we can test how the defaults perform in the wild

Aligned to make it 100 for the release cluster though!

ansible/roles/prover/configs/broker/8453.toml
```diff
@@ -1,7 +1,7 @@
 [market]
 min_mcycle_price = "0.000000025"
+expected_probability_win_secondary_fulfillment = 100 # assume we win all secondary fulfillments -> min_mcycle_price is used to judge whether to accept a request or not
```
willpote: Is it possible to go > 100? With this cluster, given its intention is to backstop, it would be good to have it always prioritize secondaries asap

PR #1649, opened 2026-02-16
BM-2461: [slasher] use confirmed block for range queries

capossele:
Use a confirmed block as the upper bound for block range queries instead of the chain tip.

**Problem**: Using the latest block as `to_block` caused “invalid block range params” when querying logs on Base mainnet, where the RPC rejects ranges that include the unstable tip.

**Change**: `current_block()` now prefers the safe block (`BlockNumberOrTag::Safe`). If the RPC doesn’t support it or the call fails, it falls back to latest 2 blocks.

### DIFF
--- a/crates/slasher/src/lib.rs
+++ b/crates/slasher/src/lib.rs
@@ -95,6 +95,8 @@ pub struct SlashServiceConfig {
     pub skip_addresses: Vec<Address>,
     pub tx_timeout: Duration,
     pub max_block_range: u64,
+    /// Blocks to subtract from latest when "safe" is not available (0 = use latest; used by tests).
+    pub block_confirmation_delay: u64,
 }
 
 impl SlashService<ProviderWallet> {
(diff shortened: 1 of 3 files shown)


crates/slasher/src/lib.rs
```diff
@@ -537,8 +538,21 @@ where
...
-        Ok(self.boundless_market.instance().provider().get_block_number().await?)
+        const FALLBACK_CONFIRMATION_DELAY: u64 = 2;
+        let provider = self.boundless_market.instance().provider();
+        // Prefer "safe" block so we don't query the unstable tip
+        match provider.get_block_by_number(BlockNumberOrTag::Safe).await {
+            Ok(Some(block)) => Ok(block.header.number),
+            Ok(None) | Err(_) => {
+                // RPC doesn't support "safe" or failed; use latest minus a few blocks
+                let latest = provider.get_block_number().await?;
+                Ok(latest.saturating_sub(FALLBACK_CONFIRMATION_DELAY))
```
austinabell: perhaps worth warn logging here to know if there is an error with getting the block number so that context isn't dropped?

PR #1654, opened 2026-02-17
BM-2466: Build Docker images once per production deployment

capossele:
Production pipelines currently rebuild the same Docker image independently for each chain stack (Base Sepolia, Eth Sepolia, Base Mainnet, etc.), wasting ~20-40 minutes of redundant Rust/Docker builds per service per deploy. 

This PR designates `l-prod-84532` (Base Sepolia) as the "builder" production stack. It builds the Docker image once and exports the image reference as a Pulumi stack output. Dependent production stacks (`l-prod-8453`, `l-prod-11155111`, `l-prod-1`) read this reference via pulumi stack output, set it as a config value, and skip their Docker build entirely. 

**Changes**:

1. Pipeline orchestration: builder at runOrder: 2, dependent stacks at runOrder: 3 (blocked if builder fails)
2. Buildspec: dependent stacks read image refs from the builder stack and inject them via pulumi config set
3. All services (slasher, distributor, order-stream, order-generator, indexer): conditional logic to either build the Docker image or use a pre-built IMAGE_URI
4. Indexer handles three separate images (MARKET_IMAGE_URI, REWARDS_IMAGE_URI, BACKFILL_IMAGE_URI)
5. Fixed Pulumi anti-pattern: IAM RolePolicy resources were being created inside .apply() callbacks (order-stream, indexer-infra)

Staging is unchanged, each staging stack still builds its own image. Manual pulumi up without the pipeline also works normally (falls back to building).


infra/image-builder/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (backfillDockerfile) {
+    const backfillImage = buildImage('backfill', backfillDockerfile, `backfill-${dockerTag}`);
+    outputs.backfillImageRef = backfillImage.ref;
+  }
+
+  const rewardsDockerfile = config.get('DOCKERFILE_REWARDS');
+  if (rewardsDockerfile) {
+    const rewardsImage = buildImage('rewards', rewardsDockerfile, `rewards-${dockerTag}`);
+    outputs.rewardsImageRef = rewardsImage.ref;
+  }
```
willpote, line 90: Its kind of awkward how we have this special case. Its possible to do lists in pulumi config (https://www.pulumi.com/docs/iac/concepts/config/#structured-configuration), so I'm imagining something like:

```
images:
    paths:
    - dockerfiles/indexer.dockerfile
    - dockerfiles/rewards-indexer.dockerfile
    - dockerfiles/market-indexer.dockerfile
``` 

Then we can parse the name of the image from the path, and set `outputs[name] = image.ref`

(or we could just use a comma separated list)

infra/pipelines/pipelines/base/LaunchDefaultPipeline.ts
```diff
@@ -242,6 +281,18 @@ ${additionalCommandsStr}          - ls -lt
...
+              echo "Dependent prod stack: reading image refs from builder stack $BUILDER_STACK"
+              IMAGE_REF=$(pulumi stack output imageRef --stack "$BUILDER_STACK" 2>/dev/null) || IMAGE_REF=""
+              [ -n "$IMAGE_REF" ] && pulumi config set IMAGE_URI "$IMAGE_REF" && echo "Set IMAGE_URI=$IMAGE_REF" || true
+              MARKET_REF=$(pulumi stack output marketImageRef --stack "$BUILDER_STACK" 2>/dev/null) || MARKET_REF=""
+              [ -n "$MARKET_REF" ] && pulumi config set MARKET_IMAGE_URI "$MARKET_REF" && echo "Set MARKET_IMAGE_URI" || true
+              REWARDS_REF=$(pulumi stack output rewardsImageRef --stack "$BUILDER_STACK" 2>/dev/null) || REWARDS_REF=""
+              [ -n "$REWARDS_REF" ] && pulumi config set REWARDS_IMAGE_URI "$REWARDS_REF" && echo "Set REWARDS_IMAGE_URI" || true
+              BACKFILL_REF=$(pulumi stack output backfillImageRef --stack "$BUILDER_STACK" 2>/dev/null) || BACKFILL_REF=""
+              [ -n "$BACKFILL_REF" ] && pulumi config set BACKFILL_IMAGE_URI "$BACKFILL_REF" && echo "Set BACKFILL_IMAGE_URI" || true
+            fi
```
willpote: If we use stack refs I don't think we need any of this

infra/pipelines/pipelines/base/LaunchDefaultPipeline.ts
```diff
@@ -255,8 +306,52 @@ ${additionalCommandsStr}          - ls -lt
...
+
+        // Dependent prod stacks receive BUILDER_STACK so the buildspec can read
+        // the pre-built image ref instead of rebuilding the Docker image.
+        if (builderStack) {
+            envVars.push({
+                name: "BUILDER_STACK",
+                type: "PLAINTEXT",
+                value: builderStack,
+            });
+        }
```
willpote: Also not needed with stack refs, reading stack is all done at runtime in pulumi

willpote:
One more thing, maybe lets test this on Slasher pipeline before updating doing all of them? This sort of Pulumi stuff is hard to get right the first time, and will avoid us breaking every pipeline :D

PR #1659, opened 2026-02-18
Switch docker compose to use pre-built docker images

austinabell:
The pattern will be that for `release-x` branches, we will constrain the prebuilt docker images to use latest minor version, and then for main when this gets backported, will switch that version to `latest`. `latest` overriden for every new release, `release-x.x` updated whenever new patch versions of that specific maj/min are published. We just need to constrain that patch releases don't contain breaking changes with this, otherwise we will need to specify speciifc patch versions on the version used by default (not ideal).

Other:
- CI release action switched to gh because old was deprecated and broken
- Build pre-built docker images, switch default to use those images
  - When specifying `BOUNDLESS_BUILD=...`, the binaries that are specified are build from source, using pre-built native binaries no longer used by default

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -12,7 +12,7 @@ x-base-environment: &base-environment
   RUST_BACKTRACE: 1
   POVW_LOG_ID: ${REWARD_ADDRESS:-${POVW_LOG_ID:-}}
 x-agent-common: &agent-common
-  image: ${AGENT_IMAGE:-ghcr.io/boundless-xyz/boundless/bento-agent:bento-v1.2}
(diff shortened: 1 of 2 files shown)


.github/workflows/bento-release.yml
```diff
@@ -25,7 +25,8 @@ env:
 jobs:
   build-bento:
     name: Build Bento Components
-    runs-on: [ self-hosted, prod, "Linux", "cpu" ]
+    runs-on: ubuntu-latest
```
capossele, line 28: Is this intentional to keep on `ubuntu-latest`?

austinabell: yes, the ami we were using is currently broken for building bento

capossele (approved):

Issue #1244, opened 2025-10-16
Excessive storage usage caused by images stored in MinIO


BattleB0t:
The disk usage of the MinIO volume increases by approximately 200 GB per day, causing the disk to fill up quickly.
As a result, the broker continues to claim orders until the collateral is empty, but does not complete them.

**Possible Solution**
- Set a TTL so that images are automatically removed after a certain period. (https://docs.min.io/enterprise/aistor-object-store/administration/object-lifecycle-management/create-lifecycle-management-expiration-rule/)
- Implement a scheduled cleanup process that removes images automatically every x hours.

austinabell:
Time based cleanup likely won't work because orders can be constructed in such a way for this to be a DOS condition. Cleaning up state based on the lifecycle of orders is something we are considering, thanks for the feedback!

derrix060:
I'm cleaning up doing this:

```bash
set -e
echo "=== MinIO Inputs Cleanup Sidecar Started ==="

# Configuration
INPUTS_PATH="/bitnami/minio/data/workflow/inputs"
HOURS_THRESHOLD=12
SLEEP_INTERVAL=14400  # 4 hours in seconds

while true; do
  echo "=== Cleanup cycle started at $(date) ==="
  echo "=== Storage Usage before cleanup ==="
  du -sh "${INPUTS_PATH}" 2>/dev/null || true
  
  # Find and delete directories not accessed in the last 12 hours
  # -amin +720 means accessed more than 720 minutes (12 hours) ago
  echo "Searching for input directories not accessed in the last ${HOURS_THRESHOLD} hours..."
  
  FOUND_COUNT=$(find "${INPUTS_PATH}" -maxdepth 1 -type d -amin +$((HOURS_THRESHOLD * 60)) ! -path "${INPUTS_PATH}" 2>/dev/null | wc -l || echo "0")
  echo "Found ${FOUND_COUNT} directories to delete"
  
  if [ ${FOUND_COUNT} -gt 0 ]; then
    echo "Deleting old directories..."
    find "${INPUTS_PATH}" -maxdepth 1 -type d -amin +$((HOURS_THRESHOLD * 60)) ! -path "${INPUTS_PATH}" -exec rm -rf {} + 2>/dev/null || true
    echo "Cleanup completed"
  else
    echo "No directories to delete"
  fi
  
  # Show current usage
  echo "=== Storage Usage after cleanup ==="
[...]

PR #1660, opened 2026-02-18
BM-2474: set minio TTL to 1 day, enable max order expiry config

austinabell:
aws/minio TTL is a bit tricky to test, but confirmed that running this sets the policy with:

```
docker compose exec minio mc ilm rule ls local/workflow
...
┌────────────────────────────────────────────────────────────────────────────────────────┐
│ Expiration for latest version (Expiration)                                             │
├──────────────────────┬─────────┬─────────┬──────┬────────────────┬─────────────────────┤
│ ID                   │ STATUS  │ PREFIX  │ TAGS │ DAYS TO EXPIRE │ EXPIRE DELETEMARKER │
├──────────────────────┼─────────┼─────────┼──────┼────────────────┼─────────────────────┤
│ d6ailbe6vpls76nmb770 │ Enabled │ inputs/ │ -    │              1 │ false               │
└──────────────────────┴─────────┴─────────┴──────┴────────────────┴─────────────────────┘
```

Setting the policy through aws sdk APIs are broken with minio, which is why `mc` CLI is used -- so this wouldn't set a TTL if not using minio

Alternative to this is having a script run on some interval, similar to https://github.com/boundless-xyz/boundless/issues/1244#issuecomment-3523729272 -- but this is intended to avoid having an extra process running (at the cost that expiries can only be configured based on number of days). Do not feel strongly about this, if we want to have more granular control over the cleanup deletions, happy to swap this to a loop script.

### DIFF
--- a/ansible/roles/prover/configs/broker/8453.toml
+++ b/ansible/roles/prover/configs/broker/8453.toml
@@ -19,3 +19,4 @@ max_concurrent_proofs = 2
 max_journal_bytes = 20000
 max_file_size = 500000000000
 max_mcycle_limit = 500000000000
+max_order_expiry_secs = 86400
--- a/ansible/roles/prover/configs/broker/nightly_8453.toml
+++ b/ansible/roles/prover/configs/broker/nightly_8453.toml
@@ -17,3 +17,4 @@ max_concurrent_proofs = 2
 max_journal_bytes = 20000
 max_file_size = 500000000000
 max_mcycle_limit = 500000000000
+max_order_expiry_secs = 86400
--- a/ansible/roles/prover/configs/broker/prod_11155111.toml
+++ b/ansible/roles/prover/configs/broker/prod_11155111.toml
@@ -16,6 +16,7 @@ priority_requestor_addresses = ["0x0466ACfc0F27bBA9fBB7A8508f576527e81E83Bd"]
 # max_fetch_retries = 2
 # allow_client_addresses = []
 # lockin_priority_gas = 100
+max_order_expiry_secs = 86400
 
 [prover]
 status_poll_retry_count = 3
--- a/ansible/roles/prover/configs/broker/prod_84532.toml
+++ b/ansible/roles/prover/configs/broker/prod_84532.toml
@@ -19,6 +19,7 @@ deny_requestor_addresses = [
 # max_fetch_retries = 2
 # allow_client_addresses = []
 # lockin_priority_gas = 100
+max_order_expiry_secs = 86400
 
 [prover]
 status_poll_retry_count = 3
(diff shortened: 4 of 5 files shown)


compose.yml
```diff
@@ -182,6 +182,26 @@ services:
...
+    depends_on:
+      minio:
+        condition: service_healthy
+    restart: "no"
+    entrypoint:
+      - /bin/sh
+      - -c
+      - |
+        mc alias set local http://$${MINIO_HOST}:9000 $${MINIO_ROOT_USER} $${MINIO_ROOT_PASSWORD} &&
+        mc ilm rule add local/$${MINIO_BUCKET} --prefix "inputs/" --expire-days "$${INPUT_TTL_DAYS}" &&
```
capossele: `mc ilm rule add` appends a new rule every time. If you docker compose down && docker compose up with the same volume, you accumulate duplicate rules. Maybe could be improved with something like:

```
mc ilm rule list local/${MINIO_BUCKET} --prefix "inputs/" --json | grep -q "inputs/" ||
mc ilm rule add local/${MINIO_BUCKET} --prefix "inputs/" --expire-days "${INPUT_TTL_DAYS}"
```

crates/boundless-market/src/prover_utils/mod.rs
```diff
@@ -601,6 +601,17 @@ pub trait OrderPricingContext {
             });
         }
 
+        // Check if the order expiry exceeds the max allowed duration
+        if let Some(max_expiry) = config.max_order_expiry_secs {
```
capossele, line 605: I'm a bit worried on its potential misuses. Do we know looking at the market how long the expiration is? I mean is a day enough for instance? I feel this type of information would be useful to provide to educate the user on how to set this and what values are safe/recommended

capossele: and potentially we could warn if it gets set to a uncommon value?

austinabell: On the requestor side, it would warn if a request expiry is past that deadline because it runs this function now.

I do get the concern, that there may be a reason for an order with expiries (lock timeout and expiry) longer than a day, but at least to me seems like a more reasonable default than allowing unbounded on the prover

austinabell: or do you mean user being a prover and what they should override this config to if they want?

capossele: yes that, It could also just live on the docs

austinabell: https://github.com/boundless-xyz/docs/pull/40 sufficient?

PR #1663, opened 2026-02-18
BM-2012: Per-requestor and per-selector `min_mcycle_price` pricing overrides

capossele:
Adds support for configuring `min_mcycle_price` overrides per requestor address, per selector, or per requestor+selector combo via the broker.toml file (`market.pricing-overrides` section). This allows operators to price specific requestors or proof types (e.g. Groth16) differently without changing the global default.

### DIFF
--- a/crates/broker/src/order_picker.rs
+++ b/crates/broker/src/order_picker.rs
@@ -3080,7 +3080,6 @@ pub(crate) mod tests {
     async fn test_calculate_exec_limits_proof_type_override_takes_priority() {
         let mut market_config = MarketConfig::default();
         market_config.min_mcycle_price = Amount::parse("0.001 ETH", None).unwrap();
-        market_config.min_mcycle_price_collateral_token = Amount::parse("10 ZKC", None).unwrap();
         market_config.max_mcycle_limit = 100_000;
 
         let config = ConfigLock::default();


broker-template.toml
```diff
@@ -25,6 +25,27 @@
...
+#     },
+#     "by_requestor_selector": {
+#       "0xAbC...123:0x12345678": { "min_mcycle_price": "0.001 USD" }
+#     }
+#   }
+#
+# Resolution priority: requestor+selector > selector > requestor > global default.
+# The file is automatically hot-reloaded every 60 seconds; no broker restart needed.
+# Copy pricing-overrides.template.json to pricing-overrides.json to get started.
+pricing_overrides_path = "pricing-overrides.json"
```
austinabell: Why have it as another file that is being watched instead of just handling it as toml within this config?

capossele: It felt more natural to use a JSON for this particular data structure; and the idea was to keep this file clean.

But ultimately I'm fine with anything, don't have a strong opinion either way

crates/boundless-market/src/prover_utils/config.rs
```diff
@@ -513,6 +516,13 @@ pub struct MarketConfig {
     /// market. This should remain false to avoid losing partial PoVW jobs.
     #[serde(default)]
     pub cancel_proving_expired_orders: bool,
+    /// Per-requestor and per-selector `min_mcycle_price` overrides.
+    ///
+    /// When the broker evaluates an order, it resolves the effective `min_mcycle_price`
+    /// with priority: `by_requestor_selector` > `by_selector` > `by_requestor` > global default.
+    /// Changes to this section are picked up automatically when the broker reloads `broker.toml`.
+    #[serde(default)]
+    pub pricing_overrides: PricingOverrides,
```
willpote: Overall this looks good, I'm think we should name it "proof type" instead of "selector", as proof type is what we use in the docs + SDK

willpote (commented):
Overall this looks good, I'm think we should name it "proof type" instead of "selector", as proof type is what we use in the docs + SDK

crates/boundless-market/src/prover_utils/config.rs
capossele: Done in https://github.com/boundless-xyz/boundless/pull/1663/commits/346f7e81cd7aee5afa8a1c8b5b696e1d78e6aa5e

broker-template.toml
capossele: removed

PR #1669, opened 2026-02-19
BM-2449: Pre-lock check before lock tx

capossele:
Before sending a lock transaction, we re-check that gas cost is within the order’s current price. If not, we skip locking (and, potentially, retry later). This avoids locking when gas has spiked.


crates/boundless-market/src/prover_utils/config.rs
```diff
@@ -398,6 +403,9 @@ pub struct MarketConfig {
     /// `gas_priority_mode = { custom = { base_fee_multiplier_percentage = 300, priority_fee_multiplier_percentage = 150, priority_fee_percentile = 15.0, dynamic_multiplier_percentage = 5 } }`.
     #[serde(default = "defaults::priority_mode")]
     pub gas_priority_mode: PriorityMode,
+    /// When true (default), run a pre-lock check (expiry + gas vs max price) immediately before sending the lock tx. Set to false to skip the check and lock with the order as-is.
+    #[serde(default = "defaults::pre_lock_check_enabled")]
+    pub pre_lock_check_enabled: bool,
```
austinabell: What is a reason you wouldn't want to do this check? Seems a bit unnecessary to have this as a config option (I don't feel strongly about this, feel free to keep)

crates/boundless-market/src/prover_utils/mod.rs
```diff
@@ -238,6 +238,92 @@ impl std::fmt::Display for OrderRequest {
...
+    supported_selectors: &SupportedSelectors,
+    current_gas_price: u128,
+) -> Result<(), PreLockSkipReason> {
+    let now = now_timestamp();
+    if order.expiry() <= now {
+        return Err(PreLockSkipReason::Expired);
+    }
+    if !supported_selectors.is_supported(order.request.requirements.selector) {
+        return Err(PreLockSkipReason::UnsupportedSelector);
+    }
```
austinabell: Doesn't make sense to me why do the supported selectors check here also? It's done during pricing and the `LightPreLockChecker` doesn't even update dynamically based on config?

crates/boundless-market/src/prover_utils/mod.rs
```diff
@@ -238,6 +238,92 @@ impl std::fmt::Display for OrderRequest {
...
+pub async fn check_order_ok_to_lock(
+    order: &OrderRequest,
+    config: &MarketConfig,
+    supported_selectors: &SupportedSelectors,
+    current_gas_price: u128,
+) -> Result<(), PreLockSkipReason> {
+    let now = now_timestamp();
+    if order.expiry() <= now {
+        return Err(PreLockSkipReason::Expired);
+    }
```
austinabell: a bit redundant, because the entries are only kept for as long as they are not expired, and also it's checked in `is_within_deadline` which I don't see this removing?

crates/boundless-market/src/prover_utils/mod.rs
```diff
@@ -238,6 +238,92 @@ impl std::fmt::Display for OrderRequest {
...
+        let fulfill_gas = estimate_gas_to_fulfill(config, supported_selectors, &order.request)
+            .await
+            .map_err(|e| PreLockSkipReason::Unexpected(e.to_string()))?;
+        U256::from(lock_gas + fulfill_gas)
+    };
+    let order_gas_cost = U256::from(current_gas_price) * order_gas;
+    let max_price = U256::from(order.request.offer.maxPrice);
+    if order_gas_cost > max_price {
+        return Err(PreLockSkipReason::GasTooHigh { order_gas_cost, max_price });
+    }
```
austinabell: Checking `order_gas_cost <= max_price` seems not what we want here, because it would be locking before this max price so this could still lock when gas > actual reward

crates/boundless-market/src/prover_utils/mod.rs
```diff
@@ -238,6 +238,92 @@ impl std::fmt::Display for OrderRequest {
...
+
+/// Trait for checking that an order is still ok to lock immediately before sending the lock tx.
+/// Implementations can be full (e.g. re-run full pricing) or light (expiry + gas only).
+#[cfg(feature = "prover_utils")]
+#[async_trait::async_trait]
+pub trait PreLockChecker: Send + Sync {
+    /// Returns Ok(()) if the order should be locked with its current request/signature; Err with
+    /// a reason to skip locking.
+    async fn check_ok_to_lock(&self, order: &OrderRequest) -> Result<(), PreLockSkipReason>;
+}
```
austinabell: This trait seems like it's just used to be able to mock a failure in a test. Is it not possible to test the failure case(s) without this so we actually test the logic added in the function? Seems a bit unnecessary to add this trait if we can go without it

crates/broker/src/order_monitor.rs
```diff
@@ -161,6 +213,10 @@ pub struct OrderMonitor<P> {
     supported_selectors: SupportedSelectors,
     rpc_retry_config: RpcRetryConfig,
     gas_priority_mode: Arc<tokio::sync::RwLock<PriorityMode>>,
+    /// Consecutive pre-lock retry count per order. Cleared on successful lock or
+    /// non–PreLockCheckRetry error in lock_and_prove_orders, and in skip_order
+    /// when invalidating lock_and_prove_cache so the map does not grow without bound.
+    pre_lock_retry_count: Arc<Mutex<HashMap<String, u32>>>,
```
austinabell: would be nice not to have to keep a mutex map of just the retries, curious if you've considered adding a field to `OrderRequest`? Then you could modify with an atomic or just having the mutex per order rather than having a shared cache here?

austinabell:
My intuition is that during pricing, you have the gas estimated that will be used, so you could just pull gas price and check profitability that way? We are * by gas price currently but could change or add another field to `OrderRequest` to avoid duplicate calculations. Seems like it would be a cheap way to do this check?

PR #1674, opened 2026-02-20
BM-2493: Update dockerfile and ansible to use pre-built dockerfiles

austinabell:
Backport of #1659 with the following additional changes:
- Uses `latest` instead of latest 1.2
- Add support for configuring these through ansible

TODO:
- [ ] Switch release clusters to use `release-1.2` branch instead of `main`: `prover_version: "release-1.2"` in inventory (not necessary if using all pre-built dockerfiles, can keep using main if we assume no compose breaking change compatibilities between the two, but I doubt this)
  - `prover_image_version: "1.2"` if we don't do this
- [ ] `prover_boundless_build: "all"` in nightly inventory file
- [ ] For any services that override the docker image, `prover_boundless_build: "rest_api agent"` replacing
- [x] Fix `prover_boundless_build` handling. Currently the justfile handles how `BOUNDLESS_BUILD=".."` is handled where it unsets the dockerfile reference and builds only the services mentioned. The current ansible setup just builds all (which might work, but docker compose can be finnicky.
- [ ] Use `--pull always` flag to ensure latest images pulled when starting

Depending on which option taken, the ansible config can likely be cleaned up. Ideally `BOUNDLESS_BUILD=".."` is set to define which services are built from scratch, docker images can be overriden for those that are (in cases of CPU only docker images), and then for when using prebuilt docker images either the git branch used defines which is used or is specified through ansible config with the versioning.

### DIFF
--- a/ansible/roles/docker/templates/daemon.json.j2
+++ b/ansible/roles/docker/templates/daemon.json.j2
@@ -1,6 +1,11 @@
 {
     "iptables": false,
     "log-driver": "{{ docker_log_driver }}"{{ "," if docker_zfs_support == "true" or docker_nvidia_enabled else "" }}
+    {% if docker_log_driver == "journald" %}
+    "log-opts": {
+      "tag": "bento"
+    }{{ "," if docker_zfs_support == "true" or docker_nvidia_enabled else "" }}
+    {% endif %}
     {% if docker_zfs_support == "true" %}
     "storage-driver": "zfs",
     "storage-opts": ["zfs.fsname=prover"]{{ "," if docker_nvidia_enabled else "" }}
--- a/ansible/roles/nvidia/tasks/main.yml
+++ b/ansible/roles/nvidia/tasks/main.yml
@@ -6,29 +6,41 @@
   changed_when: false
   failed_when: false
 
-- name: Install NVIDIA drivers and CUDA toolkit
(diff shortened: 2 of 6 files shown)


.github/workflows/bento-release.yml
```diff
@@ -65,8 +70,8 @@ jobs:
           ls -la /usr/local/cuda/bin/ || echo "CUDA bin not found"
           ls -la /usr/bin/gcc || echo "System gcc not found"
 
-      - name: setup sccache + s3
-        uses: ./.github/actions/sccache
+      # - name: setup sccache + s3
```
zeroecco, line 73: are we dropping sccache?

austinabell: Morgan mentioned that it wouldn't work with this new runner (previous Ami was broken). You'd know better than me, I'd like to keep sccache

zeroecco: k.

austinabell:
> approved pending checks

The check failing wasn't really related to this PR, just that sccache was broken for source build, and we haven't been doing source docker builds in CI for this check, and CI was just running the docker compose build.

I decided to make the change to have docker compose use pre-builds by default for now, as this is also less breaking from existing workflows, at the cost that we will need to keep this pre-build binary version updated still. Can open an issue to make this more streamlined later.

PR #1678, opened 2026-02-23
BM-2496: Use 50th percentile and 20% buffer for requestor gas estimation

capossele:
Updates gas estimation in the offer layer: use `PriorityMode::High` (50th percentile) instead of the previous custom 75th-percentile setup, and replace the 2× gas cost multiplier with a 20% buffer when computing the adjusted max price for requests.

The previous gas estimate overhead on `max_price` was affecting the price at which provers were locking on staging. Switching to the 50th percentile and a 20% buffer reduces that overhead while still leaving room for gas volatility


capossele:
Re-evaluating the correct fix

PR #1689, opened 2026-02-25
BM-2511: cap `ramp_up_start` to 1 hour

capossele:
Currently `ramp_up_start` is computed as 

`now_timestamp() + self.base_ramp_up_delay.max(self.executor_time(Some(cycle_count)) as u64)`

that is unbound.

This PR introduces `MAX_RAMP_UP_PERIOD = 3600` used as upper bound.


crates/boundless-market/src/request_builder/mod.rs
```diff
@@ -773,11 +773,15 @@ impl ParameterizationMode {
     /// The ramp up start is calculated as the current timestamp plus the base ramp up delay
     /// or the required executor time, whichever is greater.
     fn recommended_ramp_up_start(&self, cycle_count: Option<u64>) -> u64 {
+        const MAX_RAMP_UP_DELAY: u64 = 3600; // 1 hour
```
capossele, line 776: To give some context, Signal requests use 30 minutes

PR #1691, opened 2026-02-25
BM-2513: Add `zstd` compression support for GuestEnv inputs

capossele:
- Adds a new V2 encoding version for `GuestEnv` that applies zstd compression on top of MessagePack, reducing input size for both inline and uploaded data
- Decompression is transparent, all existing `GuestEnv::decode()` call sites handle V2 automatically with zero changes
- Includes decompression bomb protection via streaming size-limited decompression (256 MiB default cap, configurable via `decode_with_limit`)

## Details

**New APIs:**
- `GuestEnv::encode_compressed()` — produces V2 (`[0x02][zstd(msgpack)]`)
- `GuestEnv::decode_with_limit(bytes, max)` — decode with custom decompression cap
- `GuestEnvBuilder::with_compression()` — opt-in compression flag for the builder
- `StorageLayerConfig::compress_input` — layer-level compression toggle

**Encoding format:** `[version_byte: u8][payload]`
- V0: raw bytes (legacy, unchanged)
- V1: MessagePack (default, unchanged)
- V2: MessagePack + zstd (new, opt-in on encode, transparent on decode)


crates/boundless-market/src/input.rs
```diff
@@ -116,6 +163,21 @@ impl GuestEnv {
...
     }
 
+    /// Encode the [GuestEnv] with zstd compression (V2) for inclusion in a proof request.
+    ///
+    /// Produces a V2-encoded payload: `[0x02][zstd(msgpack(self))]`.
+    /// Use this when the input data is large enough that compression provides meaningful
+    /// size reduction (e.g. for storage/bandwidth savings).
+    pub fn encode_compressed(&self) -> Result<Vec<u8>, Error> {
+        let msgpack = rmp_serde::to_vec_named(self)?;
+        let compressed = zstd::encode_all(msgpack.as_slice(), ZSTD_COMPRESSION_LEVEL)
```
jonastheis: should add a check here to not accidentally produce a decompression bomb and then fallback to uncompressed I guess

crates/boundless-market/src/input.rs
```diff
@@ -20,6 +22,12 @@ use serde::{Deserialize, Serialize};
 
 use crate::contracts::RequestInput;
 
+/// Default zstd compression level for V2 encoding.
+const ZSTD_COMPRESSION_LEVEL: i32 = 3;
```
jonastheis: why only 3? we should make it either customizable or go a higher by default. it's only more expensive at compression time, not decompression time

PR #1704, opened 2026-03-03
ensure that for a given user you complete their oldest job first

zeroecco:
```sql
taskdb=# EXPLAIN ANALYZE SELECT * FROM request_work('prove');
                                                   QUERY PLAN
-----------------------------------------------------------------------------------------------------------------
 Function Scan on request_work  (cost=0.25..10.25 rows=1000 width=116) (actual time=0.390..0.390 rows=0 loops=1)
 Planning Time: 0.029 ms
 Execution Time: 0.402 ms
(3 rows)

taskdb=#
```

### DIFF
--- a/bento/crates/taskdb/migrations/6_tasks.sql
+++ b/bento/crates/taskdb/migrations/6_tasks.sql
@@ -25,7 +25,7 @@ BEGIN
       WHERE tasks.stream_id = stream
         AND tasks.state = 'ready'
       ORDER BY
-        (SELECT MIN(t2.created_at) FROM tasks t2 WHERE t2.job_id = tasks.job_id AND t2.state = 'ready'),
+        (SELECT MIN(t2.created_at) FROM tasks t2 WHERE t2.job_id = tasks.job_id),
         created_at ASC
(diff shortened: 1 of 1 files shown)


bento/crates/taskdb/migrations/6_tasks.sql
```diff
@@ -0,0 +1,56 @@
...
+  FOR UPDATE SKIP LOCKED;
+
+  IF stream IS NOT NULL THEN
+    WITH selected_task AS (
+      SELECT tasks.job_id, tasks.task_id, tasks.task_def, tasks.max_retries
+      FROM tasks
+      WHERE tasks.stream_id = stream
+        AND tasks.state = 'ready'
+      ORDER BY
+        (SELECT MIN(t2.created_at) FROM tasks t2 WHERE t2.job_id = tasks.job_id AND t2.state = 'ready'),
```
austinabell:
```suggestion
        (SELECT MIN(t2.created_at) FROM tasks t2 WHERE t2.job_id = tasks.job_id),
```

I think? Confirming now

PR #1705, opened 2026-03-03
BM-2559: speed up ansible, reduce config churn

zeroecco:
mv to upstream gha
nvidia stop running everytime
minio lifecycle at the end of a run


ansible/roles/prover/handlers/main.yml
```diff
@@ -10,3 +8,63 @@
     name: bento
     state: restarted
     daemon_reload: true
+  notify: Configure MinIO lifecycle
+
+- name: Check MinIO health for lifecycle configuration
+  ansible.builtin.uri:
+    url: "http://127.0.0.1:9000/minio/health/live"
```
willpote: So are we not using the default minio lifecycle configuration we provide in the compose anymore?

zeroecco: This existed before that. I just cleaned it up so it didn't give false failures. I can drop it if that is the preference.

willpote: Yes, think we've discussed before but we want to have as much coverage on the compose file we are vending to provers as possible, so would be ideal to drop

zeroecco: 8cac5452

PR #1711, opened 2026-03-04
BM-2596: Add telemetry APIs to order stream

willpote:
This is part 1 of the telemetry effort, and just includes order stream changes. Part 2 will be broker changes.

Adds telemetry endpoints for brokers to report heartbeats and per-request evaluation/completion data. Data flows through signed API endpoints on the order-stream, into Kinesis Data Streams, and lands in Redshift Serverless for analytics. Intention is that broker will call the API periodically, submitting telemetry in batches. As (weak) spam protection, we require submitters to maintain a balance of ZKC on the market, similar to our Websocket endpoint. Additional fields will be added to the data model during Part 2, current ones are mostly a starting point.

See READMEs for more information about architecture/design.

## Changes
* New telemetry types in `boundless-market` (`BrokerHeartbeat`, `RequestHeartbeat`, `RequestEvaluated`, `RequestCompleted`) with EIP-191 signature auth and collateral balance checks
* Two new order-stream endpoints: `POST /api/v2/heartbeats` (broker config/uptime) and `POST /api/v2/heartbeats/requests` (evaluation + completion records)
* Pulumi component (`TelemetryInfra`) provisioning 3 Kinesis streams, Redshift Serverless namespace/workgroup, IAM roles, and security groups — always enabled for all deployments
* Redshift streaming ingestion via materialized views on raw JSON (`SUPER`), with typed SQL views on top for easy schema evolution
* Numbered migration system (`apply.sh`) with idempotent user management and `envsubst` templating, integrated into the deployment pipeline via `post-deploy.sh`
* Telemetry client helpers in `order_stream_client.rs` for signing and sending heartbeats
* Integration test (`tests/telemetry.rs`) that sends heartbeats to a live deployment and verifies data in Redshift
* READMEs for both the crate and infra directories

### DIFF
--- a/crates/boundless-cli/src/lib.rs
+++ b/crates/boundless-cli/src/lib.rs
@@ -197,19 +197,6 @@ pub async fn ensure_prover_has_image(
     Ok(())
 }
 
-/// The default prover implementation.
-/// This [DefaultProver] uses the default zkVM prover.
-/// The selection of the zkVM prover is based on environment variables.
-///
-/// The `RISC0_PROVER` environment variable, if specified, will select the
-/// following [Prover] implementation:
-/// * `bonsai`: [BonsaiProver] to prove on Bonsai.
-/// * `local`: LocalProver to prove locally in-process. Note: this
-///   requires the `prove` feature flag.
-/// * `ipc`: [ExternalProver] to prove using an `r0vm` sub-process. Note: `r0vm`
-///   must be installed. To specify the path to `r0vm`, use `RISC0_SERVER_PATH`.
-///
-/// If `RISC0_PROVER` is not specified, the following rules are used to select a
 /// Orchestrates the fulfillment of proof requests by coordinating order proving,
 /// set building, and assessor verification.
 ///
--- a/dockerfiles/agent.cpu.dockerfile
+++ b/dockerfiles/agent.cpu.dockerfile
@@ -2,7 +2,7 @@
 # CPU-only agent build (no CUDA/GPU support)
 ARG RUST_IMG=rust:1.88-bookworm
 ARG RUNTIME_IMG=debian:bookworm-slim
-ARG S3_CACHE_PREFIX="public/rust-cache-docker-Linux-X64/sccache"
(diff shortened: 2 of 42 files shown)


crates/order-stream/src/heartbeat.rs
```diff
@@ -0,0 +1,406 @@
...
+const BALANCE_CACHE_MAX_ENTRIES: u64 = 10_000;
+
+/// Cached collateral balance keyed by address.
+pub(crate) type BalanceCache = Cache<Address, U256>;
+
+/// Creates a new balance cache with TTL eviction.
+pub(crate) fn new_balance_cache() -> BalanceCache {
+    Cache::builder()
+        .max_capacity(BALANCE_CACHE_MAX_ENTRIES)
+        .time_to_live(Duration::from_secs(BALANCE_CACHE_TTL_SECS))
```
capossele, line 43: Does that mean that if I realize I need to put more collateral, the new value gets updated only after this cache expires? e.g., at worst after `BALANCE_CACHE_TTL_SECS`?

crates/order-stream/src/heartbeat.rs
```diff
@@ -0,0 +1,406 @@
...
+) -> Result<StatusCode, Response> {
+    let heartbeat: BrokerHeartbeat = serde_json::from_slice(&body).map_err(|e| {
+        (StatusCode::BAD_REQUEST, format!("Invalid heartbeat payload: {e}")).into_response()
+    })?;
+
+    verify_heartbeat_auth(&state, &headers, &body, heartbeat.broker_address)
+        .await
+        .map_err(|e| e.into_response())?;
+
+    let partition_key = format!("{:?}", heartbeat.broker_address);
```
capossele: Would it be better to use Display for the address here and everywhere else?

```suggestion
    let partition_key = format!("{heartbeat.broker_address}");
```

crates/order-stream/src/heartbeat.rs
willpote: Thats correct, so your telemetry would be blocked for a bit. This is probably too long though so will reduce it

PR #1715, opened 2026-03-06
BM-2565: feat(broker): experimental L1Monitor using eth_getBlockReceipts to reduce RPC overhead

jonastheis:
Adds an experimental `--experimental-rpc` flag to the broker that replaces the standard two-service monitoring architecture (ChainMonitorService + MarketMonitor) with a single unified ChainMonitorV2. Instead of polling `eth_getLogs` for market events on every tick, the ChainMonitorV2 uses `eth_getBlockReceipts` to fetch all receipts per block in a single call, then filters for market events locally. This significantly reduces the number of RPC calls — especially on chains like Base where `eth_getLogs` can be expensive or rate-limited.

The ChainMonitorV2 also performs local EIP-1559 gas estimation from receipt data, removing the need for separate `eth_feeHistory` calls that the standard path relies on.

Additionally, this PR introduces a `SequentialFallbackTransport` — a Tower-based Alloy transport layer that tries RPC providers in priority order rather than in parallel (as alloy's built-in FallbackLayer does), minimizing calls to paid/metered fallback endpoints. An `RpcMetricsLayer` is also included for per-method RPC call observability.

The retry and sequential fallback layers are stacked so that on a single RPC failure the sequential fallback immediately tries the next URL, and the outer retry layer only kicks in once all URLs have been exhausted.

```mermaid
graph TB
    subgraph "Standard Path"
        A1[ChainMonitorService] -->|eth_getBlockByNumber| RPC1[RPC Provider]
        A1 -->|eth_feeHistory| RPC1
        A2[MarketMonitor] -->|eth_getLogs per chunk| RPC1
    end

    subgraph "Experimental Path (--experimental-rpc)"
        B1[ChainMonitorV2] -->|eth_getBlockByNumber| RPC2[RPC Provider]
        B1 -->|eth_getBlockReceipts| RPC2
        B1 -.->|extract events + gas locally| B1
    end

    subgraph "Sequential Fallback Transport"
        RPC2 --> Retry[RetryBackoffLayer]
        Retry --> T1[Primary RPC]
        T1 -->|on failure| T2[Fallback RPC 1]
        T2 -->|on failure| T3[Fallback RPC N]
        T3 -->|all failed| Retry
    end
```

## Changes

- **`chain_monitor_v2.rs`** — Unified monitor: polls blocks, fetches receipts via `eth_getBlockReceipts`, decodes market events, estimates gas from receipt data, and tracks chain head — all in a single loop. On startup, uses adaptive log retrieval (`adaptive_get_logs`) with binary-search chunking to auto-discover the max block range accepted by the RPC provider, catching up on missed events for open orders
- **`chain_monitor.rs`** — New `ChainMonitorApi` trait abstracting chain head + gas queries, allowing `ChainMonitorV2` and `ChainMonitorService` to be used interchangeably
- **`sequential_fallback.rs`** — `SequentialFallbackTransport` with health tracking: skips providers after consecutive failures, retries periodically to detect recovery
- **`rpcmetrics.rs`** — Tower layer recording per-method RPC latency and error counts
- **`block_history.rs`** — Ring buffer of recent block fee data for local EIP-1559 gas price estimation
- **`market_monitor.rs`** — Refactored to extract shared event processing functions (`process_log`, `process_order_submitted`, etc.) as `pub(crate)` for reuse by `ChainMonitorV2`
- **`lib.rs`** — Trait-based dispatch between standard and experimental monitor paths; `OrderMonitorApi` trait
- **`broker.rs`** — `--experimental-rpc` CLI flag + provider construction with `SequentialFallbackLayer` and `RpcMetricsLayer`. Adds `--rpc-request-timeout` flag (default 15s) to cut off hanging RPC requests so retry/fallback layers can activate. Defaults to `https://base.gateway.tenderly.co` when `--experimental-rpc` is set with no other RPC URLs provided.

## Evaluation
https://github.com/boundless-xyz/protocol-experiments/tree/main/broker-rpc provides a Python script to analyze the logs of a run of the broker. 

Following is an example of a run with 
```bash
RISC0_DEV_MODE=1 RUST_LOG=INFO,broker::chain_monitor_v2=debug,broker::rpcmetrics=debug PROVER_RPC_URL=https://base.drpc.org \
cargo run --bin broker -- --listen-only --experimental-rpc --set-verifier-address=0x1Ab08498CfF17b9723ED67143A050c8E8c2e3104 2>&1 | tee rpc_metrics.log
```

As can be seen in the results it is possible to run the broker with a free public RPC with the `--experimental-rpc` flag insofar that the broker is always able to recover from any transient, strange, and inconsistent errors from the RPC. However, it is evident that the RPC is fairly unreliable at times and thus for production use a dedicated RPC is recommended. On Base the steady state requires roughly 2 requests per block and should thus reduce RPC cost significantly: 
- `eth_getBlockByNumber` to follow the chain head and get the base fee
- `eth_getBlockReceipts` to get all the receipts and tx fee data

rpc_analysis

```
======================================================================================
  BROKER LOG ANALYSIS
======================================================================================

  ── Log Level Distribution ───────────────────────────────────────────────────
    WARN          159  (0.9%)
    INFO        1,537  (9.1%)
    DEBUG      15,211  (90.0%)

  ── RPC Metrics ──────────────────────────────────────────────────────────────
    Time range  : 2026-03-18 07:49:00.181111 → 2026-03-18 09:19:10.566588
    Duration    : 5410.4s (90.17 min)
    Total reqs  : 6,295
    Req bytes   : 561.7 KB
    Resp bytes  : 2625.66 MB
    Total data  : 2626.21 MB
    Avg req/min : 69.8
    Avg data/min: 29.12 MB

    Method                          Count  Req/min      Req ↑     Resp ↓   Avg Resp   Avg ms   P50 ms   P99 ms
    ------------------------------ ------ -------- ---------- ---------- ---------- -------- -------- --------
    eth_getBlockReceipts             3198     35.5   255.7 KB 2582.26 MB   826.8 KB   1054.6      921     4397
    eth_getBlockByNumber             2937     32.6   248.0 KB   43.30 MB    15.1 KB    623.5      599     1633
    eth_call                          153      1.7    52.9 KB    36.3 KB      243 B    838.6      705     2399
    eth_chainId                         2      0.0       94 B       84 B       42 B    932.0     1470     1470
    eth_getCode                         2      0.0      228 B      436 B      218 B    613.5      616      616
    eth_estimateGas                     2      0.0     4.5 KB       90 B       45 B    771.0      905      905
    eth_getLogs                         1      0.0      242 B    64.0 KB    64.0 KB   1805.0     1805     1805

    High latency calls (>5s): 21 total
    Time                         Method                         Duration       Resp
    2026-03-18 09:08:24.066357   eth_getBlockReceipts            11661ms  1023.9 KB
    2026-03-18 09:10:47.734619   eth_getBlockReceipts            10705ms   929.6 KB
    2026-03-18 08:55:25.291796   eth_getBlockReceipts             9154ms    1.04 MB
    2026-03-18 08:11:43.336403   eth_getBlockReceipts             9039ms   837.8 KB
    2026-03-18 08:13:01.434832   eth_getBlockReceipts             8787ms   801.1 KB
    2026-03-18 08:33:52.311998   eth_getBlockReceipts             8556ms   821.4 KB
    2026-03-18 09:03:43.251144   eth_getBlockReceipts             8073ms   938.6 KB
    2026-03-18 08:58:58.018526   eth_getBlockReceipts             7820ms    1.08 MB
    2026-03-18 09:01:20.168979   eth_getBlockReceipts             7732ms   784.3 KB
    2026-03-18 08:15:47.176581   eth_getBlockReceipts             7259ms   781.5 KB
    2026-03-18 08:28:57.844729   eth_getBlockReceipts             6913ms  1018.1 KB
    2026-03-18 08:21:12.921446   eth_getBlockReceipts             6071ms   632.3 KB
    2026-03-18 09:10:56.232178   eth_getBlockReceipts             6012ms    1.80 MB
    2026-03-18 08:05:50.823871   eth_getBlockReceipts             5940ms   768.6 KB
    2026-03-18 09:07:59.601999   eth_getBlockReceipts             5684ms   946.8 KB
    2026-03-18 08:24:41.028566   eth_getBlockReceipts             5676ms   849.7 KB
    2026-03-18 07:54:57.777164   eth_getBlockReceipts             5642ms    1.13 MB
    2026-03-18 07:52:34.870491   eth_getBlockReceipts             5605ms   856.6 KB
    2026-03-18 09:18:25.938589   eth_getBlockReceipts             5589ms   981.0 KB
    2026-03-18 07:55:15.274738   eth_getBlockReceipts             5478ms    1.13 MB

    eth_getLogs: 1 calls, 0 empty (0.0%), wasted 0 B

  ── RPC Failures (528 failed) ────────────────────────────────────────────────
    Total failed : 528  (8.4% of 6,295 calls)

    Error category                  Count
    ------------------------------ ------
    rpc_error -32601                  435
    HTTP 500                           54
    HTTP 408                           23
    rpc_error -32001                    9
    deserialization                     7

    Method                         Failures    Rate
    ------------------------------ -------- -------
    eth_getBlockReceipts                494   15.4%  rpc_error -32601×435, HTTP 500×30, HTTP 408×22
    eth_getBlockByNumber                 20    0.7%  HTTP 500×17, rpc_error -32001×3
    eth_call                             14    9.2%  HTTP 500×7, rpc_error -32001×6, HTTP 408×1

  ── L1 Monitor ───────────────────────────────────────────────────────────────
    Polling cadence (1,272 polls, 173 skips):
      Avg interval   : 4.21s
      Median interval: 2.02s
      Max gap        : 239.9s
      Gaps > 10s     : 42
      New blocks/poll distribution:
         1 blocks:   744x  (67.7%)
         2 blocks:   264x  (24.0%)
         3 blocks:    35x  (3.2%)
         4 blocks:    10x  (0.9%)
         5 blocks:     6x  (0.5%)
         6 blocks:     5x  (0.5%)
         7 blocks:     3x  (0.3%)
         8 blocks:     2x  (0.2%)
         9 blocks:     3x  (0.3%)
        10 blocks:     1x  (0.1%)
        11 blocks:     1x  (0.1%)
        15 blocks:     1x  (0.1%)
        16 blocks:     3x  (0.3%)
        17 blocks:     1x  (0.1%)
        20 blocks:     1x  (0.1%)
        22 blocks:     1x  (0.1%)
        24 blocks:     1x  (0.1%)
        26 blocks:     1x  (0.1%)
        27 blocks:     1x  (0.1%)
        32 blocks:     1x  (0.1%)
        36 blocks:     1x  (0.1%)
        37 blocks:     1x  (0.1%)
        47 blocks:     1x  (0.1%)
        49 blocks:     1x  (0.1%)
        57 blocks:     1x  (0.1%)
        61 blocks:     1x  (0.1%)
        75 blocks:     2x  (0.2%)
        84 blocks:     1x  (0.1%)
        89 blocks:     1x  (0.1%)
        108 blocks:     1x  (0.1%)
        121 blocks:     1x  (0.1%)
        128 blocks:     1x  (0.1%)
        172 blocks:     1x  (0.1%)
      Significant gaps (>10s) with correlated slow RPCs:
        03-18 07:49:53 → 07:50:24 (31s): (no correlated slow RPC found)
        03-18 07:50:24 → 07:50:42 (17s): (no correlated slow RPC found)
        03-18 07:52:27 → 07:52:37 (10s): eth_getBlockReceipts 5605ms
        03-18 08:00:31 → 08:00:44 (14s): (no correlated slow RPC found)
        03-18 08:09:36 → 08:09:49 (13s): (no correlated slow RPC found)
        03-18 08:09:49 → 08:10:04 (15s): (no correlated slow RPC found)
        03-18 08:10:26 → 08:10:44 (18s): (no correlated slow RPC found)
        03-18 08:10:44 → 08:13:15 (151s): eth_getBlockReceipts 9039ms, eth_getBlockReceipts 8787ms
        03-18 08:13:15 → 08:14:35 (80s): (no correlated slow RPC found)
        03-18 08:14:35 → 08:14:58 (23s): (no correlated slow RPC found)
        03-18 08:14:58 → 08:17:14 (136s): eth_getBlockReceipts 7259ms
        03-18 08:17:14 → 08:18:42 (88s): (no correlated slow RPC found)
        03-18 08:18:42 → 08:20:11 (89s): (no correlated slow RPC found)
        03-18 08:20:11 → 08:24:11 (240s): eth_getBlockReceipts 6071ms
        03-18 08:24:11 → 08:26:41 (150s): eth_getBlockReceipts 5676ms
        … and 27 more gaps

    Receipts per block (2,704 blocks):
      Avg: 201.2  Max: 1314  Min: 115
      101-500   :   2667 blocks (98.6%)
      501-100000:     37 blocks (1.4%)

    Hourly block processing:
      Hour                            Range  Blocks   Blk/min   No-rcpt WARNs
      2026-03-18T07:00   43515422- 43515725     303      30.1
      2026-03-18T08:00   43515726- 43517501    1775      30.0
      2026-03-18T09:00   43517590- 43518101     511      30.1

    Block continuity:
      Polled range        : 43515422 → 43518101 (2680 blocks)
      Unique blocks polled: 1099
      Skip polls (block already seen)          : 173
      Never-observed as latest (gaps): 1581
        43515423-43515437 (15 blocks)
        43515439-43515443 (5 blocks)
        … and 1561 more
      Receipt coverage    : 1098/1099 (99.9%)
      Blocks polled without receipts (1):
        43518101

    On-chain events detected:
      RequestSubmitted      :   204
      RequestLocked         :   371
      RequestFulfilled      :   367

  ── Supervisor Recoveries (99 total) ─────────────────────────────────────────
    OffchainOrderStream: 96
      [ 96x] WebSocket error: Failed to connect to ws endpoint (wss://base-mainnet.boundless.network/ws/v1/orders): https://base-main…
    B-CMV2: 3
      [  3x] [B-CMV2-400] RPC error: Failed to get block receipts for N: Max retries exceeded server returned an error response: erro…

  ── Order Lifecycle ──────────────────────────────────────────────────────────
    Detected  :   204  (on-chain: 204, off-chain: 0)
    Locked    :     0
    Completed :     0
    Skipped   :   235  (over mcycle limit)

  ── CoinGecko Price Oracle ───────────────────────────────────────────────────
    OK      :      0  (0.0%)
    Failed  :     18  (100.0%)

  ── WARN Patterns (60 total) ─────────────────────────────────────────────────
    [   24x] broker::futures_retry: Operation [get_block_receipts] failed: broker::chain_monitor_v2::ChainMonitorV2Err [B-CMV2-N]  RPC error: Failed to get block receipts for N…
    [   18x] boundless_market::price_oracle::composite_oracle: Source CoinGeckoPriceSource(ETH/USD) failed for ETH/USD: [B-PO-N] HTTP error: error decoding response body
    [    6x] boundless_market::price_oracle::composite_oracle: Source ChainlinkSource(ETH/USD) failed for ETH/USD: [B-PO-N] RPC error: server returned an error response: error code -N: incorrect response…
    [    3x] aws_config::imds::region: failed to load region from IMDS err=failed to load IMDS session token: dispatch failure: io error: client error (Connect): tcp connect error…
    [    3x] broker::futures_retry: Operation [get_block_receipts] failed after N retries, returning last error: Some(broker::chain_monitor_v2::ChainMonitorV2Err [B-CMV2-N]  RP…
    [    3x] broker::futures_retry: Operation [get_block_by_number] failed: broker::chain_monitor_v2::ChainMonitorV2Err [B-CMV2-N]  RPC error: Failed to get block header for N:…
    [    1x] aws_runtime::env_config::normalize: section [sso-session 'Jonas Macbook'] ignored; `'Jonas Macbook'` is not a valid identifier
    [    1x] broker: LISTEN-ONLY MODE: Broker will monitor the market and evaluate orders but will NOT lock, prove, or submit. No on-chain transactions will be s…
    [    1x] broker: WARNING: Running the Broker in dev mode does not generate valid receipts. Receipts generated from this process are invalid and should never …

======================================================================================
```


crates/broker/src/l1_monitor.rs
```diff
@@ -0,0 +1,568 @@
...
+
+                        let block_id = BlockId::Number(block_num.into());
+                        let receipts = match self
+                            .any_provider
+                            .get_block_receipts(block_id)
+                            .await
+                        {
+                            Ok(Some(receipts)) => receipts,
+                            Ok(None) => {
+                                // TODO: this should probably be an error and lead to retry
```
capossele: Agree, otherwise the risk is that if there actually were (potentially) events in there we just loose them

crates/broker/src/l1_monitor.rs
```diff
@@ -0,0 +1,568 @@
...
+
+                        // Extract priority fees and update gas estimate.
+                        let priority_fees: Vec<u128> = base_fee_per_gas
+                            .map(|bf| {
+                                receipts
+                                    .iter()
+                                    .map(|r| r.effective_gas_price.saturating_sub(bf))
+                                    .collect()
+                            })
+                            .unwrap_or_default();
```
capossele: Not a biggy but I'm curious if on OP Stack we hit this case often, so just adding a log to keep an eye to it
```suggestion
                        let priority_fees: Vec<u128> = match base_fee_per_gas {
                          Some(bf) => receipts
                              .iter()
                              .map(|r| {
                                  if r.effective_gas_price < bf {
                                      tracing::debug!(
                                          "effective_gas_price {} below base fee {}",
                                          r.effective_gas_price,
                                          bf
                                      );
                                  }
                                  r.effective_gas_price.saturating_sub(bf)
                              })
                              .collect(),
                          None => Vec::new(),
                      };
```

crates/broker/src/l1_monitor.rs
```diff
@@ -0,0 +1,568 @@
...
+            match self.provider.get_logs(&ranged_filter).await {
+                Ok(logs) => {
+                    all_logs.extend(logs);
+                    chunk_found = true;
+                    from = to + 1;
+                }
+                Err(e) => {
+                    if chunk_found {
+                        // Chunk size is already validated; unexpected failure.
+                        return Err(L1MonitorErr::RpcErr(anyhow::anyhow!(
```
capossele: would it makes sense here to retry before returning?

compose.yml
```diff
@@ -345,7 +345,7 @@ services:
     #   target: /target/riscv-guest/guest-assessor/assessor-guest/riscv32im-risc0-zkvm-elf/release/assessor-guest.bin
     environment:
       <<: *broker-environment
-    entrypoint: /app/broker --db-url 'sqlite:///db/broker.db' --config-file /app/broker.toml --bento-api-url http://localhost:8081
+    entrypoint: ["/bin/sh", "-c", "/app/broker --db-url 'sqlite:///db/broker.db' --config-file /app/broker.toml --bento-api-url http://localhost:8081 ${BROKER_EXTRA_ARGS:-}"]
```
capossele: Maybe worth adding `exec` before `/app/broker` so the broker becomes PID 1. Also worth noting that `BROKER_EXTRA_ARGS` is shell-interpreted, so quoting/escaping may be fragile for complex values.

crates/broker/src/tests/e2e.rs
```diff
@@ -166,6 +177,7 @@ fn broker_args(
         rpc_retry_cu: 1000,
         log_json: false,
         listen_only: false,
+        experimental_rpc: false,
```
capossele, line 181: Would it be worth duplicating a test to enable and excercise this new flow?

crates/broker/src/l1_monitor.rs
jonastheis: This is like any other RPC call we make so not sure why we would need explicit retries here? 

All RPC requests are already retried with the `CustomRetryPolicy` anyway: https://github.com/boundless-xyz/boundless/blob/44f945d693d8fb58be14d9b3c1f35fa7b6a411ef/crates/broker/src/bin/broker.rs#L76

crates/broker/src/lib.rs
```diff
@@ -782,40 +779,103 @@ where
             })
             .transpose()?;
 
+        // TODO: create inside the L1 monitor and return
```
willpote: Any reason not to do now?

crates/broker/src/chain_monitor_v2.rs
```diff
@@ -0,0 +1,983 @@
...
+                            let receipts = receipts_result?;
+                            (base_fee, receipts_root, receipts)
+                        };
+
+                        tracing::debug!("Fetched {} receipts for block {block_num}", receipts.len());
+
+                        // TODO: Verify the receipts match the block header commitment.
+                        // This cryptographically ensures the integrity of the receipts
+                        // and makes sure we process all market logs.
+                        // verify_receipts_root(&receipts, _receipts_root)?;
```
willpote, line 514: Plan for this?

crates/broker/src/l1_monitor.rs
```diff
@@ -0,0 +1,983 @@
...
+}
+
+/// Experimental replacement for `ChainMonitorService` + `MarketMonitor`.
+///
+/// A single polling loop drives everything: one `eth_getBlockByNumber(Latest)` per tick,
+/// then one `eth_getBlockReceipts` per unprocessed block to decode market events.
+/// The chain-head atomics are updated on every tick and read synchronously by
+/// [`ChainMonitorApi`] callers (e.g. `OrderPicker`, `OrderMonitor`).
+#[derive(Clone)]
+pub(crate) struct L1Monitor<P, ANP> {
```
willpote: I think this name is confusing as to me L1 = Eth Mainnet. I'd almost prefer we call it `ChainMonitorV2` and rename when we remove the old logic

crates/broker/src/chain_monitor_v2.rs
jonastheis: The main reason I haven't enabled this is twofold:
1. We need to use `op-alloy` (+ potentially other implementations for other chains) or write our own using the `AnyProvider` stuff. Which introduces either more dependencies or more code that needs to be maintained
2. If a chain has a breaking change in terms of receipts we need to support this and also upgrade the broker. Otherwise the broker would become unusable. 

While having this verifiable would be super nice, I'm not sure it's worth the risks.

Now that I'm thinking about it: maybe we can just log an error if there's a mismatch but still continue so the upgrade issue wouldn't completely disable a broker that hasn't upgraded (which would solve 2). Then the question is still how to solve 1.

crates/broker/src/l1_monitor.rs
jonastheis: This seems to be the case for deposit transactions. I left the log in place and added a filter to remove them from the list

crates/broker/src/lib.rs
jonastheis: Done

PR #1716, opened 2026-03-06
BM-2568: move maint_streams to async trigger

zeroecco:


austinabell:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

bento/crates/taskdb/src/lib.rs
```diff
@@ -1447,8 +1447,9 @@ async fn delete_job_test(pool: PgPool) -> sqlx::Result<()> {
     Ok(())
 }
 
-/// Run AUX maintenance to recalculate stream priorities and counters
-pub async fn maint_streams(pool: &PgPool) -> Result<(), TaskDbErr> {
-    sqlx::query("SELECT maint_streams()").execute(pool).await?;
+/// Bulk refresh stream ready/running counters from actual task counts.
+/// Called periodically by aux workers (e.g. every 1–2s) instead of per-row triggers.
+pub async fn refresh_stream_counters(pool: &PgPool) -> Result<(), TaskDbErr> {
+    sqlx::query("SELECT refresh_stream_counters()").execute(pool).await?;
```
austinabell, line 1463: I don't see this function in taskdb, am I missing something?

Also minor point: Is this operation cancellation safe? Just noting that on shutdown this task will be cancelled. (assuming it would be rolled back on cancel, so likely good)

zeroecco: fixed in `3095088e`
cancel should be fine. the task re-writes a few counters. worse case the data is temp wrong.

PR #1874, opened 2026-04-07
BM-2818: Add docker system prune after prover deployment

willpote:
Runs `docker system prune -af` at the end of the prover Ansible role to clean up unused images, containers, networks, and build cache after each deploy. Over time old images pile up on prover hosts — this keeps disk usage in check.

Changes
* New task "Prune unused Docker resources" appended to `ansible/roles/prover/tasks/main.yml`, runs after bento service enablement


capossele (approved):
awesome!

zeroecco (approved):

PR #1876, opened 2026-04-08
BM-2820: Market Indexer Query Optimization

capossele:
### Summary

Comprehensive query optimization for the market indexer, targeting both the per-block indexing path and the hourly/daily aggregation pipeline. Changes include new indexes (migration 51), SQL query rewrites, and JOIN elimination.

### Methodology

- Backfilled the full production dataset (617K proof_requests, 1.3M events, 3.3M transactions) from Base mainnet block 35,060,420 to 44,409,392
- Profiled all queries with `EXPLAIN (ANALYZE, BUFFERS)` before and after changes
- All benchmarks run on the same hardware with default PostgreSQL settings (`work_mem=4MB`)
- All 211 tests pass (181 unit + 30 integration)

### Results

| Query | Before | After | Speedup |
|---|---|---|---|
| **cycle_counts updated_at** (runs every block) | 13.2ms | **0.35ms** | **38x** |
| Median program_cycles (market) | 76.7ms | **3.5ms** | **22x** |
| Median total_cycles (market) | 50.0ms | **3.1ms** | **16x** |
| Requestor fulfilled count | 126ms | **2.6ms** | **48x** |
| Requestor locked count | 121ms | **2.4ms** | **50x** |
| Requestor slashed count | 8.7ms | **0.14ms** | **62x** |
| Requestor program_cycles | 72ms | **4.9ms** | **15x** |
| Requestor total_cycles | 48.9ms | **5.1ms** | **10x** |
| Pricing data (market summary) | 58.7ms | **10.8ms** | **5.4x** |
| Market program_cycles SUM | 76.7ms | **5.5ms** | **14x** |
| Market total_cycles SUM | 50.0ms | **5.1ms** | **10x** |
| Prover program_cycles | 9.2ms | **1.7ms** | **5.4x** |
| Prover total_cycles | 8.6ms | **2.1ms** | **4.1x** |
| Prover fees earned | 15.2ms | **11.2ms** | **1.4x** |
| Lock collateral JOIN | 102ms | **91ms** | **1.1x** |
| Secondary fulfillments | 105ms | **96ms** | **1.1x** |

Queries unchanged (inherent to data access pattern): expiring requests (~17ms), COUNT(DISTINCT prover) all-time (~317ms), COUNT(DISTINCT client) all-time (~73ms).

#### Migration 51: `51_query_optimization_indexes.sql`

**Dropped:**
- `idx_request_status_client_fulfilled_locked_digest` — redundant, superseded by migration 48

**Added (11 indexes):**
- `idx_request_status_fulfilled_program_cycles` — partial index for market cycle sums
- `idx_request_status_fulfilled_total_cycles` — same for total_cycles
- `idx_request_status_prover_fulfilled_mhz` — composite index for prover MHz percentile
- `idx_request_status_client_fulfilled_program_cycles` — requestor cycle queries
- `idx_request_status_client_fulfilled_total_cycles` — same for total_cycles
- `idx_request_status_prover_fulfilled_program_cycles` — prover cycle queries
- `idx_request_status_prover_fulfilled_total_cycles` — same for total_cycles
- `idx_cycle_counts_updated_at` — eliminates per-block seq scan
- `idx_proof_requests_digest_lock_collateral` — covering index for collateral JOIN
- `idx_request_locked_events_digest_prover` — composite index for secondary fulfillments anti-join
- `idx_request_status_client_slashed_at` — partial index for slashed count (tiny, <1% of rows)

#### Query rewrites: `market.rs`, `provers.rs`, `requestors.rs`

**Fetch-all-then-sum-in-Rust to SQL SUM** (12 queries):

Queries that previously fetched thousands of rows into a Rust `Vec` and summed with `checked_add` now use `SUM(CAST(... AS NUMERIC))` in SQL, returning a single row. This eliminates network transfer of up to 28K rows per query.

Affected functions:
- `get_period_total_program_cycles`, `get_period_total_cycles`
- `get_period_all_lock_collateral`, `get_period_locked_and_expired_collateral` (also changed trait signature from `Vec<String>` to `U256`)
- `get_period_prover_total_fees_earned`, `_collateral_locked`, `_collateral_slashed`, `_collateral_earned`, `_program_cycles`, `_total_cycles`
- `get_period_requestor_total_program_cycles`, `_total_cycles`

**JOIN elimination** (4 queries in `requestors.rs`):

Requestor aggregation queries that JOINed event tables with `request_status` to filter by `client_address` now query `request_status` directly, since `fulfilled_at`, `locked_at`, `slashed_at`, and `source` are already denormalized there. Verified 100% data consistency across 617K rows.

- `get_period_requestor_fulfilled_count` — was JOIN with `request_fulfilled_events`
- `get_period_requestor_total_requests_submitted_onchain` — was JOIN with `request_submitted_events`
- `get_period_requestor_total_requests_locked` — was JOIN with `request_locked_events`
- `get_period_requestor_total_requests_slashed` — was JOIN with `prover_slashed_events`

#### Other changes

- `provers.rs`: Added `AND prover_effective_prove_mhz IS NOT NULL` filter to the prover MHz query to match the partial index (safe: Rust code already skips NULL values)
- `market.rs`: Fixed `get_all_request_digests` and `count_request_digests_by_timestamp` to query `proof_requests` instead of `request_status`, so that the backfill `statuses_and_aggregates` mode correctly populates statuses for all known requests (was a bug — only recomputed existing statuses)
- `aggregation/requestors.rs`: Disabled adjusted fulfillment rate computation with explanation (97-100% of requests have unique input data, making the DISTINCT operation produce the same result as a regular COUNT)

### DIFF
--- a/crates/indexer/src/db/market.rs
+++ b/crates/indexer/src/db/market.rs
@@ -941,23 +941,23 @@ pub trait IndexerDb {
         period_end: u64,
     ) -> Result<Vec<LockPricingData>, DbError>;
 
-    /// Gets collateral amounts for all locked requests in the half-open period [period_start, period_end).
+    /// Gets total collateral for all locked requests in the half-open period [period_start, period_end).
     /// Filters by `request_locked_events.block_timestamp` (when the lock event occurred on-chain).
     /// Used for total_collateral_locked metric which tracks all locks regardless of fulfillment.
     async fn get_period_all_lock_collateral(
         &self,
         period_start: u64,
         period_end: u64,
-    ) -> Result<Vec<String>, DbError>;
+    ) -> Result<U256, DbError>;
 
-    /// Gets collateral amounts for locked requests that expired during the half-open period [period_start, period_end).
+    /// Gets total collateral for locked requests that expired during the half-open period [period_start, period_end).
     /// Filters by `request_status.expires_at` (when the request's deadline passed).
     /// Note: These requests may have been locked in an earlier period.
     async fn get_period_locked_and_expired_collateral(
         &self,
         period_start: u64,
         period_end: u64,
-    ) -> Result<Vec<String>, DbError>;
+    ) -> Result<U256, DbError>;
 
     /// Gets the count of requests that expired during the half-open period [period_start, period_end).
     /// Filters by `request_status.expires_at` (when the request's deadline passed).
@@ -3687,25 +3687,20 @@ impl IndexerDb for MarketDb {
         &self,
         period_start: u64,
         period_end: u64,
-    ) -> Result<Vec<String>, DbError> {
-        let rows = sqlx::query(
-            "SELECT pr.lock_collateral
+    ) -> Result<U256, DbError> {
+        let row = sqlx::query(
+            "SELECT COALESCE(LPAD(SUM(CAST(pr.lock_collateral AS NUMERIC))::TEXT, 78, '0'), LPAD('0', 78, '0')) as total
              FROM request_locked_events rle
              JOIN proof_requests pr ON rle.request_digest = pr.request_digest
              WHERE rle.block_timestamp >= $1 AND rle.block_timestamp < $2",
         )
         .bind(period_start as i64)
         .bind(period_end as i64)
-        .fetch_all(&self.pool)
+        .fetch_one(&self.pool)
         .await?;
 
-        let mut results = Vec::new();
-        for row in rows {
-            let lock_collateral: String = row.get("lock_collateral");
-            results.push(lock_collateral);
-        }
-
-        Ok(results)
+        let total_str: String = row.try_get("total")?;
+        padded_string_to_u256(&total_str)
     }
 
     /// Gets collateral amounts for locked requests that expired during the half-open period [period_start, period_end).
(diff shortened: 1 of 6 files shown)


crates/indexer/src/db/market.rs
```diff
@@ -3944,15 +3922,17 @@ impl IndexerDb for MarketDb {
         end_timestamp: u64,
         limit: i64,
     ) -> Result<Vec<(B256, u64)>, DbError> {
+        // Query proof_requests (not request_status) so that backfill can populate
+        // statuses for all known requests, including newly backfilled chain data.
```
willpote, line 3926: Why switch tables here? proof_requests is kind of the raw/unprocessed table, request_status is more the processed + tested table we can trust. Probably no difference for this use-case but generally prefer querying request_status

crates/indexer/src/market/service/aggregation/provers.rs
willpote: Unrelated to your PR, but this SQL should live in teh DB rather than inline, would appreciate if you could move it

willpote (approved):
Nice! Any way we can convert the methodology you used to generate before/after table into some sort of test? 

I'm imagining we could populate our local postgres with a bunch of dummy data, then generate a snapshot file of query performance like we do for gas in Solidity

crates/indexer/src/db/market.rs
capossele: This change fixes a bug in the backfill: when you run `market-indexer-backfill --mode statuses_and_aggregates` after a `chain_data` backfill, the new proof requests exist in `proof_requests` but not yet in `request_status` (that's what the status backfill is supposed to populate). Querying `request_status` here means the backfill only recomputes existing statuses and never creates new ones for the backfilled data. With the old code, `request_status` stayed at 85K after backfilling 617K proof_requests.

crates/indexer/src/market/service/aggregation/provers.rs
capossele: https://github.com/boundless-xyz/boundless/pull/1876/commits/55cbb21a08a9bfa06ccde64be09abe9b6b987199

PR #1878, opened 2026-04-08
BM-2822: [next] Add skip_gas_profitability_check config flag 

willpote:
Changes
* New `skip_gas_profitability_check` field in MarketConfig (defaults to false)
* Bypasses gas-exceeds-maxPrice check in order_pricer when enabled
* Bypasses pre-lock gas-vs-reward check in order_locker when enabled
* Test covering the bypass in order_locker

### DIFF
--- a/crates/broker/src/order_pricer.rs
+++ b/crates/broker/src/order_pricer.rs
@@ -601,6 +601,8 @@ where
         let skip_gas_check =
             self.config.lock_all().map(|c| c.market.skip_gas_profitability_check).unwrap_or(false);
 
+        // TODO: Once we have a price feed for the collateral token in gas tokens, extend
+        // this check to cover lock_expired orders (where the reward is a fraction of the collateral).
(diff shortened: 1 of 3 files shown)


crates/broker/src/order_pricer.rs
```diff
@@ -598,9 +598,10 @@ where
             .await
             .map_err(|err| OrderPricerErr::RpcErr(Arc::new(err.into())))?;
 
-        if order_gas_cost > order.request.offer.maxPrice && !lock_expired {
-            // Cannot check the gas cost for lock expired orders where the reward is a fraction of the collateral
-            // TODO: This can be added once we have a price feed for the collateral token in gas tokens
```
austinabell, line 603: This TODO should likely still be kept, or we open an issue? Probably misplaced though, since it just would be to handle the `lock_expired` case, not within this if statement

austinabell (approved):

PR #1881, opened 2026-04-09
BM-2825: Add nightly CI job to execute latest signal proof request

capossele:
- Adds a scheduled GitHub Actions workflow that runs daily at 3 AM UTC
- Fetches the latest proof request from the signal requestor (`0x734df7809c4ef94da037449c287166d114503198`) via the Base mainnet indexer API
- Executes it locally with `boundless prover execute` and reports cycles + execution time                                                                                         
- Alerts via Slack if the cycle count drifts beyond ±5% of the expected range (`SIGNAL_REQUESTOR_MIN_CYCLES` / `SIGNAL_REQUESTOR_MAX_CYCLES` from `crates/indexer/src/market/service/cycle_counts.rs`)

### DIFF
--- a/.github/workflows/nightly-signal-execute.yml
+++ b/.github/workflows/nightly-signal-execute.yml
@@ -8,6 +8,10 @@ on:
     paths:
       - .github/workflows/nightly-signal-execute.yml
 
+concurrency:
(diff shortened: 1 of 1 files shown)


willpote (approved):
nice

PR #1883, opened 2026-04-09
BM-2831: Use next on nightly and staging resources

zeroecco:


.github/workflows/bento-release.yml
```diff
@@ -60,44 +162,45 @@ jobs:
...
+            ${{ needs.release-metadata.outputs.bundle_archive }}
+            ${{ needs.release-metadata.outputs.bundle_checksum }}
           retention-days: 30
 
   create-release:
     name: Create GitHub Release
-    needs: extract-binaries
-    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/bento-v') && github.event.inputs.dry_run != 'true'
+    needs: [release-metadata, extract-binaries]
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') && github.event.inputs.dry_run != 'true'
```
capossele: This means any tag push (infra-v*, broker-v*, etc.) would attempt to create a release. Instead only those prover related should no? Maybe something like:

```suggestion
    if: >-
    github.event_name == 'push'
    && (startsWith(github.ref, 'refs/tags/bento-v') || startsWith(github.ref, 'refs/tags/prover-v'))
    && github.event.inputs.dry_run != 'true'
```

capossele (approved):
Just a minor

PR #1886, opened 2026-04-10
BM-2835: Deduplicate requests by request_id in indexer list queries

capossele:
## Summary                                                                                                                                                                        
  - When a requestor resubmits a request with the same `request_id` but different parameters (e.g., modified offer), the indexer stores both as separate rows keyed by`request_digest`                                                                                                                                                                  
  - List endpoints (`/requests`, `/requestors/:addr/requests`, `/provers/:addr/requests`) returned both variants, causing the UI to show unfulfilled duplicates even when a fulfilled version exists                                                                                                                                                          
  - Uses a `NOT EXISTS` anti-join to exclude rows where a digest with a more advanced status exists for the same `request_id` (priority: `fulfilled > locked > submitted > expired`)
  - The anti-join preserves the original index-driven `LIMIT` scan — cost is O(page_size) not O(table_size)                                                                         
  - `GET /requests/:request_id` intentionally still returns all variants for diagnostics                                                                                            
                                                                                                                                                                                    
  ## Benchmarks (617K rows, production snapshot)                                                                                                                                    
                                                                                                                                                                                    
  | Query | Original | NOT EXISTS |                                                                                                                                                 
  |-------|----------|------------|                                    
  | By requestor (44K rows) | 0.4 ms | 0.7 ms |                                                                                                                                     
  | Global list (617K rows) | 0.4 ms | 1.2 ms | 
  | By prover (98K rows) | 142 ms | 134 ms |                                                                                                                                                                                                                                                       
                                                                                                                                                                                    
  An earlier `DISTINCT ON` subquery approach was **329 ms / 2,351 ms** for the same queries — rejected because it forces a full scan before pagination. 
  
  **Note**: The list_requests_by_prover query has a pre-existing performance issue (~142ms) unrelated to this PR — the OR across lock_prover_address and fulfill_prover_address prevents efficient index usage, forcing a full scan with row-level filtering. This should be addressed separately (e.g., rewriting as UNION ALL or adding a composite index).

### DIFF
--- a/crates/indexer/src/db/market.rs
+++ b/crates/indexer/src/db/market.rs
@@ -3428,27 +3428,28 @@ impl IndexerDb for MarketDb {
         };
 
         // Deduplicate by request_id: when multiple digests exist for the same
-        // request_id, keep the one with the most advanced status
-        // (fulfilled > locked > submitted > expired), breaking ties by updated_at.
+        // request_id (e.g. resubmission with modified offer), exclude rows where
+        // a digest with a more advanced status exists. Uses NOT EXISTS anti-join
+        // which preserves the original index-driven LIMIT scan.
+        let dedup_clause = "NOT EXISTS (
+                       SELECT 1 FROM request_status rs2
+                       WHERE rs2.request_id = rs.request_id
+                         AND rs2.request_digest != rs.request_digest
+                         AND (CASE rs2.request_status
+                                WHEN 'fulfilled' THEN 1 WHEN 'locked' THEN 2
+                                WHEN 'submitted' THEN 3 WHEN 'expired' THEN 4 ELSE 5 END
+                              <
+                              CASE rs.request_status
+                                WHEN 'fulfilled' THEN 1 WHEN 'locked' THEN 2
+                                WHEN 'submitted' THEN 3 WHEN 'expired' THEN 4 ELSE 5 END)
+                   )";
         let rows = if let Some(c) = &cursor {
             let query_str = format!(
-                "SELECT * FROM (
-                    SELECT DISTINCT ON (request_id) *
-                    FROM request_status
-                    ORDER BY request_id,
-                        CASE request_status
-                            WHEN 'fulfilled' THEN 1
-                            WHEN 'locked'    THEN 2
-                            WHEN 'submitted' THEN 3
-                            WHEN 'expired'   THEN 4
-                            ELSE 5
-                        END,
-                        updated_at DESC
-                ) deduped
-                WHERE ({} < $1 OR ({} = $1 AND request_digest < $2))
-                ORDER BY {} DESC, request_digest DESC
-                LIMIT $3",
-                sort_field, sort_field, sort_field
+                "SELECT rs.* FROM request_status rs
+                 WHERE ({sort_field} < $1 OR ({sort_field} = $1 AND rs.request_digest < $2))
+                   AND {dedup_clause}
+                 ORDER BY {sort_field} DESC, rs.request_digest DESC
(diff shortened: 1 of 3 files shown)


unizarr (approved):

crates/indexer/src/db/market.rs
```diff
@@ -3427,13 +3427,29 @@ impl IndexerDb for MarketDb {
...
+                       SELECT 1 FROM request_status rs2
+                       WHERE rs2.request_id = rs.request_id
+                         AND rs2.request_digest != rs.request_digest
+                         AND (CASE rs2.request_status
+                                WHEN 'fulfilled' THEN 1 WHEN 'locked' THEN 2
+                                WHEN 'submitted' THEN 3 WHEN 'expired' THEN 4 ELSE 5 END
+                              <
+                              CASE rs.request_status
+                                WHEN 'fulfilled' THEN 1 WHEN 'locked' THEN 2
+                                WHEN 'submitted' THEN 3 WHEN 'expired' THEN 4 ELSE 5 END)
```
austinabell, line 3444: This logic seems a bit weird because expired orders would be prioritized over non-expired ones, and then also the issue is that it would not list multiple requests that are unlocked with different request digests, which I think it should? Am I misreading?

crates/indexer/src/db/market.rs
```diff
@@ -3427,13 +3427,29 @@ impl IndexerDb for MarketDb {
...
+                   )";
         let rows = if let Some(c) = &cursor {
             let query_str = format!(
-                "SELECT * FROM request_status
-                 WHERE {} < $1 OR ({} = $1 AND request_digest < $2)
-                 ORDER BY {} DESC, request_digest DESC
+                "SELECT rs.* FROM request_status rs
+                 WHERE ({sort_field} < $1 OR ({sort_field} = $1 AND rs.request_digest < $2))
```
austinabell, line 3449: Also this line _seems_ weird. Why would the `{sort_field} = $1 AND rs.request_digest < $2` be combined in this conditional check?

crates/indexer/src/db/market.rs
capossele: The priority ordering is: `fulfilled(1) > locked(2) > submitted(3) > expired(4)`. Lower number = better. The `NOT EXISTS` clause says "exclude this row if another digest for the same `request_id` has a lower (better) priority number." So:                                                                                
  - `fulfilled` + `submitted` -- keeps fulfilled, hides submitted                                                                                                                   
  - `submitted` + `expired` -- keeps submitted, hides expired (the resubmission is more relevant)
  - two `submitted` digests -- CASE values are equal, neither excludes the other, **both are returned**                                                                             
   
The dedup only kicks in when one digest has strictly progressed further than another. Multiple requests with the same status will still both appear.

crates/indexer/src/db/market.rs
capossele: This is the pre-existing cursor pagination pattern, unchanged by this PR (I only added the table alias `rs.` and the `AND NOT EXISTS` clause). It's keyset pagination: results are ordered by `(sort_field DESC, request_digest DESC)`, and the cursor holds `(timestamp, digest)` of the last row on the previous page. The condition `sort_field < $1 OR (sort_field = $1 AND digest < $2)` selects all rows that come after that position in descending order.

crates/indexer/src/db/market.rs
austinabell: oops, yeah misread the conditional, thought submitted got caught in the else 5

austinabell (approved):

PR #1920, opened 2026-04-14
BM-2884: prover: delay cleanup until after task marked as done

austinabell:
https://github.com/boundless-xyz/boundless/pull/1714 change, but ported over to the new prover.

Issue is that if a job is cancelled or fails in-between the deletion of keys and the task being marked as completed, the prover will be in an inconsistent state where it cannot progress because it has deleted pre-requisite keys.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/prover/crates/workflow/src/tasks/join.rs
+++ b/prover/crates/workflow/src/tasks/join.rs
@@ -5,7 +5,7 @@
 
 use crate::{
     Agent,
-    tasks::{RECUR_RECEIPT_PATH, deserialize_obj, serialize_obj},
+    tasks::{CleanupKeys, RECUR_RECEIPT_PATH, deserialize_obj, serialize_obj},
 };
 use anyhow::{Context, Result};
 use risc0_zkvm::{ReceiptClaim, SuccinctReceipt};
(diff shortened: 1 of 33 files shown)


prover/crates/workflow/src/tasks/join.rs
```diff
@@ -89,20 +89,6 @@ pub async fn join(agent: &Agent, job_id: &Uuid, request: &JoinReq) -> Result<()>
...
-    let cleanup_start = Instant::now();
-    let cleanup_left = agent.hot_delete(&left_path_key).await;
-    let cleanup_right = agent.hot_delete(&right_path_key).await;
-    let cleanup_status =
-        if cleanup_left.is_ok() && cleanup_right.is_ok() { "success" } else { "error" };
-    helpers::record_redis_operation(
-        "unlink",
-        cleanup_status,
-        cleanup_start.elapsed().as_secs_f64(),
-    );
```
capossele, line 102: Should we add this metric collection on the main loop now?

capossele (approved):
Overall good, left just a comment

prover/crates/workflow/src/tasks/join.rs
austinabell: oops, you're absolutely right, good catch. I intended to re-add and forgot

[`b733933` (this PR)](https://github.com/boundless-xyz/boundless/pull/1920/commits/b733933b3d576e6cd933513c0f615e86d6d821a5)

zeroecco (approved):

PR #1934, opened 2026-04-16
BM-2878: standalone execution agent

zeroecco:
Adds a new `crates/executor` crate — an HTTP service that mimics the Bonsai SDK's REST API, execute-only, for cycle counting. Points stock `bonsai-sdk` clients at it with just a URL change.

### Endpoints (wire-compatible with `bonsai-sdk`)
- `GET /version`, `GET /user/quotas`, `GET /health`
- `GET/PUT /images/upload/{image_id}`, `DELETE /images/{image_id}`
- `GET /inputs/upload`, `PUT /inputs/upload/{uuid}`, `DELETE /inputs/{uuid}`
- `GET /receipts/upload`, `PUT /receipts/upload/{uuid}` (assumptions accepted, no-op)
- `POST /sessions/create`, `GET /sessions/{status,logs,stop,exec_only_journal}/{uuid}`
- `POST /snark/create` → 400 (proving not supported)
Non-standard extension: `ProofReq` accepts an optional `"zkvm"` field (defaults to `"risc0"`).

## What this branch does NOT touch
Zero changes outside `crates/executor/`. No contracts, no broker, no CI config, no cross-workspace dependency bumps.

## Test plan
Crate is a separate workspace (each zkVM SDK drags in a huge dep graph), so it doesn't impact the main workspace build:
```bash
cd crates/executor
cargo check                      
cargo check --all-features
End-to-end against a real R0BF binary:

# server
cargo run --release --bin boundless-executor
# client — needs rzup install r0vm on PATH
IMG_ID=$(shasum -a 256 echo.bin | awk '{print $1}')
URL=$(curl -s localhost:8080/images/upload/$IMG_ID | jq -r .url)
curl -s -X PUT --data-binary @echo.bin "$URL"
INP=$(curl -s localhost:8080/inputs/upload)
IN_URL=$(echo "$INP" | jq -r .url); IN_UUID=$(echo "$INP" | jq -r .uuid)
printf 'hello' | curl -s -X PUT --data-binary @- "$IN_URL"
SUID=$(curl -s -X POST localhost:8080/sessions/create -H 'content-type: application/json' \
  -d "{\"img\":\"$IMG_ID\",\"input\":\"$IN_UUID\",\"assumptions\":[],\"execute_only\":true}" | jq -r .uuid)
curl -s localhost:8080/sessions/status/$SUID | jq
# → {"status":"SUCCEEDED","stats":{"total_cycles":32768,"cycles":4234,...}}
Or with the stock bonsai-sdk: set BONSAI_API_URL=http://localhost:8080 + any value for BONSAI_API_KEY.

### CODE REVIEW

File: crates/executor/README.md [hunk 1]
```diff
@@ -0,0 +1,206 @@
...
+
+No changes to the API layer are required — clients pick the backend by name
+via the `zkvm` field in `POST /sessions/create`.
+
+## Bonsai-compatible HTTP API
+
+The server mirrors the Bonsai REST surface. Every endpoint below matches the
+path, method, headers (`x-api-key`, `x-risc0-version`), and JSON shape the
+stock SDK expects. Only the **execute-only** flow is supported — proving
+requests are rejected, and the `/snark/*` endpoints return 400.
```
willpote, line 48: I don't see support for API keys `x-api-key` implemented, we'll need this for indexer => exec calls

crates/executor/src/api.rs
```diff
@@ -0,0 +1,478 @@
...
+/// `PUT /images/upload/{image_id}` — body is the raw ELF/program bytes.
+async fn put_image(
+    State(state): State<AppState>,
+    Path(image_id): Path<String>,
+    body: Bytes,
+) -> StatusCode {
+    state.storage.put_image(image_id, body);
+    StatusCode::OK
+}
+
```
willpote, line 191: Seems like we dropped the image id check from this path? i.e. prover/crates/api/src/lib.rs:387-392

crates/executor/README.md
zeroecco: was gonna have that in the next pr at the ALB level.

crates/executor/src/api.rs
zeroecco: bb7301fbb

willpote (approved):

PR #1940, opened 2026-04-21
[do not merge] BM-2907: order-stream: migrate telemetry to Redshift provisioned (ra3.large)

willpote:
Much cheaper than serverless. A single-node ra3.large cluster runs the same workload for ~$270/mo.

Includes a script to dump the current telemetry tables to S3 that I will run manually before this deploys

Changes
* Swap `redshiftserverless.Namespace` + `Workgroup` for `redshift.SubnetGroup` + `redshift.Cluster` (ra3.large, single-node, encrypted, public, final-snapshot-on-destroy).
* `REDSHIFT_ENDPOINT` output now from `cluster.dnsName` so `apply.sh` treats it as host-only.
* Add `pre-destroy.sh`: UNLOADs typed views to S3 as Parquet before cutover, with an ephemeral IAM policy attached and cleaned up via trap.
* `post-deploy.sh` + `pre-destroy.sh` fall back to env vars for admin/readonly passwords when pulumi config does not have them (dev stacks).
* Migrations and `apply.sh` unchanged — all streaming-ingestion DDL works identically on RA3.

Tested on dev (Base Sepolia): UNLOAD archived 55 Parquet files, `pulumi up` took 16m35s, `post-deploy.sh` applied all 3 migrations on first try, and `telemetry_end_to_end` passed against the new cluster


zeroecco:
approved pending sign off by leadership accepting the tradeoffs

zeroecco (approved):

willpote:
Holding on merging until running the export

PR #1942, opened 2026-04-21
BM-2909: Deploy ansible pipelines from main, drop next trigger

willpote:
Points the Ansible prover CodePipeline at `main` for staging and nightly (previously `next`) and removes the `next` source action and push trigger. Release stays on `main`.

Changes
* Collapses the two Source actions (`GithubMain` + `GithubNext`) back into a single `Github` action on `main`
* Staging and Nightly stages now consume `source_output` from `main`
* Drops the `next` branch push trigger

### DIFF
--- a/.agents/skills
+++ b/.agents/skills
@@ -1,0 +1,1 @@
+../.claude/skills
--- a/.codex/environments/environment.toml
+++ b/.codex/environments/environment.toml
@@ -1,0 +1,5 @@
+version = 1
+name = "boundless"
+
+[setup]
+script = "bash .codex/setup-worktree-unix.sh"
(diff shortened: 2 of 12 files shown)


.agents/skills
```diff
@@ -0,0 +1 @@
+../.claude/skills
```
zeroecco, line 1: newline

zeroecco (approved):
small newline nit, other then that :+1:

.agents/skills
willpote: This is a symlink (`../.claude/skills`) — the "No newline at end of file" warning is intrinsic to symlinks in git. Adding a trailing newline would make the symlink target `../.claude/skills\n`, which wouldn't resolve to a valid path.

PR #1948, opened 2026-04-27
BM-2914: Add alarm for Signal on Sepolia Consensus

capossele:
Mirror the same thresholds and severity used for Signal

### DIFF
--- a/infra/indexer/alarmConfig.ts
+++ b/infra/indexer/alarmConfig.ts
@@ -610,7 +610,7 @@ export const alarmConfig: ChainStageAlarms = {
           ],
           successRate: [],
           expiredRequests: [{
-            description: "greater than or equal to 1 expired orders across 2 hours from signal_requestor",
+            description: "greater than or equal to 1 expired orders across 2 hours from signal_sepolia_requestor",
             severity: Severity.SEV2,
             metricConfig: {
               period: 3600,
--- a/infra/indexer/alarmConfig.ts
+++ b/infra/indexer/alarmConfig.ts
@@ -581,7 +581,7 @@ export const alarmConfig: ChainStageAlarms = {
           submissionRate: [
             {
               description: "no submitted orders in 6 hours from signal_sepolia_requestor",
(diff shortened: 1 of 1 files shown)


infra/indexer/alarmConfig.ts
```diff
@@ -575,6 +575,54 @@ export const alarmConfig: ChainStageAlarms = {
...
+                period: 21600
+              },
+              alarmConfig: {
+                evaluationPeriods: 1,
+                datapointsToAlarm: 1,
+                threshold: 1,
+                comparisonOperator: "LessThanThreshold",
+                treatMissingData: "breaching"
+              }
+            },
```
austinabell: We probably shouldn't have a sev1 for sepolia?

capossele: It's a good question, it's still a prod deployment. Whatever we decide I believe should be applied to both Signal on mainnet consensus as well as on sepolia consensus

capossele: Turned all to sev2

infra/indexer/alarmConfig.ts
```diff
@@ -575,6 +575,54 @@ export const alarmConfig: ChainStageAlarms = {
...
+              },
+              alarmConfig: {
+                evaluationPeriods: 1,
+                datapointsToAlarm: 1,
+                threshold: 1,
+                comparisonOperator: "LessThanThreshold",
+                treatMissingData: "breaching"
+              }
+            }
+          ],
```
austinabell: if both sev2, then seems like these are the same alert but basically duplicated? Doesn't really matter, just noting

austinabell (approved):

infra/indexer/alarmConfig.ts
capossele: I thought the same and kept it, but you are right, I'll drop the 6 hours one

capossele: and I also bumped the number of expired orders to 5 (from 1) before triggering

PR #1949, opened 2026-04-27
BM-2915: refactor(broker): boilerplate reduction & ServiceRunner adoption (PR-1 of 3)

jonastheis:
## Summary
Issue: https://linear.app/boundlessnetwork/issue/BM-2827/refactor-code-structure

This is **PR-1 of 3** for the Broker 2.0 code restructure. The goal of this PR is to land the new boilerplate-reduction infrastructure and migrate every supervised service onto it, **without changing any behaviour**. After this PR the broker still does exactly what it does today, just through cleaner plumbing.

PRs 2 and 3 will follow based on this PR.

## What this PR introduces

Four new mechanisms in `crates/broker/src/`, each landed as additive infrastructure first and adopted incrementally:

- **`BrokerService` trait** (`task.rs`) with a blanket `RetryTask` impl. Service authors now write `async fn run(self, cancel_token) -> Result<…>` instead of the `Box::pin(self.clone() …)` ceremony. The `RetryTask` trait stays around purely to plumb the blanket impl into the existing `Supervisor`.
- **`coded_error_impl!` macro** (`errors.rs`). Generates `impl CodedError` and `impl_coded_debug!` from a compact `Variant => "NNN"` mapping. Supports tuple, struct, and unit variants. Two foreign-type error enums keep their manual impls — comments explain why.
- **`SharedReceiver<T>` + `shared_channel()`** (new `channels.rs`). Bundles the `Arc<Mutex<mpsc::Receiver<T>>>` wrap that several services were doing inline. Adopted in the four channels that feed supervised services (evaluator/committer/locker/pricer inputs).
- **`ServiceRunner` + `Criticality`** (new `service_runner.rs`). Owns the critical/non-critical `JoinSet`s and `CancellationToken`s that `start_service` previously juggled as four separate locals. Each registration is now a single `runner.spawn(svc, Criticality::X, "label")` call. `into_parts()` decomposes back into the existing two-phase shutdown loop unchanged.

Net effect: `start_chain_pipeline` lost ~100 lines and now reads as a flat list of `runner.spawn_in_span(...)` calls. Adding a new service now means writing the service plus a single registration line.

## What's next

- **PR-2 — Structural cleanup of `lib.rs` and the module tree.** Extract `Order` / `OrderStatus` / etc. into `order_types.rs`, `CoreArgs`/`ChainArgs` into `args.rs`, and the `Broker` struct + `start_service` into `broker.rs`. Group shared infrastructure into `shared/` (task, errors, config, prioritization) and utilities into `utils/`. Split the 1830-line `db/mod.rs` into focused files. Pure file moves, no behaviour change.
- **PR-3 — Per-service file splits.** Each supervised service moves into its own directory (`error.rs` / `types.rs` / `service.rs` / etc.) following the [layout](https://docs.google.com/document/d/1vFmXPa59y1HVQabvwqqDmW5stI2LaZjVNvvitQeCVUo/edit?usp=sharing). Largest review surface but minimum semantic risk — strictly mechanical moves.

Each PR will only open after the previous one lands and bakes on `main` to avoid merge-conflict pain.

## How to review

The PR is shaped for **commit-by-commit review** in order. Each commit compiles and passes tests on its own.

Suggested reading path:

1. **First four commits**: the new mechanisms in isolation. Focus on whether each does what it says on the tin (each commit includes its own unit tests).
2. **Two `Clone` derives** on `MarketMonitor` and `OffchainMarketMonitor`: one-line preps so they can satisfy the blanket impl's `Clone` bound.
3. **Service migrations** (one commit per service, ordered simplest first → most complex last): each is the same three-swap pattern — `RetryTask` → `BrokerService`, `impl CodedError` → `coded_error_impl!`, and `Arc<Mutex<Receiver>>` → `SharedReceiver` where applicable. Skim quickly; the diffs are mechanical. The `order_pricer` migration is the most worth a careful look (complex inner loop with child tokens and a `JoinSet`).
4. **Final commit** — `start_service` and `start_chain_pipeline` in `lib.rs` adopt `ServiceRunner`. Compare the before/after of `start_chain_pipeline` to see the cleanup payoff.

Things to watch for:
- Behaviour preservation: each migration commit's title says "migrate" — none should change semantics. Spot-check loop bodies stay identical aside from `service.X` → `self.X` and the dropped outer `Box::pin(self.clone() …)`.
- The `Criticality` mapping in the final commit: the comment in `service_runner.rs` documents the three variants; cross-reference against the original `with_retry_policy(RetryPolicy::CRITICAL_SERVICE)` calls (only `aggregator` and `submitter` had it).
- The `coded_error_impl!` macro is intentionally declarative (not a proc-macro). One small downside: variant names appear in both the `enum` and the macro call. Drift is caught at compile time by Rust's exhaustiveness check.

### DIFF
--- a/crates/broker/src/channels.rs
+++ b/crates/broker/src/channels.rs
@@ -29,9 +29,6 @@ use tokio::sync::{mpsc, Mutex};
 ///
 /// Use together with [`shared_channel`] when the receiver needs to be owned
 /// by a service that the supervisor may clone on restart.
-// Unused until services are migrated to consume `SharedReceiver` directly
-// (Phase 3 of the broker restructure). Kept here as additive infrastructure.
-#[allow(dead_code)]
 pub type SharedReceiver<T> = Arc<Mutex<mpsc::Receiver<T>>>;
 
 /// Creates an [`mpsc`] channel whose receiver is wrapped as a [`SharedReceiver`].
@@ -39,7 +36,6 @@ pub type SharedReceiver<T> = Arc<Mutex<mpsc::Receiver<T>>>;
 /// Equivalent to `mpsc::channel(capacity)` followed by manually wrapping the
 /// receiver in `Arc<Mutex<…>>` — bundled here so service constructors don't
 /// have to repeat the wrapping.
-#[allow(dead_code)]
 pub fn shared_channel<T>(capacity: usize) -> (mpsc::Sender<T>, SharedReceiver<T>) {
     let (tx, rx) = mpsc::channel(capacity);
     (tx, Arc::new(Mutex::new(rx)))
--- a/crates/broker/src/channels.rs
+++ b/crates/broker/src/channels.rs
@@ -1,0 +1,46 @@
+// Copyright 2026 Boundless Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Channel helpers shared across broker services.
+//!
+//! Several services (e.g. `OrderEvaluator`, `OrderPricer`, `OrderCommitter`)
+//! consume from an `mpsc::Receiver` from inside a `Clone` service struct that
+//! is restarted by the supervisor. Cloning a service requires the receiver to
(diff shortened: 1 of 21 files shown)


crates/broker/src/channels.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use tokio::sync::{mpsc, Mutex};
+
+/// A receiver that can be cheaply cloned and shared across `Clone`d services.
+///
+/// Use together with [`shared_channel`] when the receiver needs to be owned
+/// by a service that the supervisor may clone on restart.
+// Unused until services are migrated to consume `SharedReceiver` directly
+// (Phase 3 of the broker restructure). Kept here as additive infrastructure.
+#[allow(dead_code)]
```
capossele: Are you sure this is unused? I see it called throughout the rest of the PR

capossele (approved):

crates/broker/src/channels.rs
jonastheis: removed

PR #1951, opened 2026-04-27
BM-2921: e2e testing examples scripts

austinabell:
This also fixes what seems to be a bug with blake3groth16 such that the bind mount was conflicting with the files downloaded, which at least caused issues with any configuration I tried when running this outside of dev mode.

Another small change is that localnet will use a separate broker.toml file, to avoid conflicts with any other configuration or modifying the same broker.toml file.

Adds a simple example without a pinned lockfile, to ensure no issues with that on latest version, as that had been an issue before.

TLDR this command `just test-e2e-release` will spin up a localnet (dev more or full proving) and run all examples against it in parallel.

### DIFF
--- a/justfile
+++ b/justfile
@@ -482,7 +482,9 @@ job-status job_id:
 # Intended to be run before cutting a release, and from CI via .github/workflows/release-e2e.yml.
 test-e2e-release:
     #!/usr/bin/env bash
-    set -euo pipefail
+    set -uo pipefail
+    mkdir -p target/e2e-logs
+    rm -f target/e2e-logs/*.log
     # Guarantee teardown on any exit path.
     trap 'just _e2e-teardown "$?" || true' EXIT
     just localnet up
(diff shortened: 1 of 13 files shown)


justfile
```diff
@@ -482,3 +483,316 @@ bento-setup:
...
+    just localnet down || true
+
+# Private: run the submit_echo example in either on-chain or off-chain mode.
+# Arg `mode` must be "onchain" or "offchain".
+# Per-example wall-clock timeouts (seconds). Dev mode uses fake proofs; real
+# proving via Bento is dramatically slower (Groth16 + recursion), so we scale up.
+E2E_SUBMIT_ECHO_TIMEOUT := if env_var_or_default("RISC0_DEV_MODE", "1") == "1" { "180" } else { "1800" }
+E2E_EXAMPLE_TIMEOUT := if env_var_or_default("RISC0_DEV_MODE", "1") == "1" { "600" } else { "3600" }
+# Anvil default-mnemonic keys. Slot 0 = deployer, slot 3 = prover — both reserved.
+# See docs/superpowers/plans/2026-04-23-release-e2e-phase-2-3.md for the full allocation.
```
capossele: should be dropped

capossele (approved):

PR #1952, opened 2026-04-27
BM-2918: make ChainMonitorV2 the default, gate legacy path behind --legacy-rpc

capossele:
## Summary

  - Flip the broker chain-monitor default: `ChainMonitorV2` (eth_getBlockReceipts polling loop) is now the default. The legacy `ChainMonitorService` + `MarketMonitor` pair (eth_getLogs) is opt-in via a new `--legacy-rpc` flag.
  - Soft rename: `--experimental-rpc` is kept as a hidden, deprecated, no-op flag so existing invocations that pass it continue to parse without error.
  - The ansible role mirrors the same shape: `prover_broker_experimental_rpc` is preserved as a deprecated no-op; new `prover_broker_legacy_rpc` opts into the legacy path.

  ## Migration notes for prover operators

  Whether you run the broker via `just prover` / `cargo run --bin broker`, the published Docker image, or an ansible-managed deployment, the migration is the same:

  - **No action required** if you previously ran with the broker default (V1) and are comfortable moving to V2 — V2 is now selected automatically on upgrade.
  - **No action required** if you previously passed `--experimental-rpc` (or set `prover_broker_experimental_rpc: true` in ansible) — the flag is redundant now but still accepted; behavior is unchanged (V2 stays selected).
  - **To stay on the old path**: pass `--legacy-rpc` on the broker CLI, or set `prover_broker_legacy_rpc: true` in your ansible inventory.

  ## Follow-up (not in this PR)

  After V2 has run as the default in prod for ~2–4 weeks without regressions: delete `chain_monitor.rs`, `market_monitor.rs`, the `--legacy-rpc` flag, the deprecated `--experimental-rpc` field, the legacy ansible vars, and the `simple_e2e_legacy_rpc` regression test.

### DIFF
--- a/crates/broker/src/bin/broker.rs
+++ b/crates/broker/src/bin/broker.rs
@@ -270,7 +270,7 @@ async fn main() -> Result<()> {
             )?;
 
         let rpc_urls =
-            collect_rpc_urls(args.rpc_url.clone(), args.rpc_urls.clone(), args.experimental_rpc)?;
+            collect_rpc_urls(args.rpc_url.clone(), args.rpc_urls.clone(), args.legacy_rpc)?;
 
         let config = base_config_watcher.config.clone();
         let (provider, any_provider, gas_priority_mode) =
@@ -399,7 +399,7 @@ async fn main() -> Result<()> {
 fn collect_rpc_urls(
     rpc_url: Option<String>,
     rpc_urls: Vec<String>,
-    experimental_rpc: bool,
+    legacy_rpc: bool,
 ) -> Result<Vec<Url>> {
     // Use Vec to preserve insertion order: PROVER_RPC_URL is always inserted first,
     let mut all_rpc_urls: Vec<Url> = Vec::new();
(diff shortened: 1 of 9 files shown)


crates/broker/src/bin/broker.rs
```diff
@@ -439,7 +439,7 @@ fn collect_rpc_urls(
         }
     }
 
-    if all_rpc_urls.is_empty() && experimental_rpc {
+    if all_rpc_urls.is_empty() && !legacy_rpc {
```
jonastheis, line 442: Not related to this PR but how does this work with Taiko? 
2 lines below: `Url::parse("https://base.gateway.tenderly.co")`

Oh I see: with multiple chains we don't call it and we don't allow to set multiple RPCs. https://github.com/boundless-xyz/boundless/blob/2389d4046385d9b035e9b91bdf56ed7142b606e8/crates/broker/src/lib.rs#L1843

jonastheis (approved):

PR #1955, opened 2026-04-28
BM-2924: remove deprecated config in template, warn in generate config

austinabell:
Closes BM-2917

removes the `min_mcycle_price_collateral_token` config from broker files, fixes cli to insert the `expected_probability_win_secondary_fulfillment` if missing (would only modify if already existing), and then adds some backwards compatibility support for certain renamed keys.

### DIFF
--- a/crates/boundless-cli/src/commands/prover/generate_config.rs
+++ b/crates/boundless-cli/src/commands/prover/generate_config.rs
@@ -1079,9 +1079,7 @@ impl ProverGenerateConfig {
                 }
                 for (key, msg) in REMOVED_MARKET_KEYS {
                     if market.contains_key(key) {
-                        display.warning(&format!(
-                            "[market].{key} {msg}; please remove this entry"
-                        ));
+                        display.warning(&format!("[market].{key} {msg}; please remove this entry"));
                     }
                 }
             }
@@ -1238,8 +1236,7 @@ impl ProverGenerateConfig {
                 let probability_value = toml_edit::value(
                     first_chain.expected_probability_win_secondary_fulfillment as i64,
                 );
-                if let Some(item) =
-                    market.get_mut("expected_probability_win_secondary_fulfillment")
+                if let Some(item) = market.get_mut("expected_probability_win_secondary_fulfillment")
                 {
                     *item = probability_value;
                 } else {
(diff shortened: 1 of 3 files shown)


crates/boundless-cli/src/commands/prover/generate_config.rs
```diff
@@ -49,6 +49,28 @@ const PRIORITY_REQUESTOR_LIST_LARGE: &str =
 const PRIORITY_REQUESTOR_LIST_XL: &str =
     "https://requestors.boundless.network/boundless-recommended-priority-list.xl.json";
 
+// Keys under [market] that were renamed and whose backwards-compatible serde
+// aliases have been removed. If found in an existing broker.toml, the broker
+// will fail to deserialize them, so the wizard surfaces these as a migration
```
capossele: I don't think the broker will fail to deserialize, instead they should just be silently dropped, so old keys are ignored, not rejected

crates/boundless-cli/src/commands/prover/generate_config.rs
```diff
@@ -49,6 +49,28 @@ const PRIORITY_REQUESTOR_LIST_LARGE: &str =
...
+// warning. Pairs are (old_name, new_name).
```
capossele:
```suggestion
// Keys under [market] that were renamed. Old names are silently dropped on load, 
// which looks like the user's tuning is in effect when it isn't. 
// The wizard surfaces these so the user notices and renames.
```

crates/boundless-cli/src/commands/prover/generate_config.rs
```diff
@@ -1043,6 +1065,26 @@ impl ProverGenerateConfig {
...
 
+        // Surface any keys whose serde aliases were removed — the broker will
+        // refuse to load these names — and any keys that have been removed
+        // outright and are now silently dropped on load.
+        if matches!(strategy, FileHandlingStrategy::ModifyExisting) {
+            if let Some(market) = doc.get("market").and_then(|v| v.as_table()) {
+                for (old_key, new_key) in RENAMED_MARKET_KEYS {
+                    if market.contains_key(old_key) {
+                        display.warning(&format!(
+                            "[market].{old_key} is no longer accepted; rename to [market].{new_key}"
```
capossele:
```suggestion
                            "[market].{old_key} is no longer used; rename to [market].{new_key}"
```

capossele (approved):

PR #1956, opened 2026-04-28
BM-2925: add delay to submission failure retry, change default to 3

austinabell:
Currently the retry is instant, and can cause issues if the rpc returns inconsistent state.

Adds a configurable delay and also increases the default retry count from 2 to 3. Only 2 total tries for a submission seems low given the cost.

### DIFF
--- a/crates/boundless-market/src/prover_utils/config.rs
+++ b/crates/boundless-market/src/prover_utils/config.rs
@@ -86,7 +86,11 @@ pub mod defaults {
     }
 
     pub const fn max_submission_attempts() -> u32 {
-        2
+        3
+    }
+
+    pub const fn submit_retry_delay_ms() -> u64 {
+        1500
     }
 
     pub const fn reaper_interval_secs() -> u32 {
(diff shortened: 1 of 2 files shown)


crates/boundless-market/src/prover_utils/config.rs
```diff
@@ -1213,4 +1227,11 @@ min_batch_size = 4
         assert_eq!(config.market.min_mcycle_price, Amount::parse("0.1 ETH", None).unwrap());
         assert_eq!(config.batcher.min_batch_size, 4);
     }
+
+    #[test]
+    fn test_batcher_config_defaults() {
+        let config = BatcherConfig::default();
+        assert_eq!(config.max_submission_attempts, 3);
+        assert_eq!(config.submit_retry_delay_ms, 1000);
```
capossele:
```suggestion
        assert_eq!(config.submit_retry_delay_ms, 1500);
```

capossele (approved):

PR #1957, opened 2026-04-28
BM-2926: switch prover2 compose namespace

austinabell:
The issue is that currently bento and prover2 share the same namespace, which means that redis data is overlapping, but also if you run a clean using the new stack, it will clear the previous data. I think this is a footgun that either could cause issues for provers.

This change might cause some complications for our prover, so one way we can avoid this is just changing the redis namespace (and double check nothing else overlapping), but this seems safer to avoid any potential overlap. cc @zeroecco

### DIFF
--- a/ansible/roles/vector/templates/vector.yaml.j2
+++ b/ansible/roles/vector/templates/vector.yaml.j2
@@ -62,7 +62,7 @@ sources:
     mode: scheduled
     scheduled:
       exec_interval_secs: {{ vector_bento_process_interval_secs | default(60) }}
-    command: ["/bin/sh", "-c", "PATH=/usr/bin:/bin docker ps -q --filter label=com.docker.compose.project=bento 2>/dev/null | wc -l"]
+    command: ["/bin/sh", "-c", "PATH=/usr/bin:/bin; b=$(docker ps -q --filter label=com.docker.compose.project=bento 2>/dev/null | wc -l); p=$(docker ps -q --filter label=com.docker.compose.project=prover 2>/dev/null | wc -l); echo $((b + p))"]
 {% endif %}
 
 {% if vector_track_kailua_process | default(false) | bool %}
--- a/compose.blake3-local.yml
+++ b/compose.blake3-local.yml
@@ -15,7 +15,6 @@
 # `just bento up` / `just prover up` auto-include this file whenever
 # BLAKE3_GROTH16_SETUP_DIR is set in the environment.
 
-name: bento
 services:
   gpu_prove_agent:
     volumes:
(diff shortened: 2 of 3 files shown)


capossele:
**1. `compose.blake3-local.yml` silently undoes the namespace separation when used with the new stack.**

The Justfile composes `-f prover-compose.yml -f compose.blake3-local.yml` (in that order) when `PROVER_STACK=experimental` and `BLAKE3_GROTH16_SETUP_DIR` are both set. Compose multi-file merge is last-one-wins for scalar top-level fields like `name:`.
With this PR's `name: prover` in `prover-compose.yml` but `name: bento` still in `compose.blake3-local.yml:18`, the merged project name resolves back to `bento`. Reproduced:

```
$ docker compose -f prover-compose.yml -f compose.blake3-local.yml config | grep ^name:
name: bento   # PR's intent silently undone
```

Fix: drop `name: bento` from `compose.blake3-local.yml`. Override files don't need their own `name:` — they inherit from whichever base file precedes them. Verified:

- prover-compose + fixed blake3-local → `name: prover` ✓
- compose.yml + fixed blake3-local → `name: bento` ✓ (legacy unaffected)

```diff
  # `just bento up` / `just prover up` auto-include this file whenever
  # BLAKE3_GROTH16_SETUP_DIR is set in the environment.

-name: bento
  services:
    gpu_prove_agent:
```

**2. Vector → CloudWatch `bento_containers` metric will go to 0 on prover2 hosts → false alarms.**

`ansible/roles/vector/templates/vector.yaml.j2:65` filters Docker containers by `label=com.docker.compose.project=bento`. After this PR, prover2 containers carry `=prover` instead, so the metric reports 0. That metric feeds CloudWatch alarms in `infra/cw-monitoring/index.ts:162` (and is shown on `components/dashboard.ts:100,101,156,158`) — every prover2 host will trip the "no containers running" alarm even when fully healthy.

The companion `bento_active` metric (line 59, `systemctl is-active bento`) is fine — the systemd unit is still named `bento.service` regardless of stack.

Fix: broaden the filter to count both project labels. Two `docker ps` calls summed (a single `--filter label=...` repeat with values is AND-style, not OR, so this is the reliable form):

```diff
-    command: ["/bin/sh", "-c", "PATH=/usr/bin:/bin docker ps -q --filter label=com.docker.compose.project=bento 2>/dev/null | wc -l"]
+    command: ["/bin/sh", "-c", "PATH=/usr/bin:/bin; b=$(docker ps -q --filter label=com.docker.compose.project=bento 2>/dev/null | wc -l); p=$(docker ps -q --filter label=com.docker.compose.project=prover 2>/dev/null | wc -l); echo $((b + p))"]
```

capossele (approved):

PR #1961, opened 2026-04-29
BM-2929: CLI: improve network discovery and switching

willpote:
Tweaks the boundless CLI so the `networks` subcommand is easier to find and so the network-switching workflow is more discoverable from the welcome screen and per-module `config` output.

## Changes
* Move `networks` to second position (just below `config`) in the `requestor`, `prover`, and `rewards` subcommand lists.
* Replace the per-module `config` tip to cover switching networks
* Expand the welcome-screen tip into two bulleted tips (one for listing, one for switching), each on its own line with a colored bullet, and call out that `--set` accepts a display name or a chain ID.

## Ways to switch / override the active network
* `boundless <module> networks --set "Base Mainnet"` — persists the active network for the module to the config file.
* `boundless <module> networks --set 8453` — same as above using a chain ID.
* `boundless <module> networks --set base-mainnet` — same as above using the kebab-case key.
* `boundless --network "Base Mainnet" <module> <cmd>` — global `--network` flag, applies to a single invocation only and does not persist. Accepts display name, kebab-case key, or chain ID.
* `BOUNDLESS_NETWORK="Base Mainnet" boundless <module> <cmd>` — env-var equivalent of the global flag (also accepts kebab-case key or chain ID).
* `boundless <module> setup` — interactive wizard that writes the active network plus credentials.

### DIFF
--- a/crates/boundless-cli/src/commands/config_display.rs
+++ b/crates/boundless-cli/src/commands/config_display.rs
@@ -48,12 +48,30 @@ impl ModuleType {
         }
     }
 
-    /// Get the help command for this module
-    pub fn help_command(&self) -> &'static str {
+    /// Get the `networks` subcommand for this module
+    pub fn networks_command(&self) -> &'static str {
         match self {
-            ModuleType::Requestor => "boundless requestor --help",
-            ModuleType::Prover => "boundless prover --help",
-            ModuleType::Rewards => "boundless rewards --help",
+            ModuleType::Requestor => "boundless requestor networks",
+            ModuleType::Prover => "boundless prover networks",
+            ModuleType::Rewards => "boundless rewards networks",
+        }
+    }
+
+    /// Example switch-network commands for this module: (display name, chain ID)
+    pub fn switch_network_examples(&self) -> [&'static str; 2] {
+        match self {
+            ModuleType::Requestor => [
+                "boundless requestor networks --set \"Base Mainnet\"",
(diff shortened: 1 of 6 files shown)


crates/boundless-cli/src/commands/config_display.rs
```diff
@@ -151,9 +169,22 @@ pub fn display_not_configured(display: &DisplayManager, module: ModuleType) {
     display.warning(&format!("Not configured - run '{}'", module.setup_command()));
 }
 
-/// Display tip message for a module
-pub fn display_tip(display: &DisplayManager, module: ModuleType) {
-    display.note(&format!("💡 Tip: Run '{}' to see available commands", module.help_command()));
+/// Display tip message for a module suggesting how to switch networks
+pub fn display_tip(_display: &DisplayManager, module: ModuleType) {
```
capossele: nit: could we add a `display.tip_list(header, items)` method on `DisplayManager` that produces the same bulleted output and have `display_tip` call that?

capossele (approved):

crates/boundless-cli/src/commands/config_display.rs
willpote: Done in 837a0c5. Added `TipItem` and `DisplayManager::tip_list(header, items)` in `crates/boundless-cli/src/display.rs`, and reworked both `display_tip` and the welcome-screen tip block to call it. Output is byte-identical to before.

PR #1978, opened 2026-05-07
BM-2950: Emit CommitmentComplete on retries-exhausted path to prevent committer deadlock

capossele:
When the submitter exhausts retries on a permanently-failing batch (`Submitter::process_next_batch`, post-`futures_retry`), it marks each order as Failed in the database but **does not emit a `CommitmentComplete` event** to `proving_completion_rx`. As a result, the orders' slots in `OrderCommitter::in_flight` are never released. Because the committer's capacity gate is global (`if in_flight.len() >= max_concurrent_proofs { return; }` at `order_committer/service.rs:222`), enough leaks deadlock new-order dispatch on **every chain**, not just the chain whose batch failed.

  This patch adds the missing emit, mirroring the pattern already used by `handle_expired_requests_error` (`submitter/service.rs:548`).

  ## Root cause

  `submitter/service.rs:619-627` — the retries-exhausted block updates DB state but is missing the in-memory release:

  ```rust
  for order_id in batch.orders.iter() {
      if let Err(db_err) = self.db.set_order_failure(order_id, "Failed to submit batch").await { ... }
      // <-- no proving_completion_tx.try_send(...) here
  }
  ```

  Other failure paths in the same file (e.g. per-order submission-prep error at :375, all-expired path at :548) do emit the completion event. Only the retries-exhausted path was missing it.

  Symptoms observed in production

  - Staging-nightly broker on Taiko mainnet (chain 167000) stopped locking and fulfilling across all chains after a fulfill batch hit RequestIsNotLockedOrPriced and exhausted retries.
  - Per-chain order_locker tasks were alive (no SupervisorErr::Recover warns, no panicked at, no clean-exit DEBUG logs).
  - order_committer logged at DEBUG every 5 s for ~80 minutes:
  Order committer at capacity, deferring N pending orders pending_commitment_order_count=N in_flight_count=1 max_concurrent_proofs=1
  - Recovery required a full process restart.

### DIFF
--- a/crates/broker/src/submitter/service.rs
+++ b/crates/broker/src/submitter/service.rs
@@ -32,6 +32,7 @@ use boundless_market::{
         FulfillmentDataImageIdAndJournal, FulfillmentDataType, PredicateType,
     },
     selector::{is_blake3_groth16_selector, is_groth16_selector},
+    telemetry::CompletionOutcome,
 };
 use hex::FromHex;
 use risc0_aggregation::{SetInclusionReceipt, SetInclusionReceiptVerifierParameters};
@@ -46,6 +47,7 @@ use tokio::sync::mpsc;
 use crate::{
     config::ConfigLock,
     db::DbObj,
+    errors::{handle_order_failure, BrokerFailure, CodedError},
     is_dev_mode, now_timestamp,
     order_committer::{CommitmentComplete, CommitmentOutcome},
     provers::ProverObj,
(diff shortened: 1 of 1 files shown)


crates/broker/src/submitter/service.rs
```diff
@@ -624,6 +624,14 @@ where
...
                 );
             }
+            // Release the committer's in-flight slot for this order. Without this, a permanent
+            // batch failure leaks a slot in OrderCommitter::in_flight, eventually exhausting
+            // max_concurrent_proofs and silently halting dispatch on all chains.
+            let _ = self.proving_completion_tx.try_send(CommitmentComplete {
+                order_id: order_id.to_string(),
+                chain_id: self.chain_id,
+                outcome: CommitmentOutcome::ProvingFailed,
+            });
```
willpote: We should use the shared helpers that are used in othe rplaces in the code, I think we can use `handle_order_failure` see crates/broker/src/shared/errors.rs:129

Which also ensures we send telemetry

willpote: I think this may be a miss in this file entirely as i see same pattern at L370-380 and L539

willpote (approved):

PR #1986, opened 2026-05-11
BM-2963: Backend provider abstraction for per-order proof routing

capossele:
## Summary

  Introduces the foundational backend abstraction layer used by the broker to route per-order proof post-processing through a backend provider instead of calling RISC Zero-specific helpers directly.

  This PR is the first slice of the backend-decoupling stack. It keeps the broker behavior R0-equivalent today, but establishes the SDK/broker types and dispatch points needed by later PRs for SDK requestor cleanup, order pricing abstraction, and broker aggregation abstraction.

  ## What Changed

  - Added generic market-boundary types:
    - `ProgramId`
    - `PublicOutput`
    - `VerifierSelector`

  - Added backend abstraction traits and registry types:
    - `BackendProvider`
    - `BackendProviderObj`
    - `BackendRegistry`
    - `ComputeClaimDigest`
    - `Sha256ClaimDigest`

  - Added `boundless-r0-backend` with the initial RISC Zero `BackendProvider` implementation:
    - STARK to Groth16 compression
    - STARK to Blake3-Groth16 compression
    - Groth16 / Blake3-Groth16 seal encoding
    - selector-aware claim-digest computation

  - Routed broker per-order paths through `BackendRegistry` / `BackendProvider`:
    - proving compression
    - compressed seal encoding
    - per-order claim digest computation
    - prover lookup by selector for proof lifecycle operations

  - Replaced SDK-boundary R0 digest payloads with `FixedBytes<32>` where appropriate:
    - `Predicate`
    - `FulfillmentData`

  - Added stricter registry semantics:
    - duplicate selector registrations are rejected
    - duplicate selectors inside a single registration are rejected

  - Removed the decorative `boundless-market/risc0` feature from this PR because R0 dependencies remain unconditional until the later SDK dependency cleanup work.

  ## Explicitly Deferred

  This PR does not make the full stack backend-agnostic yet. Follow-up PRs handle:

  - SDK requestor API cleanup and GuestEnv/R0 builder isolation
  - order pricing abstraction
  - broker aggregation and assessor abstraction
  - optional dependency cleanup / true R0-free core crate builds
  - set-inclusion aggregation routing

  ## Notes

  The broker still registers a single RISC Zero backend covering the existing supported selectors. The intent here is not to enable a second backend by itself, but to land the stable routing surface that the next PRs build on.


crates/boundless-market/src/backend_registry.rs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+/// Selector → `BackendEntry` registry. Used by paths that need
+/// composition + on-chain sealing.
+#[derive(Clone, Default)]
+#[non_exhaustive]
+pub struct BackendRegistry {
+    by_selector: HashMap<crate::VerifierSelector, Arc<BackendEntry>>,
+    backends: Vec<Arc<BackendEntry>>,
+}
```
bobbobbio, line 66: perhaps something like https://crates.io/crates/multi_index_map could be useful here

crates/boundless-market/src/backend_registry.rs
```diff
@@ -0,0 +1,211 @@
...
+        prover: ProverObj,
+    ) {
+        let name = name.into();
+        for selector in &selectors {
+            if let Some(_prev) = self.by_selector.insert(*selector, prover.clone()) {
+                tracing::warn!(
+                    selector = %hex::encode(selector.0),
+                    new_backend = %name,
+                    "ProverRegistry: selector already registered; overwriting binding"
+                );
```
bobbobbio, line 185: pretty similar to the code above, also a bit weird to me that the message has a different level of detail. If you're not going to use the binding can just use `.is_some()`

crates/boundless-market/src/claim_digest.rs
```diff
@@ -0,0 +1,84 @@
...
+//! 32-byte commitment a verifier can check. Backend-specific formulas
+//! live in their own crates.
+
+use sha2::{Digest, Sha256};
+
+/// Computes a 32-byte claim digest binding `(program_id, public_output)`.
+///
+/// Implementations are expected to match the formula expected by the
+/// application verifier they are paired with.
+pub trait ComputeClaimDigest: Send + Sync {
```
bobbobbio, line 25: consider using named types instead of bare arrays

crates/boundless-market/src/program.rs
```diff
@@ -0,0 +1,99 @@
...
+use serde::{Deserialize, Serialize};
+
+/// Canonical 32-byte program identity at the market boundary.
+///
+/// For RISC Zero this is the image id. Other backends provide the canonical
+/// 32-byte projection of their native program-identity construction.
+///
+/// Inner field is private; construct via [`Self::new`] / [`From`] and read
+/// via [`Self::as_bytes`] / [`Self::as_fixed_bytes`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
```
bobbobbio, line 30: maybe `#[serde(transparent)]` ?

crates/boundless-market/src/program.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// Public output produced by a zkVM execution (e.g. R0 journal).
+///
+/// Treated as opaque bytes at the generic market boundary; backends interpret
+/// the contents per their own conventions.
+///
+/// Inner field is private; construct via [`Self::new`] / [`From`] and read
+/// via [`Self::as_bytes`] / [`Self::as_alloy_bytes`].
+#[derive(Clone, Debug, PartialEq, Eq, Default, Serialize, Deserialize)]
+pub struct PublicOutput(Bytes);
```
bobbobbio, line 70: maybe also `#[serde(transparent)]`?

crates/r0-backend/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "boundless-r0-backend"
```
bobbobbio, line 2: is there rational to choose `r0` instead of `risc0` here?

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -864,6 +876,32 @@ impl Predicate {
...
+                program_id: &[u8; 32],
+                output_digest: &[u8; 32],
+            ) -> [u8; 32] {
+                let image_id = Digest::from_bytes(*program_id);
+                let journal_digest = Digest::from_bytes(*output_digest);
+                <[u8; 32]>::from(
+                    ReceiptClaim::ok(image_id, MaybePruned::Pruned(journal_digest)).digest(),
+                )
+            }
+        }
```
bobbobbio, line 902: consider just moving to outer scope for indentation relief

crates/broker-stress/src/main.rs
```diff
@@ -91,7 +91,10 @@ async fn request_spawner<P: Provider>(
                 ctx.customer_signer.address(),
                 ctx.customer_market.index_from_nonce().await?,
             ),
-            Requirements::new(Predicate::prefix_match(Digest::from(ECHO_ID), Bytes::default())),
+            Requirements::new(Predicate::prefix_match(
+                <[u8; 32]>::from(Digest::from(ECHO_ID)),
```
bobbobbio, line 95: IMO all these places that are `<[u8; 32]>::from` are sort of unfortunate to read. If you had created your own type like `Digest2` (but a better name) you could have made these functions be `impl Into<Digest2>` and implemented `From<risc0_zkp::Digest> for Digest2` and there would have been no need to adjust all these callsites

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -730,27 +727,27 @@ pub enum PredicateError {
 /// A Predicate is a function over the claim that determines whether it meets the clients requirements.
 pub enum Predicate {
     /// The image id and digest must match the provided image id and journal digest.
-    DigestMatch(Digest, Digest),
```
bobbobbio, line 733: I think it a bit of a shame to loose the more specific type, wish we were replacing with our own digest type, here and everywhere else.

crates/broker/src/utils/storage.rs
```diff
@@ -110,7 +110,10 @@ pub async fn upload_image_uri(
     let predicate = Predicate::try_from(request.requirements.predicate.clone())
         .with_context(|| format!("Failed to parse predicate for request {:x}", request.id))?;
 
-    let image_id_str = predicate.image_id().map(|image_id| image_id.to_string());
+    // FixedBytes<32>::to_string produces "0x..." (66 chars). The prover and
+    // R0's Digest::from_hex expect plain hex (64 chars), so encode the bytes
+    // directly to keep the round-trip working.
+    let image_id_str = predicate.image_id().map(|image_id| hex::encode(image_id.as_slice()));
```
bobbobbio, line 116: here too having our own type instead of using `FixedBytes` directly would have avoided having to change this

crates/boundless-market/src/prover_utils/prover.rs
```diff
@@ -160,6 +160,28 @@ pub trait Prover {
         &self,
         proof_id: &str,
     ) -> Result<Option<Vec<u8>>, ProverError>;
+
+    /// Compute the 32-byte program identity from the ELF bytes the prover
+    /// would upload as an image. Backends that derive program identity by
+    /// hashing the binary in their own format override this.
+    ///
+    /// Default uses the RISC Zero formula.
+    fn compute_image_id_from_bytes(&self, elf: &[u8]) -> Result<[u8; 32], ProverError> {
```
bobbobbio, line 169: here too would be nice to have a type instead of bare array

crates/boundless-market/src/selector.rs
```diff
@@ -235,6 +235,21 @@ impl SupportedSelectors {
...
 
+/// Classify a selector into a [`ProofType`].
+///
+/// Returns [`ProofType::Groth16`] or [`ProofType::Blake3Groth16`] for the
+/// matching dedicated forms, and [`ProofType::Any`] for the unspecified
+/// selector or any set-inclusion / pass-through selector.
+pub fn classify_selector(selector: FixedBytes<4>) -> ProofType {
+    if is_groth16_selector(selector) {
+        ProofType::Groth16
+    } else if is_blake3_groth16_selector(selector) {
```
bobbobbio, line 246: I saw another place that had code very similar to this, except it checked for a "fake" blake3 groth16 selector, wonder why that separate and why this doesn't have to check for that

crates/broker/src/proving/service.rs
```diff
@@ -181,19 +183,30 @@ impl ProvingService {
...
 
+                // Image upload, input upload, prove_stark, and the later
+                // wait_for_stark all run on the same registered prover.
+                let selector = order.request.requirements.selector;
+                let order_prover = self
+                    .backends
+                    .find(selector)
+                    .map(|entry| entry.prover.clone())
+                    .ok_or_else(|| {
+                    anyhow!("no backend registered for selector {selector:?} (order {order_id})")
```
bobbobbio, line 194: this indentation seems strange, is this a rustfmt bug or just a place that was forgotten?

crates/broker/src/submitter/service.rs
```diff
@@ -266,8 +266,19 @@ where
...
+                let order_selector = order_request.requirements.selector;
+                let order_journal_prover = self
+                    .backends
+                    .find(order_selector)
+                    .map(|entry| entry.prover.clone())
+                    .ok_or_else(|| {
+                        anyhow!(
+                            "no backend registered for selector {order_selector:?} (order {order_id})"
+                        )
+                    })?;
```
bobbobbio, line 280: second chunk of same code, maybe this is a better as a function on `self`

crates/boundless-market/src/backend_registry.rs
```diff
@@ -73,29 +91,45 @@ impl BackendRegistry {
...
+    /// [`Self::try_register`] to handle duplicate selectors explicitly.
     pub fn register(&mut self, registration: BackendEntry) {
-        let registration = Arc::new(registration);
+        self.try_register(registration).expect("duplicate selector in BackendRegistry")
+    }
+
+    /// Register a backend.
+    ///
+    /// Duplicate-selector registrations are configuration bugs and are
+    /// rejected before any registry state is mutated.
```
bobbobbio, line 115: somewhat subtle point, I would have just used `HashMap::entry` and not considered this, might be good to add `BackendRegistry` unit test showing it doesn't add any backends if one of them conflicts.

crates/boundless-market/src/backend_registry.rs
```diff
@@ -205,32 +223,65 @@ impl ProverRegistry {
...
         selectors: Vec<crate::VerifierSelector>,
         prover: ProverObj,
     ) {
+        self.try_register(name, selectors, prover).expect("duplicate selector in ProverRegistry")
+    }
+
+    /// Register a prover for a set of selectors.
+    ///
+    /// Duplicate-selector registrations are configuration bugs and are
+    /// rejected before any registry state is mutated.
```
bobbobbio, line 257: same comment about the backend registration

crates/boundless-market/src/backend_registry.rs
```diff
@@ -132,7 +135,15 @@ impl BackendRegistry {
...
     /// rejected before any registry state is mutated.
     pub fn try_register(&mut self, registration: BackendEntry) -> Result<(), BackendRegistryError> {
+        let mut seen = HashSet::new();
         for selector in &registration.selectors {
+            if !seen.insert(*selector) {
+                return Err(BackendRegistryError::DuplicateSelector {
+                    selector: hex::encode(selector.0),
+                    existing_backend: registration.name.clone(),
+                    new_backend: registration.name.clone(),
+                });
```
bobbobbio, line 145: this originally missing edge case increases my desire for unit tests

crates/boundless-market/src/backend_registry.rs
bobbobbio, line 158: probably would be good to assert about the return value here

bobbobbio, line 300: probably good to assert about the return value here

bobbobbio (commented):
Reading through was educational for me, still understanding the architecture.

From a pure code point of view, my main feedback boils down to
- would love some wrapper types instead of bare [u8; 32] arrays.
- would love some unit tests for the new BackendRegistry type

have other little comments on all the various commits

crates/boundless-market/src/contracts/mod.rs
jonastheis: Agree, we should introduce a type here. cleaner and should be more convenient to use with From impl

crates/indexer/tests/market/execution.rs
```diff
@@ -91,7 +92,10 @@ fn test_config(uri: String, max_iterations: u32) -> IndexerServiceExecutionConfi
 fn generate_request(id: u32, client: &Address) -> ProofRequest {
     ProofRequest::new(
         RequestId::new(*client, id),
-        Requirements::new(Predicate::prefix_match(ECHO_ID, Bytes::default())),
+        Requirements::new(Predicate::prefix_match(
```
jonastheis, line 95: Would be better to re-export that as a `ProgramId` or otherwise make sure that it's the correct type so we don't need to import `risc0_zkvm` in so many places.

crates/boundless-market/src/prover_utils/prover.rs
```diff
@@ -160,6 +160,36 @@ pub trait Prover {
         &self,
         proof_id: &str,
     ) -> Result<Option<Vec<u8>>, ProverError>;
+
```
jonastheis, line 163: Do we need to adjust the prover trait? This still seems to have risc0 types and it would be great to use the ProgramId here instead of a str (and then internally can convert to what bento needs I guess)

crates/boundless-market/src/lib.rs
```diff
@@ -72,6 +72,27 @@ pub mod input;
 #[cfg(not(target_os = "zkvm"))]
 pub use input::{GuestEnv, GuestEnvBuilder};
 
+/// Generic program-identity and public-output types at the market boundary.
+pub mod program;
+pub use program::{ProgramId, PublicOutput};
+
+/// 4-byte on-chain verifier selector. Routes to the verifier contract that
+/// validates the proof seal.
+pub type VerifierSelector = alloy_primitives::FixedBytes<4>;
```
jonastheis, line 81: Why is this not defined in `selector.rs`? And maybe instead of a type alias we can make it its own type and define `From` for easy conversion

crates/boundless-market/src/selector.rs
```diff
@@ -235,6 +235,21 @@ impl SupportedSelectors {
     }
 }
 
+/// Classify a selector into a [`ProofType`].
+///
+/// Returns [`ProofType::Groth16`] or [`ProofType::Blake3Groth16`] for the
+/// matching dedicated forms, and [`ProofType::Any`] for the unspecified
+/// selector or any set-inclusion / pass-through selector.
+pub fn classify_selector(selector: FixedBytes<4>) -> ProofType {
```
jonastheis, line 243: Why doesn't this take the `VerifierSelector` type or is a fn on the type?

crates/broker/src/broker.rs
```diff
@@ -766,6 +769,22 @@ impl Broker {
...
+        // through it. Pricing and preflight remain tied to the scalar
+        // `prover` so the supported set and the preflight prover stay in
+        // lockstep.
+        let r0_provider: boundless_market::backend_provider::BackendProviderObj = Arc::new(
+            RiscZeroBackend::new(prover.clone(), aggregation_prover.clone(), is_dev_mode()),
+        );
+        let supported_selectors = SupportedSelectors::default();
+        let backends = BackendRegistry::with_backend(BackendEntry::new(
+            "risc0",
+            supported_selectors.selectors.keys().copied().collect(),
```
jonastheis, line 783: A backend should know its supported selectors no? (ie `selectors(&self) -> &[VerifierSelector]` or something on the trait)

I think this should just take the backend itself and internally it can maybe implement From trait or simply call a method on the backend trait that returns an entry with all that information

crates/boundless-market/src/backend_registry.rs
```diff
@@ -0,0 +1,308 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Selector → backend registries used by broker pricing & proving paths.
+//!
+//! - `ProverRegistry`: selector → `Prover`. Used for cycle-counting /
+//!   preflight only (SDK pricing + broker `OrderPricer`).
+//! - `BackendRegistry`: selector → `BackendEntry` (`Prover` +
+//!   `BackendProvider`). Used by broker proving / aggregator / submitter
+//!   paths that need composition + on-chain sealing.
```
jonastheis, line 21: I'm not sure I understand the exact difference between prover and backend registry. Could `prover` be seen as an implementation detail of a `Backend(Entry)`?

crates/boundless-market/src/backend_registry.rs
```diff
@@ -0,0 +1,308 @@
...
+        }
+        supported
+    }
+}
+
+/// Selector → `Prover` registry. Used by paths that only need
+/// cycle-counting / preflight (SDK pricing, broker `OrderPricer`).
+#[derive(Clone, Default)]
+#[non_exhaustive]
+pub struct ProverRegistry {
```
jonastheis, line 192: Why do we need this?

crates/boundless-market/src/backend_registry.rs
```diff
@@ -0,0 +1,308 @@
...
+        let mut seen = HashSet::new();
+        for selector in &registration.selectors {
+            if !seen.insert(*selector) {
+                return Err(RegistryError::DuplicateSelector {
+                    selector: hex::encode(selector.0),
+                    existing_registration: registration.name.clone(),
+                    new_registration: registration.name.clone(),
+                });
+            }
+            if let Some(prev) = self.by_selector.get(selector) {
```
jonastheis, line 129: doesn't this also handle the case above?

crates/broker/src/proving/service.rs
```diff
@@ -84,12 +89,20 @@ impl ProvingService {
...
         snark_proof_id: Option<String>,
+        selector: FixedBytes<4>,
     ) -> Result<OrderStatus> {
-        let proof_res = self
-            .prover
+        // The prover that ran `prove_stark` for this selector also owns the
+        // wait-for-stark cache for this `proof_id`.
+        let stark_prover = self
+            .backends
+            .find(selector)
```
jonastheis, line 102: This is repeated a bunch of times: why not introduce something like `BackendRegistry::require_entry(selector) -> Result<&BackendEntry, RegistryError>` to have a consistent error and easier usage.

crates/boundless-market/src/backend_registry.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+}
+
+/// Selector → `BackendEntry` registry. Used by paths that need
+/// composition + on-chain sealing.
+#[derive(Clone, Default)]
+#[non_exhaustive]
+pub struct BackendRegistry {
+    by_selector: HashMap<crate::VerifierSelector, Arc<BackendEntry>>,
+    backends: Vec<Arc<BackendEntry>>,
```
jonastheis, line 86: Is this `Vec` really necessary?

crates/boundless-market/src/backend_registry.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+impl BackendEntry {
+    /// Construct a `BackendEntry`. Required because the struct is
+    /// `#[non_exhaustive]` (out-of-crate struct-literal construction is
+    /// blocked).
+    pub fn new(
+        name: impl Into<String>,
+        selectors: Vec<crate::VerifierSelector>,
+        prover: ProverObj,
```
jonastheis, line 73: Why is the prover also here? I thought the backend already has the prover? Probably related to my question above how prover and backend are related/different.

crates/boundless-market/src/contracts/mod.rs
```diff
@@ -864,6 +876,32 @@ impl Predicate {
         }
         claim_digest_data.or(claim_digest_predicate)
     }
+
+    /// Convenience wrapper over [`Predicate::eval_with`] using the RISC Zero
+    /// claim-digest formula (`ReceiptClaim::ok(image_id, journal).digest()`).
+    /// For non-R0 backends, call [`Predicate::eval_with`] directly with the
+    /// appropriate formula.
+    pub fn eval(&self, fulfillment_data: &FulfillmentData) -> Option<FixedBytes<32>> {
```
jonastheis, line 884: Do we need this or is this just an intermediary helper? this duplicates code and adds risc0 dependencies here. We should add a TODO and eventually remove this and only use `eval_with`

crates/boundless-market/src/backend_provider.rs
```diff
@@ -0,0 +1,93 @@
...
+    async fn encode_seal(
+        &self,
+        compressed_proof_id: &str,
+        selector: crate::VerifierSelector,
+    ) -> Result<Vec<u8>, BackendProviderError>;
+
+    /// Compute the 32-byte claim digest binding `(program_id, public_output)`
+    /// under the given selector. Different selectors under the same
+    /// backend may use different claim-digest formulas.
+    fn compute_claim_digest(
```
jonastheis, line 84: I think we could simplify and abstract away a bunch of stuff if we do 
```
fn claim_digest(&self, selector: VerifierSelector) -> Result<&dyn ComputeClaimDigest, BackendProviderError>;
```

crates/broker/src/proving/service.rs
```diff
@@ -650,13 +673,87 @@ mod tests {
...
+        BackendRegistry::with_backend(BackendEntry::new(
+            "risc0",
+            supported.selectors.keys().copied().collect(),
+            prover,
+            provider,
+        ))
+    }
+
+    #[test]
+    fn registry_covers_every_supported_selector() {
```
jonastheis, line 705: I feel like we should have some mock tooling for testing for all of these types in the corresponding crate. otherwise we will repeat the same thing countless times and also pollute with specific implementations/dependencies

crates/broker/src/proving/service.rs
```diff
@@ -105,46 +118,30 @@ impl ProvingService {
...
+            // Per-selector composition runs through the registered
+            // `BackendProvider`. The provider owns the STARK to SNARK path
+            // for its selector and verifies the compressed receipt before
+            // returning the id.
+            let provider = self
+                .backends
+                .find(selector)
+                .map(|entry| entry.provider.clone())
+                .ok_or_else(|| anyhow!("no backend registered for selector {selector:?}"))?;
+            let program_id = ProgramId::from(<[u8; 32]>::from(
```
jonastheis, line 130: Why is this not part of the backend?

crates/r0-backend/src/risc_zero.rs
```diff
@@ -0,0 +1,388 @@
...
+        &self,
+        selector: boundless_market::VerifierSelector,
+        program_id: &ProgramId,
+        public_output: &PublicOutput,
+    ) -> Result<[u8; 32], BackendProviderError> {
+        Self::dispatch_claim_digest(selector, program_id, public_output)
+    }
+}
+
+impl RiscZeroBackend {
```
jonastheis, line 228: why 2 impl sections?

capossele:
Closing for now and will reopen with a better design

PR #1988, opened 2026-05-11
Add backend-aware request pricing

capossele:
## Summary

  Adds a backend-aware request-pricing abstraction for SDK request building.

  This PR introduces `RequestPricer` and `BackendRequestPricerFactory`, plus an R0 implementation in `boundless-r0-backend`. The standard request builder can now run its pricing check through a backend-provided pricer instead of being tied directly to the inline R0 preflight path.

  This is stacked on the core backend-routing PR. It does not attempt the broader SDK redesign; it only isolates request-pricing/preflight cost evaluation behind a small API.

  ## What Changed

  - Added `RequestPricer`, `RequestPricingResult`, and `RequestPricingContext<P>` to `boundless-market`.
  - Added `BackendRequestPricerFactory<P>` for backend-specific pricer construction.
  - Added `RiscZeroRequestPricer` and `RiscZeroRequestPricingBackend` in `boundless-r0-backend`.
  - Added `Client::with_request_pricer(...)` for direct pricer injection.
  - Added `Client::with_request_pricing_backend::<B>()` for backend factory wiring.
  - Updated `StandardRequestBuilder` to dispatch pricing checks through an installed `RequestPricer`.
  - Kept the existing R0 pricing/preflight path as the fallback when no custom pricer is installed.
  - Switched request-pricing internals from `LocalExecutor` to `ProverObj` so the request builder and pricer share the same preflight cache.
  - Added selector to the preflight cache key to avoid cross-selector cache collisions.

  ## Notes

  `RequestPricingContext<P>` is `#[non_exhaustive]` so future backend-specific context can be added without forcing another positional-argument API break.


crates/boundless-market/src/request_builder/mod.rs
```diff
@@ -202,6 +202,15 @@ pub struct StandardRequestBuilder<P = DynProvider, U = StandardUploader, D = Sta
     /// If `None`, falls back to checking the `BOUNDLESS_IGNORE_PREFLIGHT` environment variable.
     #[builder(setter(into, strip_option), default)]
     pub skip_preflight: Option<bool>,
+
+    /// Backend-supplied request pricer for the pricing-check step.
+    ///
+    /// When `Some`, `run_pricing_check` dispatches through it. When `None`,
+    /// the check runs the inline R0 `requestor_order_preflight` path keyed
+    /// off the preflight layer's executor.
```
jonastheis, line 210: Why leave the fallback? Doesn't that mean the trait is not really doing what it should do ie hide/abstract everything R0 related behind the trait?

I think we should set this somewhere upstream to the R0 *implementation* in case it is not set but remove all the inline code.

crates/boundless-market/src/request_builder/mod.rs
```diff
@@ -246,7 +270,7 @@ where
 
         if let Err(e) = requestor_order_preflight(
             request.clone(),
-            self.preflight_layer.executor_cloned(),
+            self.preflight_layer.prover_obj_cloned(),
```
jonastheis, line 273: I don't understand why we leave this fallback path. Now this code is R0 specific AND duplicated with what's in the `RiscZeroRequestPricer` implementation in https://github.com/boundless-xyz/boundless/blob/817753a7019191cec24f44547615b10da4f1f5ef/crates/r0-backend/src/order_pricer.rs#L68

crates/r0-backend/src/order_pricer.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+#[async_trait]
+impl<P> RequestPricer for RiscZeroRequestPricer<P>
+where
+    P: Provider<Ethereum> + 'static + Clone,
+{
+    async fn price(&self, request: &ProofRequest) -> anyhow::Result<RequestPricingResult> {
+        let work_units = requestor_order_preflight(
```
jonastheis, line 68: Why is this method not moved into this crate?

crates/boundless-market/src/order_pricer.rs
```diff
@@ -0,0 +1,176 @@
...
+
+use crate::{
+    contracts::ProofRequest, price_oracle::PriceOracleManager, price_provider::PriceProviderArc,
+    prover_utils::prover::ProverObj,
+};
+
+/// Environment available to backend request-pricer factories.
+#[derive(Clone)]
+#[non_exhaustive]
+pub struct RequestPricingContext<P> {
```
jonastheis, line 36: I'm a bit confused as to how `RequestPricingContext` and `OrderPricingContext` are related.

jonastheis (commented):
After reviewing this PR and #1986 I was thinking a bit more about the current structure and with Claude's help came up with this. Let's discuss later sync.

# Backend trait map — current state and proposed shape

Written as part of reviewing PRs #1986 (backend-routing) and #1988 (request-pricing).
Scope: the traits that define what a "backend" (zkVM provider) owns in Boundless, how
they interact today, and a target shape that consolidates them.

## 1. Current state

```
═══════════════════════════════════════════════════════════════════════════════
                          boundless-market (SDK / core)
═══════════════════════════════════════════════════════════════════════════════

           ┌───────────────────────────────────────────────────────────┐
           │ trait ComputeClaimDigest                                  │
           │   fn compute(pid, out)         -> [u8;32]                 │
           │   fn compute_from_output_digest(pid, hash) -> [u8;32]     │
           └─┬─────────────────────────────────────────────────────────┘
             │
             │ has DEFAULT impl:  Sha256ClaimDigest  (unused)
             │
             │ ◄─ used by Predicate::eval_with<C>  (generic) ✓
             │ ◄─ DUPLICATED inline inside Predicate::eval()  ⚠ R0 LEAK
             │ ◄─ owned standalone by Submitter   ⚠ (passed in via ctor,
             │                                       not pulled from registry)
             │ ◄─ called internally by BackendProvider::compute_claim_digest
             │     (R0 backend dispatches selector → which CCD impl to use,
             │      but doesn't EXPOSE that choice)

           ┌───────────────────────────────────────────────────────────┐
           │ trait Prover                                              │
           │   has_image / upload_image / upload_input                 │
           │   preflight / prove_stark / wait_for_stark                │
           │   compress / compress_blake3_groth16                      │
           │   get_receipt -> risc0_zkvm::Receipt        ⚠ R0 LEAK     │
           │   get_journal / get_compressed_receipt                    │
           │   compute_image_id_from_bytes  { default = R0 }  ⚠        │
           │   decode_input_bytes           { default = R0 }  ⚠        │
           └─┬─────────────────────────────────────────────────────────┘
             │
             │ ◄─ impl LocalExecutor    (preflight ONLY; panics on prove ⚠)
             │ ◄─ impl Bonsai           (full-featured, in broker)
             │ ◄─ impl DefaultProver    (dev mode, in broker)

           ┌───────────────────────────────────────────────────────────┐
           │ trait BackendProvider                                     │
           │   async compress_proof(pid, primitive_id, selector)       │
           │   async encode_seal(compressed_id, selector)              │
           │   fn   compute_claim_digest(selector, pid, out) -> [u8;32]│   <── opaque
           └─┬─────────────────────────────────────────────────────────┘       returns
             │                                                                 bytes,
             │ ◄─ impl RiscZeroBackend  (in r0-backend)                        not a
             │       holds:  prover: ProverObj,                                trait
             │               snark_prover: ProverObj,        ⚠ same Arc       object
             │               dev_mode: bool                    also in
             │                                                 BackendEntry
             │
             └─ INTERNALLY chooses between RiscZeroClaimDigest vs
                Blake3Groth16ReceiptClaim per selector — selection
                lives inside the impl, not returned to caller


  ┌─────────────────────────────────────────────────────────────────────┐
  │ struct BackendEntry { name, selectors[], prover, provider }         │
  │ struct BackendRegistry { HashMap<VerifierSelector, Arc<Entry>> }    │
  │ struct ProverRegistry  (DORMANT — no consumer)                      │
  └─────────────────────────────────────────────────────────────────────┘
            ▲                                            ▲
            │ find(selector) -> Option<&BackendEntry>    │
            │   (callers .ok_or_else(...) at every site) │
            │                                            │
   ┌────────┴────────┐                            ┌──────┴──────────┐
   │ broker          │                            │ Submitter ALSO  │
   │   ProvingService                             │ takes:          │
   │   Aggregator                                 │  assessor_claim_│
   │   Submitter      ──── per-order ───┐         │  digest:        │
   └─────────────────┘                  │         │  Arc<dyn CCD>   │
                                        │         │  (standalone,   │
                                        │         │   not via       │
                                        │         │   registry) ⚠   │
                                        │         └─────────────────┘
                                        ▼
                                  ┌──────────────────────────────────┐
                                  │ trait OrderPricingContext        │
                                  │   broker-side pricing (~30 mthd) │
                                  │   fn prover() -> &ProverObj      │  ◄── single
                                  │   fn supported_selectors()       │      prover,
                                  │   fn preflight_cache()           │      single
                                  │   ... (gas, oracle, balances ...)│      backend,
                                  └─┬────────────────────────────────┘      bypasses
                                    │                                       registry
                                    │ ◄─ impl: broker's OrderPricer         ⚠
                                    │     (the SERVICE that drives          (deliber-
                                    │      `price_order`)                    ate non-
                                    │                                        goal per
                                    │                                        new doc
                                    │                                        comment)
                                    ▼
                              shared with no
                              other trait
                              ↓
                         (siblings,
                          do not share types)
                              ↑
   ┌──────────────────────────┴─────────────────────────────────────────┐
   │ PR #1988                                                            │
   │                                                                     │
   │   trait RequestPricer                                               │
   │     async fn price(request) -> Result<Accept{work_units}|Skip>      │
   │                                                                     │
   │     ◄─ impl RiscZeroRequestPricer<P>                                │
   │           holds: prover (a LocalExecutor wrapped as ProverObj ⚠),   │
   │                  provider, market_addr, chain_id,                   │
   │                  price_provider, price_oracle                       │
   │                                                                     │
   │   trait BackendRequestPricerFactory<P>                              │
   │     type Pricer: RequestPricer                                      │
   │     fn make_pricer(ctx: RequestPricingContext<P>) -> Pricer         │
   │                                                                     │
   │     ◄─ impl RiscZeroRequestPricingBackend                           │
   │                                                                     │
   │   struct RequestPricingContext<P>  (6 flat fields, #[non_exhaustive])│
   │                                                                     │
   │   Client::with_request_pricer / with_request_pricing_backend        │
   │     installs ──► StandardRequestBuilder.pricer                      │
   │                                                                     │
   │     fallback when None: INLINE R0 path (duplicates the R0 pricer)  ⚠│
   └─────────────────────────────────────────────────────────────────────┘
```

### What's wrong with the current shape

- **`ComputeClaimDigest` lives in four places**: registry-routed (inside the R0
  backend), Submitter-injected, Predicate::eval inline, BackendProvider internal.
- **`Prover` is one trait conflating preflight + full proving.** LocalExecutor
  "implements" it but panics on most methods.
- **`OrderPricingContext` (broker) and `RequestPricer` (SDK) are siblings**
  answering "is this order priced OK?" with no shared type.
- **`BackendRegistry` exists but isn't load-bearing for several consumers**:
  - SDK pricing has its own factory mechanism.
  - Submitter passes claim-digest standalone.
  - Broker `OrderPricer` holds a scalar `prover`, not the registry.
  - `Predicate::eval` doesn't know the registry exists.
- **R0 types still surface in "generic" traits**: `Prover::get_receipt`,
  `Predicate::eval`, `Prover` default-impl bodies.
- **Primitive-type signatures.** `&str`, `Vec<u8>`, `String` flow through
  trait boundaries — easy to confuse "image id" with "input id" with
  "proof id". The type system can't help.
- **`find(sel) -> Option<&BackendEntry>` forces every call site to write
  `.ok_or_else(...)` with its own ad-hoc error message** — six near-identical
  copies across the broker. Missing-backend isn't actually a "runtime miss" —
  it's a configuration / wiring bug — but the type pretends it's recoverable.

---

## 2. Proposed shape

```
═══════════════════════════════════════════════════════════════════════════════
                  boundless-market (core, R0-free)
═══════════════════════════════════════════════════════════════════════════════

──────────────────────────── strong types at the seam ─────────────────────────

  pub struct ProgramId(FixedBytes<32>);          // already exists
  pub struct PublicOutput(Bytes);                // already exists; "journal"
                                                  // for R0
  pub struct ImageBytes(Vec<u8>);                // raw ELF / program binary
  pub struct InputBytes(Vec<u8>);                // raw input blob
  pub struct InputId(String);                    // prover-assigned input handle
  pub struct ProofId(String);                    // prover-assigned proof handle
  pub struct CompressedProofId(String);
  pub struct PreflightSessionId(String);
  pub struct CompressedReceipt(Vec<u8>);         // opaque, backend-defined
  pub struct Seal(Vec<u8>);                      // on-chain bytes
  pub struct WorkUnits(u64);                     // backend-defined cost metric

  All carry: derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)
  All carry: From<inner> + AsRef<inner-borrow> for ergonomic conversion at
             boundaries. No more passing raw `String` and hoping it's the
             right kind of String.

  Why: a function signature that reads
       fn compress(&self, &ProofId) -> Result<CompressedProofId, …>
       cannot be misused. A signature that reads
       fn compress(&self, &str)     -> Result<String, …>
       absolutely will be.

────────────────────────── traits, layered ────────────────────────────────────

  ┌──────────────────────────────────────────────────────────────────────┐
  │ trait ComputeClaimDigest                                              │
  │   (unchanged — small, generic, owned by the formula)                  │
  └──────────────────────────────────────────────────────────────────────┘

  ┌──────────────────────────────────────────────────────────────────────┐
  │ trait PreflightProver        (NEW — extracted from `Prover`)         │
  │   async has_image(&ProgramId)                                         │
  │       -> Result<bool, ProverError>;                                   │
  │   async upload_image(&ProgramId, ImageBytes)                          │
  │       -> Result<(), ProverError>;                                     │
  │   async upload_input(InputBytes)                                      │
  │       -> Result<InputId, ProverError>;                                │
  │   async preflight(&ProgramId, &InputId, …)                            │
  │       -> Result<PreflightResult, ProverError>;                        │
  │   async get_preflight_journal(&PreflightSessionId)                    │
  │       -> Result<Option<PublicOutput>, ProverError>;                   │
  │                                                                       │
  │   // identity conversions — no free hex/format helpers leaking R0:    │
  │   fn parse_program_id(&self, s: &str)                                 │
  │       -> Result<ProgramId, ProverError>;                              │
  │   fn format_program_id(&self, id: &ProgramId) -> String;              │
  │   fn decode_input_bytes(&self, raw: &InputBytes)                      │
  │       -> Result<InputBytes, ProverError>;                             │
  └──────────────────────────────────────────────────────────────────────┘
            ▲                  ▲
            │                  │ super-trait
            │                  │
  ┌─────────┴─────────────┐    │
  │ impl LocalExecutor    │    │   (LocalExecutor only implements preflight,
  │   (preflight-only,    │    │    so the type system enforces what callers
  │    no panics, types   │    │    can ask of it)
  │    say so)            │    │
  └───────────────────────┘    │
                               │
  ┌────────────────────────────┴──────────────────────────────────────────┐
  │ trait Prover : PreflightProver                                        │
  │   async prove_stark(&ProgramId, &InputId, Vec<ProofId>)               │
  │       -> Result<ProofId, ProverError>;                                │
  │   async wait_for_stark(&ProofId)                                      │
  │       -> Result<ProofResult, ProverError>;                            │
  │   async cancel_stark(&ProofId)                                        │
  │       -> Result<(), ProverError>;                                     │
  │   async compress(&ProofId)                                            │
  │       -> Result<CompressedProofId, ProverError>;                      │
  │   async compress_blake3_groth16(&ProofId)                             │
  │       -> Result<CompressedProofId, ProverError>;                      │
  │   async get_compressed_receipt(&CompressedProofId)                    │
  │       -> Result<Option<CompressedReceipt>, ProverError>;              │
  │   async get_journal(&ProofId)                                         │
  │       -> Result<Option<PublicOutput>, ProverError>;                   │
  │     (no R0 types in any signature)                                    │
  └─┬─────────────────────────────────────────────────────────────────────┘
    │
    │ ◄─ impl Bonsai / DefaultProver  (in broker)


  ┌──────────────────────────────────────────────────────────────────────┐
  │ trait Backend : Send + Sync                                          │
  │   /// One value, per-zkVM. Carries everything a backend owns.        │
  │                                                                       │
  │   fn selectors(&self) -> &[VerifierSelector];                         │
  │   fn preflight_prover(&self) -> &dyn PreflightProver;                 │
  │   fn prover(&self)           -> &dyn Prover;                          │
  │   fn claim_digest(&self, sel) -> Result<                              │
  │       &dyn ComputeClaimDigest, UnknownSelectorError>;                 │   ◄── trait
  │                                                                       │       object,
  │   async fn compress(&self, &ProgramId, &ProofId, VerifierSelector)    │       not bytes
  │       -> Result<CompressedProofId, BackendError>;                     │
  │   async fn encode_seal(&self, &CompressedProofId, VerifierSelector)   │
  │       -> Result<Seal, BackendError>;                                  │
  │                                                                       │
  │   /// Pricing: backend owns its cost model.                          │
  │   async fn price_request(&self, &ProofRequest)                        │
  │       -> Result<RequestPricingOutcome, BackendError>;                 │
  │   fn pricing_context(&self) -> &PricingContext;                       │
  └─┬────────────────────────────────────────────────────────────────────┘
    │
    │ ◄─ impl RiscZeroBackend (in r0-backend)
    │       │
    │       ├ holds: preflight_prover, prover, snark_prover, dev_mode,
    │       │        provider, market_addr, chain_id, price_providers
    │       └ exposes the rest as methods over its state


  ┌──────────────────────────────────────────────────────────────────────┐
  │ struct BackendRegistry                                               │
  │   HashMap<VerifierSelector, Arc<dyn Backend>>                        │
  │                                                                       │
  │   pub fn register(&mut self, b: Arc<dyn Backend>)                    │
  │       -> Result<(), RegistryError>           // reads b.selectors()  │
  │                                                                       │
  │   pub fn find(&self, sel) -> Result<                                 │
  │       &Arc<dyn Backend>, UnknownSelectorError>                       │
  │                                                                       │
  │   pub fn known_selectors(&self)                                      │
  │       -> impl Iterator<Item = VerifierSelector> + '_                 │
  │                                                                       │
  │ (~ 30 LOC; ProverRegistry, BackendEntry.name, BackendEntry.selectors │
  │  Vec, len/is_empty/backends, prover_view, supported_selectors all    │
  │  deleted)                                                            │
  └──┬────────────────────────────────────────────────────────────────────┘
     │
     │ THE ONE entry point. Every backend lookup goes through here.
     │
     ├────► broker ProvingService  ── registry.find(sel)?.prover()
     ├────► broker Aggregator      ── registry.find(sel)?.prover()
     ├────► broker Submitter       ── registry.find(sel)?.{compress, encode_seal,
     │                                                       claim_digest}
     │                                  registry.find(assessor_sel)?.claim_digest()
     │                                  (NO standalone Arc<dyn CCD> ctor arg)
     ├────► broker OrderPricer     ── registry.find(sel)?.{preflight_prover,
     │                                                      pricing_context}
     │                                  (replaces OrderPricingContext::prover())
     ├────► SDK Client             ── registry.find(sel)?.price_request(...)
     │                                  (replaces RequestPricer factory)
     └────► Predicate::eval_with   ── registry.find(sel)?.claim_digest()
                                       (replaces inline R0 ComputeClaimDigest;
                                        Predicate::eval is deleted)


──────────────────── unknown-selector handling ────────────────────────────────

  pub struct UnknownSelectorError(pub VerifierSelector);

  Note: `find()` returns `Result`, not `Option`. A missing backend isn't a
  recoverable miss — it's a wiring / init problem. The error type makes
  that intent explicit at every call site and standardises the message
  ("no backend registered for selector 0xabcdef01") instead of six ad-hoc
  copies across the broker.

  Because the registry is selector-keyed and selectors are 4 bytes, the
  set of legitimate selectors at runtime is small (≤ 5 today) and known
  at startup. We can — and should — filter incoming requests against
  `registry.known_selectors()` at the market-monitor / order-evaluator
  boundary:

      market_monitor: drop request with unknown selector → SKIP_UNSUPPORTED_SELECTOR
                      (existing pricing skip code, just moved earlier)
                      ▼
      order_evaluator / pricing / proving / aggregator / submitter:
          registry.find(sel) ALWAYS succeeds — the request was filtered
          earlier. Hot-path call sites can `.expect(...)` with confidence
          and the trace points to the upstream filter as the invariant
          owner.

  In short: registry.find() returns Result for type honesty; the request
  pipeline is structured so the Result is always Ok by the time it
  reaches per-order code.

═══════════════════════════════════════════════════════════════════════════════
                       What this unification gives you
═══════════════════════════════════════════════════════════════════════════════

  • ONE trait per backend, ONE Arc per backend in memory.
  • ONE place where "which backend for selector X" gets answered.
  • Strong newtypes mean signatures self-document and the compiler catches
    accidental string/byte confusion. From<inner> keeps construction cheap.
  • ComputeClaimDigest exposed as a returned trait object → Submitter,
    Predicate::eval_with, and BackendProvider all consume the same impl,
    via the same lookup.
  • LocalExecutor can no longer pretend to implement proving — the type
    system says it implements PreflightProver only.
  • Prover trait has no R0 types in any signature.
  • SDK pricing and broker pricing share the registry; they may still
    have different cost models, but both ask the same backend.
  • Adding a new zkVM = implement ONE trait, register ONE Arc, done.
  • Unknown selectors fail fast at the front of the pipeline; downstream
    code treats registry.find() as a static invariant.
```

---

## 3. What happens to `OrderPricingContext` and `RequestPricingContext`

Both go away — for different reasons.

### `OrderPricingContext` (the ~30-method broker trait) is a god trait

It conflates two structurally unrelated concerns. The methods split cleanly:

| Methods | What they really are |
|---|---|
| `prover()`, `supported_selectors()`, `preflight_cache()`, `preflight_execute()`, `preflight_journal()`, `calculate_exec_limits()` | **Backend capability** — how does this zkVM count work? What's its preflight cache? |
| `market_config()`, `current_gas_price()`, `estimate_erc1271_gas()`, `check_available_balances()`, `check_requestor_allowed()`, `denied_requestor_addresses()`, `convert_to_eth()`, `convert_to_zkc()`, `is_priority_requestor()`, `downloader()` | **Broker policy / deployment env** — what does *this broker* allow, what does its wallet hold, what's the gas market right now? |

They got jammed into one trait because `price_order()` (the actual decision
function) needed both. But they have nothing structurally in common, and
the "trait" never had more than one implementor.

### `RequestPricingContext<P>` (the PR #1988 struct) exists only because the factory has no state

Once the factory pattern goes away (backend is a value, constructed once at
startup), there's nothing to pass to a context struct.

### What replaces them

**Backend-capability half** moves into `Backend`:

```rust
trait Backend {
    // ...the rest of the trait from §2...

    /// What this backend thinks this request costs, ignoring local policy.
    async fn price_request(&self, req: &ProofRequest)
        -> Result<RequestPricingOutcome, BackendError>;
}

pub enum RequestPricingOutcome {
    Priced(RequestPricing),
    Reject { reason: SkipReason },  // unsupported, malformed, blake3 size, ...
}

pub struct RequestPricing {
    /// Native units, in this backend's own metric (R0 cycles, OpenVM instructions, ...).
    /// For logging / telemetry / per-backend safety caps only. The broker
    /// does NOT compare this across backends.
    pub work_units: WorkUnits,

    /// Estimated wall-clock proving time on THIS backend's deployment.
    /// The broker sums these across in-flight orders for capacity math
    /// and compares to the request's deadline.
    pub estimated_duration: Duration,

    /// Estimated proving cost in ETH. The backend converts its native
    /// metric via its own cost model (cycles × $/cycle, plus oracle if
    /// needed). The broker compares `estimated_cost_eth + on_chain_gas`
    /// to `offer.price` — no cycle math.
    pub estimated_cost_eth: U256,

    /// Journal / output size. Affects on-chain gas; not backend-specific.
    pub output_size: u64,
}
```

That's the **shared** entry point. SDK request builder calls it. Broker
`OrderPricer` calls it. Same method, same answer, same backend.

### Work units don't compare across backends — normalize at the seam

`WorkUnits(u64)` alone is wrong as a decision variable: an R0 cycle and an
OpenVM instructions are different physical things. You can't sum them ("how much
capacity am I using?"), compare them to one threshold (`max_mcycle_limit`
means something different to each backend), or multiply them by one rate.
The moment a second zkVM joins, native cycles stop being a decision unit.

Three possible answers:

| | Approach                                                                                                                                                                                                                      | Pros | Cons |
|---|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---|---|
| (a) | **Per-backend broker config.** `[backend.r0]` and `[backend.openVM]` sections each with their own cycle limits, prices, peak-khz. Broker picks the right section per selector.                                                | Simple. Native units stay native. | Capacity math across backends still impossible; config sprawl. |
| (b) | **Normalize at the seam.** Backend translates its native cycles into the two units the broker actually cares about: wall-clock **duration** and **ETH cost**. Cycle-specific knobs move inside the backend's own constructor. | Broker policy is unit-universal; cross-backend capacity works; broker config stops mentioning cycles. | Backend must know its own deployment throughput (calibration knob). |
| (c) | **Push all policy into the backend.** Backend gets the broker's allowlist / capacity / margin config injected at startup and decides itself. Broker just asks "should I take this?"                                           | Broker is a thin orchestrator. | Per-backend policy duplication; broker can't make cross-backend decisions. |

**The proposed shape uses (b).** The broker fundamentally cares about two
questions: *does the proof finish in time* and *is the offer profitable*.
Both reduce to **duration** and **ETH** — universal across backends. Cycles
are a counter neither the broker nor the user cares about.

The cycle-specific knobs (`max_mcycle_limit`, `peak_prove_khz`,
`min_mcycle_price`) move into the **backend's** constructor as deployment
calibration. The R0 backend reads `[backend.r0]` at startup, the OpenVM
backend reads `[backend.openVM]`, each enforces its own cycle caps internally
and reports back as `Reject { reason: "exceeds R0 cycle cap" }` if
breached. The broker config stops mentioning cycles entirely.

**Broker-policy half** is struct state on the broker's `OrderPricer` service —
not a trait at all:

```rust
pub struct OrderPricer {
    backends: BackendRegistry,
    config: ConfigLock,            // max_mcycle_limit, peak_prove_khz, ...
    price_oracle: Arc<PriceOracleManager>,
    allow_requestors: AllowRequestors,
    erc1271_gas_cache: Erc1271GasCache,
    // ...
}

impl OrderPricer {
    pub async fn price_order(&self, req: &OrderRequest) -> OrderPricingOutcome {
        let backend = self.backends.find(req.request.requirements.selector)?;

        // 1. Backend says what the work costs (duration + ETH, both universal).
        let priced = match backend.price_request(&req.request).await? {
            RequestPricingOutcome::Reject { reason } => return Skip(reason),
            RequestPricingOutcome::Priced(p) => p,
        };

        // 2. Local policy decides whether we take it — all unit-universal.

        // requestor allowlist / denylist (selector-independent)
        if !self.allow_requestors.check(req)? { return Skip("denied requestor"); }

        // capacity: sum of durations, no cycles
        if self.in_flight_duration() + priced.estimated_duration > self.config.max_in_flight {
            return Skip("broker at capacity");
        }

        // deadline: time vs time
        if priced.estimated_duration > req.deadline_remaining() {
            return Skip("cannot finish in time");
        }

        // profitability: ETH vs ETH
        let gas_cost = self.estimate_gas_cost(req, priced.output_size).await?;
        if priced.estimated_cost_eth + gas_cost > req.offer.price {
            return Skip("unprofitable");
        }

        self.decide_lock_timing(req, priced)
    }
}
```

`apply_local_policy` becomes much smaller than `OrderPricingContext`
suggested: requestor lists, duration capacity, deadline, profitability,
gas — all unit-universal. The cycle gymnastics that today live in
`calculate_exec_limits` (ETH-based limit vs collateral-based limit,
`ProveLimitReason::EthPricing` / `CollateralPricing` / `ConfigCap` /
`DeadlineCap`) all move **inside the backend**, where the cycle unit is
meaningful. The 30-method god trait dissolves further than §3 first
implied.

**SDK side**, with all the indirection gone:

```rust
impl Client {
    pub fn with_backends(self, registry: BackendRegistry) -> Self { … }
}

// inside StandardRequestBuilder.run_pricing_check:
let backend = self.backends.find(req.requirements.selector)?;
match backend.price_request(req).await? {
    RequestPricingOutcome::Priced { work_units, .. } => tracing::debug!(…),
    RequestPricingOutcome::Reject { reason }        => tracing::warn!(…),
}
```

No `RequestPricer` trait, no `BackendRequestPricerFactory`, no
`RequestPricingContext`. The backend is already constructed with whatever
environment it needs; the client hands the registry to the request builder.

### Why this split is honest

What was wrong with `OrderPricingContext` wasn't that it had 30 methods — it
was that two unrelated concerns (backend-cost-model and broker-config) were
forced through one trait so one method (`price_order`) could call both. Once
`price_request` lives on `Backend` and `apply_local_policy` lives on the
broker struct, the god trait dissolves naturally. Neither half is a trait
anymore.

What was wrong with `RequestPricingContext` wasn't the struct — it was the
factory pattern that required it. With backends-as-values, you construct the
backend once at startup with whatever it needs, and from then on it's just
`registry.find(sel)?.price_request(req)`.

### Subtlety: lock-timing decisions stay broker-side

The broker's `price_order` today returns three outcomes: `Lock`,
`ProveAfterLockExpire`, `Skip`. The first two encode timing decisions
(when to lock, when to commit) that depend on the broker's risk appetite.
Those belong in `apply_local_policy`, not in `Backend::price_request`.

The backend's answer is: *"this request takes N seconds and M ETH on my
deployment."* The broker decides whether to act on it and when.

### Subtlety: backend throughput is deployment calibration

The "backend translates cycles → duration" promise only holds if the
backend knows its own throughput. That's a deployment property, not a
code property — a prover with new GPUs proves faster than one running
the old build. So `[backend.r0] cycles_per_sec = 5_000_000` becomes a
config knob the operator sets once per deployment, and the R0 backend
constructor reads it at startup. Same way `peak_prove_khz` works today,
except now it's *the backend's* knob (in `[backend.r0]`), not the
broker's global setting.

If a prover swaps to faster hardware, they update one number. The broker
doesn't have to know.

capossele:
Agree with the new design direction, closing for now to reiterate on it

PR #2000, opened 2026-05-19
BM-2988: enable gcs feature by default

capossele:
Adds the `gcs` feature to the broker and indexer's `boundless-market` dependency so deployed binaries can fetch `gs://` image and input URLs out of the box. Until now, both crates were built with only the `s3` default feature, which caused every `gs://` URL to fail at the URL-scheme dispatcher before any auth was attempted.


crates/broker/Cargo.toml
```diff
@@ -41,7 +41,7 @@ bincode = { workspace = true }
 blake3_groth16 = { workspace = true }
 bonsai-sdk = { workspace = true }
 boundless-assessor = { workspace = true }
-boundless-market = { workspace = true, features = ["prover_utils"] }
+boundless-market = { workspace = true, features = ["prover_utils", "gcs"] }
```
willpote: I was thinking we'd model at broker level:

```toml
[features]
default = ["gcs"]
gcs = ["boundless-market/gcs"]

[dependencies]
boundless-market = { workspace = true, features = ["prover_utils"] }
```

so that you can still build broker without GCS if you like via `cargo build -p broker --no-default-features`

PR #2001, opened 2026-05-19
BM-2989: Raise bento-down alarm to 2h to match broker graceful shutdown

willpote:
Bumps the SEV2 `bento-down` alarm from 1h to 2h (8 × 15-min datapoints) so a healthy graceful shutdown — where the broker waits up to `SHUTDOWN_GRACE_PERIOD_SECS` (2h) for in-flight committed orders to drain — does not page. 

Changes
* Bento-down alarm threshold raised from 4×15min (1h) to 8×15min (2h) for prod and staging; 3h escalation alarm kept as is.
* Adds a note to the ops-logs-query skill so future on-call diagnoses look for the `"in-progress orders to complete"` log before treating `bento_active=0` + `channel closed` errors as an outage


bobbobbio (approved):
only thought is how we would make sure to update this place if we were to change `SHUTDOWN_GRACE_PERIOD_SECS` in the future

zeroecco (approved):

capossele (approved):

willpote:
> only thought is how we would make sure to update this place if we were to change `SHUTDOWN_GRACE_PERIOD_SECS` in the future

Thats true, there isn't really an obvious place to put it that wouldn't create a strange dependency. And we have never updated this value thus far, so I'm inclined to leave it for now. Added a comment to the constant though

PR #2002, opened 2026-05-19
BM-2990: Retry eth_getTransactionByHash and expose underlying error detail

willpote:
Add a retry to getTransactionByHash to help with transient issues we are seeing.
Also changing the errors so it logs the actual underlying error we are getting from the RPC provider. Then we can send that to Taiko.

Changes
* Wrap each `get_transaction_by_hash` call in a retry (2 retries, 1s apart) that treats a missing tx or `block_number = null` as retryable
* Add the tx hash to the defensive `block_number` context message so the error identifies the offending transaction
* Switch `ServiceError` Display to use `{0:#}` for the anyhow variant (walks the source chain) and `{0:?}` for non-anyhow variants (`RpcError`, `MarketError`, `EventQueryError`, `DbError`) so their inner detail surfaces
* Change the top-level `bail!` in market-indexer / backfill to `{err:#}` so the chain reaches the FATAL line
* Note `forge build` as a prerequisite in the pr skill, so the cargo check step doesn't spurious-fail on Solidity-bound crates

### DIFF
--- a/crates/indexer/src/market/service/chain_data.rs
+++ b/crates/indexer/src/market/service/chain_data.rs
@@ -20,7 +20,7 @@ use super::{
 use crate::db::market::{IndexerDb, TxMetadata};
 use crate::market::ServiceError;
 use alloy::eips::{BlockId, BlockNumberOrTag};
-use alloy::network::{AnyNetwork, Ethereum};
+use alloy::network::{AnyNetwork, Ethereum, Network};
 use alloy::primitives::{Address, B256};
 use alloy::providers::Provider;
 use alloy::rpc::types::{Filter, Log};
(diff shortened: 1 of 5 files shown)


crates/indexer/src/market/service/chain_data.rs
```diff
@@ -414,7 +416,26 @@ where
...
+                                    return Err(anyhow!(
+                                        "transaction {} returned without block_number",
+                                        hex::encode(tx_hash)
+                                    ));
+                                }
+                                Ok(tx)
+                            },
+                            "get_transaction_by_hash",
+                        )
+                        .await
```
bobbobbio: I suggest making a function for this section, easier to read with less indentation

capossele (approved):

crates/indexer/src/market/service/chain_data.rs
```diff
@@ -504,7 +515,10 @@ where
 
         // Step 4: Build final map from tx_hash to TxMetadata and update service map
         for (tx_hash, tx) in tx_map {
-            let bn = tx.block_number.context("block number not found")?;
+            let bn = tx.block_number.context(anyhow!(
+                "block_number missing on transaction {} after eth_getTransactionByHash",
+                hex::encode(tx_hash)
```
bobbobbio: `with_context` is preferred to avoid running formatting code in success path

bobbobbio (approved):

bobbobbio:
FWIW I think this isn't how `thiserror` intended for chained errors to function, I don't know enough about it though

PR #2003, opened 2026-05-19
INF-51: Zeroecco/inf 51

zeroecco:
support for batch ops of kailua


infra/kailua-batch/Pulumi.prod.yaml
```diff
@@ -0,0 +1,46 @@
+secretsprovider: awskms:///arn:aws:kms:us-west-2:968153779208:alias/pulumi-secrets-key
```
willpote, line 1: Any reason these stacks aren't market specific like all our other stacks?

As is, if we ever wanted to send Kailua requests to a market on another chain (e.g. on Taiko) we'd have to introduce some inconsistencies in naming

infra/kailua-batch/index.ts
```diff
@@ -0,0 +1,549 @@
...
+    // Override entirely with kailua-batch:KAILUA_COMMAND in stack config if needed.
+    const defaultProveScript = [
+        ': "${S3_ACCESS_KEY:?S3_ACCESS_KEY is required for R2}"',
+        ': "${S3_SECRET_KEY:?S3_SECRET_KEY is required for R2}"',
+        'export AWS_ACCESS_KEY_ID="${S3_ACCESS_KEY}"',
+        'export AWS_SECRET_ACCESS_KEY="${S3_SECRET_KEY}"',
+        'FINALIZED=$(cast bn finalized -r "$L2_RPC")',
+        'START_BLOCK=$((FINALIZED - START_BLOCK_OFFSET))',
+        'echo "KAILUA_RUNNING finalized=${FINALIZED} start_block=${START_BLOCK} offset=${START_BLOCK_OFFSET} block_count=${BLOCK_COUNT}"',
+        `cd ${kailuaWorkdir}`,
```
willpote, line 213: It seems that we are just proving the last 40ish blocks from the finalized tip at the time of schedule (every 20 mins). 

We aren't making an attempt to prove the full chain + most blocks just get skipped (we prove about 40/600 over each 20 mins).

While this works for just sending work to the market, is it useable / useful in its current state to just prove 40ish blocks from tip? I was expecting us to take an approach similar to Signal where we track which blocks we haved proved up to and prove the full chain

infra/kailua-batch/index.ts
```diff
@@ -0,0 +1,549 @@
...
+                minute,
+                taskIndex,
+                startBlockOffset: minute * tasksPerMinute + taskIndex,
+                blockCount: String(taskIndex + 1),
+                cronExpression: `cron(${minute}/${windowMinutes} * * * ? *)`,
+            });
+        }
+    }
+    return specs;
+}
```
willpote: We want to scale up soon right and eventually completely replace the OGs, it seems this will involve creating hundreds, maybe thousands, of EventBridge schedules? 

It feels a bit odd, why not just have the lambda kick off multiple tasks? I.e. trigger it every 15 mins, and have it run for 15 mins looping and kicking off tasks.

infra/kailua-batch/index.ts
zeroecco: we actually have a neat overlapping system where we prove each individual block and blocks of 2 staggering ensuring coverage. It was instructed to do this initially and potentially prove every block. but that is on hold I am told.

infra/kailua-batch/Pulumi.prod.yaml
zeroecco: I was told this was the only planned deployment. I am happy to tear it down and rebuild if that is the preference. LMK and I will do it

infra/kailua-batch/index.ts
zeroecco: that function just generates the invokes inside of one event bridge. for this configuration there are two bridges 0 and 1 to create the requested volume in the ticket. so event bridges should be `n*2` the chains supported.

individually launching tasks means there is no H.O.L. blocking if something goes wrong in one of the workers. pricing should be roughly the same since these tasks pay on a per second basis.

infra/kailua-batch/index.ts
willpote: Yes but since we don't track any state on our side, I don't really see how we can ensure we prove every block

e.g. 1/ If a task fails / event doesn't trigger, we have no tracking of what did/didn't get proved
2/ we rely on the RPC to tell us finalized block and work from that. If the RPC falls behind, or goes down, we can skip blocks and never know

I think this is fine for now just to get traffic flowing, just noting for future when we try to do all blocks

willpote (approved):
If we ever want to prove every block reliably, I think we'll need to move to more of a stateful-cursor + queue design, as any issue with Scheduled EventBridge triggers (e.g. missing trigger or triggered early/late) or RPC issues (e.g. down or lagging) will cause us to miss a minute in our window.

But since we are far from that at the moment, I think this is good

PR #2004, opened 2026-05-19
BM-3002: Broker Backend Service

capossele:
### Summary

  Extracts all RISC Zero–specific proving, aggregation, and submission logic out of the
  `broker` and behind a backend-neutral seam, so a second proving backend can be plugged
  in without touching the broker's order lifecycle.

  - **New crate `boundless-backend`** — the neutral `Backend` / `BatchProcessor` traits, a
    `BackendRouter` that dispatches by 4-byte verifier selector, and opaque durable-state
    blobs the broker persists but never interprets.
  - **New crate `risc0-backend`** — the RISC Zero implementation of those traits (set-builder,
    assessor, groth16 / blake3-groth16 wraps, and the `provers/` moved from the broker).
  - **Broker is now backend-agnostic** — it owns the order state machine, SQLite, and
    on-chain I/O, and calls the backend through the router.

  > [!WARNING]
  > This PR **requires a fresh broker DB** (no migration; existing DBs fail closed). Wipe
  > broker state on deploy — see Breaking changes.

  ### Architecture

  ```text
    +--------------------------------------------------------------+
    | broker                                                        |
    |   order lifecycle | SQLite state | on-chain submission        |
    |                                                               |
    |  order_pricer   order_processor    batcher        submitter   |
    +------|---------------|-----------------|--------------|--------+
           | evaluate      | process         | update/      | build_fulfillments
           | _request      | _order          | close_batch  | + verifier_update
           v               v                 v              v
    +--------------------------------------------------------------+
    | boundless-backend   (backend-neutral)                        |
    |   trait Backend  +  trait BatchProcessor                     |
    |   BackendRouter:   verifier selector (4 bytes) -> backend    |
    |   opaque state:    BackendOrderState / BackendBatchState     |
    +-----------------------------|--------------------------------+
                                  | dyn dispatch
                                  v
    +--------------------------------------------------------------+
    | risc0-backend   (RISC Zero implementation)                   |
    |   Risc0Backend : Backend                                     |
    |   set-builder . assessor . groth16 / blake3-groth16 wraps    |
    |   provers/  (Bonsai, Bento, Default)                         |
    +--------------------------------------------------------------+
  ```

  ### What changed

  - **Added** `crates/boundless-backend` (traits, `BackendRouter`, neutral types, `futures_retry`).
  - **Added** `crates/risc0-backend` (`Risc0Backend`, batch processor, `provers/`).
  - **Renamed** broker modules: `aggregator` -> `batcher`, `proving` -> `order_processor`.
  - **Moved** `broker/src/provers/` -> `risc0-backend/src/provers/`.
  - **SDK (`boundless-market`)** — request evaluation extracted into
    `prover_utils/request_evaluator.rs` with neutral `EvaluationRequest` / `EvaluationLimits` /
    `RequestEvaluation` / `EvaluationMetrics` types; preflight cache keyed by `max_cycles`.
  - Per-order / per-batch backend state is an opaque JSON blob authored by the backend and
    versioned (rejects newer-than-known schema, fails closed).

  ### Behavior changes vs `main`

  - **Capacity-slot release** — an order fulfilled externally *before* batch processing now
    emits `CommitmentComplete{ProvingFailed}`, freeing the proving slot (`main` leaked it; this
    matches `main`'s existing *expired* branch).
  - **`updated_at` timestamp fix** — corrects a `'$.update_at'` JSON-key typo in three DB
    update paths that silently never refreshed `updated_at` on `main`.
  - **Idempotent submit** — already-`Done`/`Failed` orders are skipped on submit retry, and an
    all-skipped batch is marked submitted.

  ### Breaking changes

  - **Requires a fresh broker DB.** `OrderStatus` / `BatchStatus` variants were renamed (their
    names are persisted in the JSON state blob) and new `backend_id` / `backend_state` fields
    were added. There is no migration; an existing DB with in-flight orders fails closed (clean
    deserialize error, not silent corruption). **Operators must wipe broker state on deploy.**

  ### Internal API changes (no requestor-SDK impact)

  - The `prover_utils`-gated evaluation API was reshaped for the backend-neutral seam:
    `OrderPricingContext` preflight hooks folded into the new evaluator, and `PreflightCacheValue`
    replaced by the neutral `RequestEvaluation`. These items are feature-gated and broker-internal
    — the only in-tree consumer is the broker, updated in this PR. The default / requestor SDK
    surface (`Client`, `ProofRequest`, storage) is unchanged, and the examples build unmodified.

  ### Out of scope (follow-ups)

  - `Prover` trait still returns `risc0_zkvm::Receipt` — migration is its own PR.
  - `Order.total_cycles` / `Order.journal_bytes` -> `EvaluationMetrics`.
  - Split `OrderPricingContext` -> `BrokerPricingPolicy` + `RequestEvaluator`.


crates/boundless-backend/src/types.rs
```diff
@@ -0,0 +1,413 @@
...
+
+impl TryFrom<&str> for BackendId {
+    type Error = anyhow::Error;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        Self::new(value)
+    }
+}
+
+macro_rules! string_id {
```
jonastheis, line 34: why can't we use this for `BackendId`?

crates/boundless-backend/src/types.rs
```diff
@@ -0,0 +1,413 @@
...
+    pub order_id: String,
+    pub error: anyhow::Error,
+}
+
+pub enum OrderFulfillmentResult {
+    Fulfilled(OrderFulfillmentArtifact),
+    Failed(OrderFulfillmentFailure),
+}
+
+pub struct SubmissionPlan {
```
jonastheis, line 271: Can you explain how this translates to a `JointVerifier` or the `OnchainAssessor`?

crates/broker/src/backend/types.rs
```diff
@@ -0,0 +1,413 @@
...
+    async fn process_order(&self, cmd: ProcessOrder) -> Result<OrderProcessProgress>;
+
+    async fn cancel_order(&self, order: &Order) -> Result<()>;
+
+    async fn estimate_batch_size(&self, cmd: BatchSizeEstimateRequest)
+        -> Result<BatchSizeEstimate>;
+
+    async fn update_batch(&self, cmd: UpdateBatch) -> Result<BatchUpdate>;
+
+    async fn close_batch(&self, cmd: CloseBatch) -> Result<BatchClose, BackendError>;
```
jonastheis: I think either `pub trait Backend: BatchProcessor + Send + Sync {` 

or better 

`fn batch_processor(&self) -> Result<BatchProcessorObj>;` 

as this keeps the `Backend` trait cleaner in case a backend doesn't support batching

crates/risc0-backend/src/lib.rs
jonastheis, line 1: Why is this in the general crate? shouldn't this have its own crate and be split in a few files maybe?

crates/broker/src/backend/risc0.rs
```diff
@@ -0,0 +1,1123 @@
...
+}
+
+#[async_trait]
+impl Backend for Risc0Backend {
+    fn id(&self) -> &BackendId {
+        &self.id
+    }
+
+    fn supported_selectors(&self) -> Vec<FixedBytes<4>> {
+        SupportedSelectors::default().selectors.keys().copied().collect()
```
jonastheis: I think we should name the selectors explicitly here to avoid any accidental issues

crates/boundless-market/src/prover_utils/request_evaluator.rs
```diff
@@ -0,0 +1,681 @@
...
+        limits: EvaluationLimits,
+    ) -> Result<RequestEvaluation, OrderPricingError>;
+}
+
+/// RISC0-backed request evaluation hooks.
+///
+/// Implementors provide the execution dependencies; the blanket [`RequestEvaluator`]
+/// implementation below handles upload, cache coalescing, bounded preflight, and
+/// journal retrieval exactly as the existing RISC0 path does.
+pub trait Risc0RequestEvaluatorContext {
```
jonastheis: Why is the risc0 type here?

crates/boundless-market/src/prover_utils/request_evaluator.rs
```diff
@@ -0,0 +1,681 @@
...
+    }
+}
+
+/// Broker-comparable work units derived from backend-native evaluation output.
+///
+/// Broker pricing and capacity policy consume this value. Backend evaluators are
+/// responsible for mapping their native work model into this normalized unit.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub struct NormalizedWork {
```
jonastheis, line 67: So currently this is risc0 cycles? and we revisit this later? then I think we should leave a TODO comment here.

crates/boundless-market/src/prover_utils/request_evaluator.rs
```diff
@@ -0,0 +1,681 @@
...
+    pub fn new(max_cycles: u64) -> Self {
+        Self { max_cycles }
+    }
+}
+
+/// Executes request preflight for pricing without making broker policy decisions.
+///
+/// This is the narrow backend-facing part of request evaluation. It returns execution facts
+/// and output bytes; profitability, collateral, deadlines, and capacity remain in
+/// [`super::OrderPricingContext`].
```
jonastheis: How exactly do these traits interact with each other?

crates/boundless-backend/src/types.rs
```diff
@@ -0,0 +1,413 @@
...
+
+    async fn update_batch(&self, cmd: UpdateBatch) -> Result<BatchUpdate>;
+
+    async fn close_batch(&self, cmd: CloseBatch) -> Result<BatchClose, BackendError>;
+}
+
+pub(crate) type BatchProcessorObj = Arc<dyn BatchProcessor>;
+
+#[async_trait]
+pub trait Backend: Send + Sync {
```
jonastheis, line 386: Why does the backend not expose `RequestEvaluator` or `OrderPricingContext`?

crates/broker/src/backend/router.rs
```diff
@@ -0,0 +1,568 @@
...
+
+use crate::Order;
+
+use super::types::{
+    Backend, BackendError, BackendId, BatchClose, BatchSizeEstimate, BatchSizeEstimateRequest,
+    BatchUpdate, CloseBatch, FulfillmentBatch, OrderProcessProgress, ProcessOrder, SubmissionPlan,
+    UpdateBatch,
+};
+
+type BackendObj = Arc<dyn Backend>;
```
jonastheis: should live in types?

crates/boundless-backend/src/router.rs
```diff
@@ -0,0 +1,568 @@
...
+    BatchUpdate, CloseBatch, FulfillmentBatch, OrderProcessProgress, ProcessOrder, SubmissionPlan,
+    UpdateBatch,
+};
+
+type BackendObj = Arc<dyn Backend>;
+
+#[derive(Clone, Default)]
+pub struct BackendRouter {
+    backends: HashMap<BackendId, BackendEntry>,
+    routes: HashMap<FixedBytes<4>, BackendId>,
```
jonastheis, line 38: Why not use the concrete type here?

crates/broker/src/backend/router.rs
```diff
@@ -0,0 +1,568 @@
...
+#[derive(Clone, Default)]
+pub struct BackendRouter {
+    backends: HashMap<BackendId, BackendEntry>,
+    routes: HashMap<FixedBytes<4>, BackendId>,
+}
+
+#[derive(Clone)]
+pub struct BackendEntry {
+    id: BackendId,
+    selectors: Vec<FixedBytes<4>>,
```
jonastheis: same here

crates/boundless-backend/src/router.rs
```diff
@@ -0,0 +1,568 @@
...
+        }
+
+        let mut entry_selectors = HashSet::new();
+        for selector in &entry.selectors {
+            if !entry_selectors.insert(*selector) {
+                anyhow::bail!(
+                    "selector {selector:?} is listed more than once for backend {}",
+                    entry.id()
+                );
+            }
```
jonastheis, line 63: we can just do 1 loop, check if exists first (below check) and then fail if with the error message (we can enhance it by checking if existing_backend == entry.id

crates/broker/src/backend/router.rs
```diff
@@ -0,0 +1,568 @@
...
+
+        for selector in &entry.selectors {
+            self.routes.insert(*selector, entry.id.clone());
+        }
+        self.backends.insert(entry.id.clone(), entry);
+        Ok(self)
+    }
+
+    fn backend_for_order(&self, order: &Order) -> Result<&BackendObj> {
+        let selector = order.request.requirements.selector;
```
jonastheis: This works for now. With #1982 we will basically need to reproduce the dispatching logic from the router contract to here. Default, classes and entries need to be registered and dispatched accordingly.

crates/broker/src/backend/router.rs
```diff
@@ -0,0 +1,568 @@
...
+            .routes
+            .get(&selector)
+            .with_context(|| format!("no backend registered for selector {selector:?}"))?;
+        self.backends
+            .get(backend_id)
+            .map(|entry| &entry.backend)
+            .with_context(|| format!("backend {backend_id} is not registered"))
+    }
+
+    fn backend_for_id(&self, backend_id: &BackendId) -> Result<&BackendObj> {
```
jonastheis: I'm not too sure about returning `&BackendObj` here. Feels like we should clone the `Arc` instead. Or maybe `&dyn Backend`

crates/broker/src/order_types.rs
```diff
@@ -171,6 +174,11 @@ pub struct Order {
     ///
     /// Populated after proof completion. if the proof is compressed
     pub(crate) compressed_proof_id: Option<String>,
+    /// Backend that processed this order.
```
jonastheis, line 168: Can't comment on the line as it's not changed but: what about `total_cycles` and `journal_bytes`?

crates/broker/src/broker.rs
```diff
@@ -294,14 +294,14 @@ impl Broker {
 
     fn build_provers(&self, config: &ConfigLock) -> Result<(ProverObj, ProverObj)> {
         let prover: ProverObj;
-        let aggregation_prover: ProverObj;
+        let batch_prover: ProverObj;
```
jonastheis: Why is the `Prover` trait from `crates/boundless-market/src/prover_utils/prover.rs` still used? It still contains risc0 specific types

crates/broker/src/broker.rs
```diff
@@ -822,53 +822,72 @@ impl Broker {
...
-            ));
-
-            runner.spawn_in_span(
-                proving_service,
-                service_runner::Criticality::Critical,
-                "proving service",
-                chain_span.clone(),
-            );
-
+            let risc0_backend_id = BackendId::new("risc0_v3")?;
```
jonastheis: I think the backend impl should name its own BackendId

crates/broker/src/broker.rs
```diff
@@ -822,53 +822,72 @@ impl Broker {
...
+                set_builder_img_id,
+                assessor_img_id,
+                deployment.boundless_market_address,
+                prover_addr,
+                chain_id,
+            ));
+            let risc0_backend = Arc::new(
+                Risc0Backend::new(
+                    risc0_backend_id.clone(),
+                    prover.clone(),
```
jonastheis: Shouldn't the backend own the prover and expose it only if necessary?

crates/broker/src/order_pricer/service.rs
jonastheis, line 1: I'm not sure I understand why `OrderPricer` still implements `OrderPricingContext` carries a `prover` and `supported_selectors`? Shouldn't this all go through the backend?

crates/broker/src/order_pricer/helpers.rs
```diff
@@ -459,12 +457,25 @@ where
             )))
         })
     }
+}
 
+impl<P> Risc0RequestEvaluatorContext for OrderPricer<P>
```
jonastheis: Here we have again some risc0 types?

jonastheis (commented):
The direction looks great. However I have a bunch of questions that I marked inline. 

High-level I think what is unclear to me are mostly the following:
- `Prover` trait and usage of it (it still contains risc0 types)
- how does `Backend` relate to `OrderPricingContext` and `RequestEvaluator`?
- why does `OrderPricer` not use `BackendRouter` but instead implements `OrderPricingContext`?
- risc0 types are not yet contained to a single crate


I haven't really looked at the submitter and batcher yet.

crates/broker/src/backend.rs
```diff
@@ -48,6 +50,53 @@ use crate::{
 };
 use anyhow::{Context, Result};
 
+/// Stable broker-side identity for a backend registration.
+///
+/// This is intentionally distinct from a verifier selector: many selectors may
+/// route to the same backend, and broker policy/batching is keyed by backend.
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct BackendId(String);
```
bobbobbio: strikes me a bit like a reimplementation of `non_empty_string::NonEmptyString`

wondering why its so important that string not be empty or purely whitespace actually

crates/broker/src/backend.rs
```diff
@@ -128,6 +128,47 @@ pub trait Backend: Send + Sync {
...
+    pub fn register_backend(
+        mut self,
+        selectors: impl IntoIterator<Item = FixedBytes<4>>,
+        backend: BackendObj,
+    ) -> Result<Self> {
+        for selector in selectors {
+            if self.routes.contains_key(&selector) {
+                anyhow::bail!("selector {selector:?} is already registered to a backend");
+            }
+            self.routes.insert(selector, backend.clone());
```
bobbobbio: This has maybe confusing behavior if you try to register a duplicate backend where it registers the ones before and not the ones after? I'm not sure what the caller wants.

I'm confused why we need these APIs to register a collection of them at all instead of one at a time.

If you leave it like this, you can use the `entry` API on `HashMap` to avoid a duplicate traversal of the `HashMap`

crates/broker/src/backend.rs
```diff
@@ -130,34 +130,67 @@ pub type BackendObj = Arc<dyn Backend>;
...
-    pub fn register_backend(
-        mut self,
-        selectors: impl IntoIterator<Item = FixedBytes<4>>,
-        backend: BackendObj,
-    ) -> Result<Self> {
-        for selector in selectors {
-            if self.routes.contains_key(&selector) {
-                anyhow::bail!("selector {selector:?} is already registered to a backend");
+    pub fn register_backend(mut self, entry: BackendEntry) -> Result<Self> {
+        if self.backends.contains_key(entry.id()) {
```
bobbobbio: again seems like `entry` API could avoid a duplicate HashMap traversal

crates/broker/src/backend.rs
```diff
@@ -130,34 +130,67 @@ pub type BackendObj = Arc<dyn Backend>;
...
+            anyhow::bail!("backend {} is already registered", entry.id());
+        }
+
+        for selector in &entry.selectors {
```
bobbobbio: Same comment about the state of the object when a selector is duplicate

crates/broker/src/backend.rs
```diff
@@ -67,6 +70,190 @@ impl BackendId {
...
+        let backend_b = Arc::new(MockBackend::new("mock_b", vec![selector(1)]));
+
+        let err = BackendRouter::new()
+            .register_backend(BackendEntry::new(vec![selector(1)], backend_a))
+            .unwrap()
+            .register_backend(BackendEntry::new(vec![selector(1)], backend_b))
+            .err()
+            .unwrap();
+
+        assert!(err.to_string().contains("is already registered to backend mock_a"));
```
bobbobbio: I think the missing test to my previous point is one where you have multiple selectors but only one of them is duplicate, and we assert about the state of the object after the error.

crates/broker/src/db/fuzz_db.rs
```diff
@@ -36,6 +37,10 @@ use boundless_market::contracts::{
     Offer, Predicate, ProofRequest, RequestId, RequestInput, RequestInputType, Requirements,
 };
 
+fn test_backend_id() -> BackendId {
+    BackendId::new("risc0_v3").unwrap()
```
bobbobbio: maybe something like "test_backend_id" would make more sense

crates/broker/src/db/sqlite.rs
```diff
@@ -957,6 +971,14 @@ mod tests {
     use risc0_aggregation::GuestState;
     use risc0_zkvm::sha::Digest;
 
+    fn test_backend_id() -> BackendId {
+        BackendId::new("risc0_v3").unwrap()
+    }
+
+    fn other_backend_id() -> BackendId {
+        BackendId::new("risc0_v4").unwrap()
```
bobbobbio: same comment here

bobbobbio (commented):
Regarding the overall design, I think it seems sensible for now, it seems like a pretty organic design based on what was there which is probably okay. I assume we will evolve it further.
If you changed the way the broker interacted with what became the backend in this PR too much it would make this PR even larger.. 

I wonder how these abstracted types will work out in practice when we add a new backend, but its a little hard to predict.

Most of my comments are minor thoughts / observations when reading it

crates/risc0-backend/src/batch.rs
```diff
@@ -566,7 +569,7 @@ impl BatchProcessor for Risc0BatchProcessor {
         );
 
         let mut assessor_secs = None;
-        let assessor_proof_id = if cmd.finalize {
+        let assessor_proof_id: Option<String> = if cmd.finalize {
```
bobbobbio, line 528: I probably would do `cmd.finalize.then(|| {`

crates/risc0-backend/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+repository = { workspace = true }
+publish = false
+
+[package.metadata.release]
+release = false
+
+[features]
+# Exposes test-only constructors (`with_test_batch_processor`, `with_set_verifier`, …) for
+# downstream crates' tests (e.g. the broker's batcher/submitter tests).
+test-support = []
```
bobbobbio: I question if there is really that much harm in just having these functions be public, less features is attractive to me generally

bobbobbio (approved):

bobbobbio:
Future work I'm noticing here as I rebase:

- Update broker tests to not test backed specific stuff (possibly no more use of Prover in broker?)
- Remove all remaining risc0_ crate type usage from broker and prune dependencies

crates/boundless-backend/src/types.rs
```diff
@@ -0,0 +1,444 @@
...
+}
+
+impl From<ClaimDigest> for [u8; 32] {
+    fn from(value: ClaimDigest) -> Self {
+        value.0.into()
+    }
+}
+
+/// Mirrors `BoundlessMarket.sol::fulfill`.
+pub struct AssessorArtifact {
```
jonastheis: Isn't this more like an implementation detail of a `Backend` that has a specific `Assessor` implementation?

jonastheis (approved):

PR #2005, opened 2026-05-20
BM-3072: feat(contracts): add OnChainAssessor — native Solidity assessor adapter

jonastheis:
## Summary

Adds the native Solidity `OnChainAssessor` adapter — a per-fill predicate-eval + ECDSA-prover-binding implementation of `IBoundlessAssessor` — and fixes two correctness gaps surfaced while wiring it end-to-end through the market. Stacked on top of #1982 ([jonas/router-decoupling](https://github.com/boundless-xyz/boundless/pull/1982)).

## The adapter

`contracts/src/router/adapters/OnChainAssessor.sol` is stateless and immutable. Per fulfillment batch it vouches for:

1. **Predicate satisfaction** — for each fill, the supplied `(claimDigest, fulfillmentData)` satisfies the requestor's `Predicate` (`DigestMatch` / `PrefixMatch` / `ClaimDigestMatch`).
2. **Claim-digest binding** — whenever the prover attaches an `ImageIdAndJournal` payload, `(imageId, journal)` must reconstruct to `fill.claimDigest`. Required so a callback dispatched off `fulfillmentData` can't receive bytes the underlying seal never proved.
3. **Prover binding** — `batch.prover` is bound via an EIP-712 ECDSA signature over `(prover, requestDigests[], claimDigests[])`. On-chain equivalent of the R0 STARK adapter's `prover` journal commitment.

Brokers select between this adapter and `R0BoundlessAssessorAdapter` by setting the first 4 bytes of `assessorSeal` to the registered adapter's router selector. The market is unchanged — selection happens at the router seam.

## Correctness fixes uncovered while wiring this in

- **Journal-hash convention.** `Predicate.eval(imageId, journal)` was using `sha256(abi.encode(journal))`, but the off-chain R0 STARK guest, `BoundlessMarketCallback`, and the broker's claim-digest computation all use `sha256(journal)` over the raw bytes. The on-chain side never matched a real fill in production. Both call sites in `Predicate.sol` are now aligned with the canonical convention.
- **`ClaimDigestMatch` callback binding.** The on-chain adapter did not reconstruct the claim digest from `(imageId, journal)` when the predicate was `ClaimDigestMatch` and the prover attached an `ImageIdAndJournal` fulfillment payload (typically to feed a callback). The off-chain assessor guest enforces this binding; the on-chain adapter now matches. Without the guard, a malicious prover could submit a valid seal for the requested claim digest while attaching `(imageId, journal)` bytes that don't match — and the callback would dispatch unproven data.

## Tests

- `contracts/test/router/adapters/OnChainAssessor.t.sol` — 26 self-contained unit tests covering each predicate kind, the new `ClaimDigestMatch + ImageIdAndJournal` reconstruction guard, predicate failures, length / malformed-seal guards, tamper detection on `requestDigest` / `claimDigest` / `fulfillmentData` / `prover`, ERC-165, the domain separator, plus a regression test pinning the type string against the contract's `FULFILLMENT_BATCH_AUTH_TYPEHASH` so a silent rename trips loudly.
- `contracts/test/BoundlessMarket.t.sol` — wires `OnChainAssessor` into the market's router as a third assessor entry, adds `createFulfillmentBatchOnChain` / `createFillsAndSubmitRootOnChain` helpers (the EIP-712 analogue of `createFillsAndSubmitRootR0`), and a `BoundlessMarketOnChainAssessorTest` contract exercising locked-request fulfill, batch fulfill, the `ClaimDigestMatch + ImageIdAndJournal` callback scenarios (both matching and mismatched journal bytes), and a mixed-adapter batch where one fulfillment goes through `OnChainAssessor` and another through `R0BoundlessAssessorAdapter` in the same transaction.
- `contracts/test/router/BenchBase.sol` / `AdapterBench.t.sol` — bench fixture journal-hash aligned with the canonical convention so adapter benches stay green under the fixed `Predicate.eval`.


contracts/src/router/adapters/OnChainAssessor.sol
```diff
@@ -0,0 +1,179 @@
...
+                keccak256("OnChainAssessor"),
+                keccak256("1"),
+                block.chainid,
+                address(this)
+            )
+        );
+    }
+
+    /// @inheritdoc IBoundlessAssessor
+    function verifyAssessor(FulfillmentBatch calldata batch, bytes32[] calldata requestDigests) external view {
```
willpote, line 97: I think we could simplify to something like:

```
for (uint256 i = 0; i < n; i++) {
    Predicate calldata predicate = batch.requests[i].predicate;
    Fulfillment calldata fill = batch.fills[i];
    bool hasImageAndJournal = fill.fulfillmentDataType == FulfillmentDataType.ImageIdAndJournal;

    // Single reconstruction guard: whenever (imageId, journal) is attached,
    // it MUST reconstruct to fill.claimDigest otherwise a downstream
    // callback would dispatch unproven bytes.
    bytes32 imageId;
    bytes calldata journal;
    if (hasImageAndJournal) {
        (imageId, journal) = FulfillmentDataLibrary.decodePackedImageIdAndJournal(fill.fulfillmentData);
        if (ReceiptClaimLib.ok(imageId, sha256(journal)).digest() != fill.claimDigest) {
            revert ClaimDigestMismatch(i);
        }
    }

    // Predicate satisfaction only dispatch differs per predicate type.
    if (predicate.predicateType == PredicateType.ClaimDigestMatch) {
        if (!predicate.eval(fill.claimDigest)) revert PredicateFailed(i);
    } else {
        if (!hasImageAndJournal) revert MissingFulfillmentData(i);
        if (!predicate.eval(imageId, journal)) revert PredicateFailed(i);
    }

    claimDigests[i] = fill.claimDigest;
}

```

This may change which error gets thrown first so test may need refactoring

contracts/src/types/Predicate.sol
```diff
@@ -64,14 +64,18 @@ library PredicateLibrary {
         if (predicate.predicateType == PredicateType.DigestMatch) {
             require(predicate.data.length == 64, "Invalid DigestMatch data length");
             bytes memory dataJournal = Bytes.slice(predicate.data, 32);
-            return bytes32(dataJournal) == sha256(abi.encode(journal)) && bytes32(predicate.data) == imageId;
+            // Journal hash convention: `sha256(journal)` over the raw bytes,
+            // matching what the off-chain R0 STARK guest commits (via
+            // `ReceiptClaim::ok(image_id, journal_bytes)`) and what
+            // `BoundlessMarketCallback` re-derives. No `abi.encode` wrap.
+            return bytes32(dataJournal) == sha256(journal) && bytes32(predicate.data) == imageId;
```
willpote, line 71: We should also fix this on `contracts/shanghai/src/types/Predicate.sol`

In fact that reminds me, we need a plan to handle the shanghai fork also so we can upgrade Taiko. We should track that in Linear

willpote (approved):
Overall looks good, few comments

contracts/src/types/Predicate.sol
jonastheis: We should double check their timeline for this upcoming fork: https://docs.taiko.xyz/protocol/unzen-fork

jonastheis: https://linear.app/boundlessnetwork/issue/BM-3020/port-changes-to-shanghai-fork-of-the-market-to-support-taiko

contracts/src/router/adapters/OnChainAssessor.sol
jonastheis: This is much better. Kind of became a frankenstein bc I added some of the checks later and didn't revisit. Done in 894e8ba07d36c3ab0edae4c3d206a19b40ec6d1b

PR #2018, opened 2026-05-26
BM-3017: Add market-price buffer to requestor auto-pricing

Wollac:
When a request is auto-priced (no explicit `max_price`), the gas part of `maxPrice` already gets a 2x cushion, but the proof part — `min(p99, 2·p50) × cycles` from the indexer — gets none. That's just a snapshot of recent fills, so if the per-cycle market drifts up between submitting and a prover actually locking, the request can sit there unlocked until it expires. Every comparable "set a max, let someone fill it" system keeps some headroom here (ethers does 2x on gas, 1inch Fusion ~2.3% on the swap); we had zero on the proof side.

This adds a configurable buffer on the market-derived max price, default +15%.

**Heads-up, this is the default path, not opt-in.** `ClientBuilder::build()` always attaches a price provider (deployment indexer, or a built-in market-pricing fallback), so any auto-priced request without an explicit price now authorizes ~15% more than before. That's the intent — it's what lets requests lock through price drift — but worth flagging for spend expectations on the CLI default and order-generators without a `--max-price` flag.

A few choices worth calling out:

- **Market price only.** Not applied to an explicit `--max-price` (taken verbatim, like today), the configured `max_price_per_cycle` (that's a deliberate ceiling — pushing it past what you said you'd pay is wrong), or the static fallback (already ~p99). The market estimate is the only piece that actually drifts.
- **15% rather than 10%.** Per-cycle proof prices swing more than ETH gas in % terms and a request can sit minutes before lock, so a bit more than the usual liquid-market cushion.
- **Lives on `OfferLayerConfig`,** next to `max_price_per_cycle`. It's a pricing-strategy knob, not an `Offer` field, so it stays off `OfferParams`.

Gas buffering, min price, timeouts and collateral are untouched.

Tests: unit tests for the buffer math, and the price-provider e2e asserts the default buffer makes `maxPrice` exactly 115/100 of the un-buffered price.

### DIFF
--- a/crates/boundless-market/src/request_builder/offer_layer.rs
+++ b/crates/boundless-market/src/request_builder/offer_layer.rs
@@ -131,6 +131,18 @@ pub(crate) fn resolve_max_price(
     })
 }
 
+/// Applies the market-price buffer to a per-cycle market price and scales by the cycle count.
+///
+/// `buffer_percentage` is a multiplier where `100` = no buffer and `115` = +15%. Multiplies
+/// before dividing to avoid truncation; any sub-wei fraction rounds down.
+pub(crate) fn buffered_market_max(
+    max_per_cycle: U256,
+    cycle_count: u64,
+    buffer_percentage: u64,
+) -> U256 {
+    max_per_cycle * U256::from(cycle_count) * U256::from(buffer_percentage) / U256::from(100)
+}
+
 struct CollateralRecommendation {
     default: U256,
     large: U256,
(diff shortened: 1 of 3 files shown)


crates/boundless-market/src/request_builder/offer_layer.rs
```diff
@@ -445,6 +476,8 @@ impl OfferParams {
             lock_timeout: Some(offer.lockTimeout),
             bidding_start: Some(offer.rampUpStart),
             ramp_up_period: Some(offer.rampUpPeriod),
+            // Reconstructing an existing on-chain offer must not re-apply the buffer.
+            market_price_buffer_multiplier_percentage: None,
```
willpote: This shouldn't really live on the `OfferParams` as it is intended to map cleanly to Offer on the contract. 

The convention we have for stuff like this is to put it in the Config struct, e.g. `OfferLayerConfig`

Wollac: addressed in 54c6ea21bcf5b665eddaaa0ba147da1572e861b9

willpote (approved):

PR #2019, opened 2026-05-27
BM-3018: Trim l-prod-8453 OGs + alarm tune

capossele:
- Drop `random-requestor` and `evm-requestor` from `l-prod-8453` (commented out their `PRIVATE_KEY` configs).
- Cut onchain/offchain order generator rates by 25% (intervals 314→419s, 257→343s).
- Widen og_onchain `submissionRate` window 30min→60min for parity with og_offchain and more headroom.
- Widen og_offchain `successRate` window 30min→60min so the 50% threshold has enough samples per window.
- Remove stale "submit every 5 mins" comment.

### DIFF
--- a/infra/indexer/alarmConfig.ts
+++ b/infra/indexer/alarmConfig.ts
@@ -297,11 +297,10 @@ export const alarmConfig: ChainStageAlarms = {
           ],
           successRate: [
             {
-              // Since current submit every 5 mins, this is >= 2 failures an hour
-              description: "less than 50% success rate for two 30 minute periods in 2 hours from og_offchain",
+              description: "less than 50% success rate for two 60 minute periods in 4 hours from og_offchain",
               severity: Severity.SEV2,
               metricConfig: {
-                period: 1800
+                period: 3600
               },
               alarmConfig: {
                 threshold: 0.50,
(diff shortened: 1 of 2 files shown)


infra/indexer/alarmConfig.ts
```diff
@@ -332,10 +331,10 @@ export const alarmConfig: ChainStageAlarms = {
             //   }
             // },
             {
-              description: "no submitted orders in 30 minutes from og_onchain",
+              description: "no submitted orders in 60 minutes from og_onchain",
               severity: Severity.SEV2,
               metricConfig: {
-                period: 1800
+                period: 3600
```
willpote: Given the self-throttling is now working, we may want to be even more lenient on this (maybe 4 hours? which is about how long each offchain order takes to hit lock expiry)

willpote (approved):

infra/indexer/alarmConfig.ts
capossele: yeah makes sense: [8169ae4](https://github.com/boundless-xyz/boundless/pull/2019/commits/8169ae4d75527c829c355f59e57d531103d1eb24)

PR #2039, opened 2026-06-10
BM-3052: risc0 SDK decoupling

bobbobbio:
Decouple risc0 specific proving types from `StandardRequestBuilder` and `ClientBuilder`. Introduce `ZkvmOps` trait that lets users specify which ZKVM they want to build proofs for (`Risc0ZkvmOps` being the one for risc0).

Move many preflight, local execution, and image id creation bits to be behind this trait. Also limit usage of `risc0` types in boundless-market crate and use new ZKVM agnostic `Digest` and `Journal` types.

The boundless-market tests, blake3, and contract related code continue to use risc0 types and will need to be addressed in future work.


crates/boundless-market/src/client.rs
```diff
@@ -874,18 +911,21 @@ pub struct Client<
...
     /// [FundingMode::MinMaxBalance] can be used to maintain a minimum balance by funding requests accordingly.
     pub funding_mode: FundingMode,
+
+    /// The ZKVM specific functions and types
+    pub zkvm_ops: Option<Z>,
 }
 
 /// Alias for a [Client] instantiated with the standard implementations provided by this crate.
-pub type StandardClient = Client<
+pub type StandardClient<Z> = Client<
```
capossele, line 923:
```suggestion
pub type StandardClient<Z> = Client<
      Z,
```

crates/boundless-market/src/client.rs
```diff
@@ -502,21 +505,21 @@ impl<U, D: StorageDownloader, S> ClientBuilder<U, D, S> {
...
 
         if let Some(timeout) = self.tx_timeout {
             client = client.with_timeout(timeout);
         }
 
-        if let Some(skip_preflight) = self.skip_preflight {
-            client = client.with_skip_preflight(skip_preflight);
-        }
+        let skip_preflight = self.skip_preflight.unwrap_or_else(|| client.zkvm_ops.is_none());
+        client = client.with_skip_preflight(skip_preflight);
```
capossele:
```suggestion
        match self.skip_preflight {
              Some(skip) => client = client.with_skip_preflight(skip),
              None if client.zkvm_ops.is_none() => client = client.with_skip_preflight(true),
              None => {} // leave None so BOUNDLESS_IGNORE_PREFLIGHT fallback applies
          }
```

crates/boundless-market/src/input.rs
```diff
@@ -174,34 +156,6 @@ impl GuestEnvBuilder {
...
-    ///     b: u32,
-    /// }
-    ///
-    /// let input1 = Input{ a: 1, b: 2 };
-    /// let input2 = Input{ a: 3, b: 4 };
-    /// let input = GuestEnv::builder()
-    ///     .write(&input1).unwrap()
-    ///     .write(&input2).unwrap();
-    /// ```
-    pub fn write<T: Serialize>(self, data: &T) -> Result<Self, Error> {
```
capossele, line 201: wouldn't be better keeping this one in one place behind a `risc0`-feature-gated extension trait (or `ZkvmOps::encode_input`) so callers aren't each importing risc0 serde? Not blocking, just flagging the trade-off

bobbobbio: yeah I can see that making sense, I guess I thought this as simpler conceptually but does lead to more verbose callsites.

capossele (approved):

bobbobbio:
I manually tested that the broker against staging in listen-only mode just as a sanity check, things seem to work

bobbobbio:
(I had to force-push to rewrite some commits missing a signature)

PR #2040, opened 2026-06-11
BM-3077: feat(broker): select the assessor per batch from the on-chain router registry

jonastheis:
## Summary

Adopts the on-chain `BoundlessRouter` model inside the broker: a request's signed verifier selector resolves to a verifier class, the class's `requiredAssessorClass` names the assessor class, and the broker selects the highest-priority candidate assessor registered there — preferring the native `OnChainAssessor` (an EIP-712 signature) over proving the R0 STARK assessor guest. Previously the guest assessor was wired but effectively unused and the assessor choice was a broker-wide config knob; now it is derived per batch from the router registry, and skipping the guest makes the on-chain assessor actually pay off.

```mermaid
flowchart LR
    A[signed verifier selector] -->|"resolve_verifier_class<br/>(sentinel / class id / entry)"| B[verifier class]
    B -->|requiredAssessorClass| C[assessor class]
    C -->|"ASSESSOR_PRIORITY<br/>(first registered wins)"| D{selected assessor}
    D -->|"0x00000022"| E["OnChainAssessor:<br/>sign FulfillmentBatchAuth,<br/>skip assessor guest"]
    D -->|"0x00000024"| F["R0 guest:<br/>prove assessor STARK,<br/>set-inclusion seal"]
```

## Main changes

**SDK (`boundless-market`)**
- `RouterRegistry`: snapshots the router (`entries` / `classes` / `defaultClassId`) once at startup via `load_router_registry`; afterwards verifier-class and assessor resolution are pure in-memory lookups — no per-submission RPC.
- Canonical assessor selector constants: `ONCHAIN_ASSESSOR_SELECTOR` (`0x00000022`, stable protocol convention) and `R0_ASSESSOR_SELECTOR` (`0x00000024`, version-coupled to the assessor guest the broker ships — same coupling as verifier selectors).
- `build_onchain_assessor_seal` / `FulfillmentBatchAuth`: the EIP-712 seal for the `OnChainAssessor` adapter.

**Backend boundary (`boundless-backend`)**
- `RouterPolicy`: backend-agnostic policy over a registry snapshot, parameterized by the backend's capabilities (producible verifier selectors, candidate assessors in priority order). Answers: which verifier selector to emit for a signed selector, which assessor seals a batch, which selectors/classes are supported.
- `Backend::assessor_group` (required method): the batch grouping key. `Ok(None)` = backend doesn't distinguish assessor classes; `Err` = order unresolvable and must not be batched.

**RISC0 backend (`risc0-backend`)**
- Construction goes through `Risc0Backend::from_deployment`, which always snapshots the registry (including listen-only mode) — the policy is mandatory, never `Option`, so tests and production run the same code path.
- The assessor is selected per batch from its orders' verifier class; the assessor guest STARK is neither proven nor aggregated when the on-chain assessor is selected (the set-builder run becomes optional for all-direct-submit batches).
- `supported_selectors` only advertises selectors/classes the backend can produce *and* seal (a candidate assessor must be registered for the class).

**Broker**
- A fulfillment batch carries one assessor seal, so batches stay single assessor class: the open batch locks to one assessor group and orders of other groups are requeued for a later batch. Verifier classes sharing one `requiredAssessorClass` still co-batch.
- Orders with no resolvable assessor are failed loudly (`[B-AGG-602]`) instead of silently deferred or batched into an unsealable batch.

**Tests / wiring**
- `boundless_test_utils::market::test_router_registry`: in-memory fixture mirroring the deployed test router topology, so unit tests exercise the real resolution logic over canned data.
- Localnet registers the `OnChainAssessor` (`scripts/localnet-deploy.sh`, `Manage.Router.s.sol::RegisterOnChainAssessor`); broker e2e covers signing the default sentinel, a verifier class id, and specific entry selectors.

## Stack

Builds on #2005 (OnChainAssessor adapter) and #2020 (legacy ABI fallback), both stacked on #1982 (router decoupling). Until those merge forward, this PR's diff includes the parent branches — the change introduced here is the head commit.


willpote (approved):

crates/broker/src/batcher/service.rs
```diff
@@ -430,6 +437,146 @@ impl BatcherService {
...
+    /// Partition resolved orders by whether their assessor group matches `target`: matching orders
+    /// are returned, the rest are requeued to `requeue_status` (the status they were claimed from)
+    /// so a later batch of their group picks them up.
+    async fn keep_group_or_requeue(
+        &self,
+        backend_id: &BackendId,
+        target: FixedBytes<4>,
+        orders: Vec<(BatchReadyOrder, Option<FixedBytes<4>>)>,
+        requeue_status: OrderStatus,
+    ) -> Result<Vec<BatchReadyOrder>, BatcherErr> {
```
willpote: Feels wrong that we claim the orders, then have to requeue them.

May also leave things in an inconsistent state if we crash in between claim and re-queue (haven't confirmed this, just a suspicion)

May make sense to first get the assessor group,  then get + filter all ready orders (needs new db method), and then only claim orders from that group

jonastheis: done in ebd3263

PR #2052, opened 2026-06-25
fix(contracts): guard open-path fulfillment against copied-proof front-running

jonastheis:
## Summary

Remediation for an audit finding on `OnChainAssessor`. This description first frames the **general problem** (proof front-running) and the **solution space**, then describes the fix shipped here: a commit-reveal gate on the open fulfillment paths.

## The general problem: proof front-running

A ZK proof / seal is a **bearer artifact**: it attests a *result*, not *who produced it*. But producing it is the expensive work, and in an open marketplace whoever submits it on-chain first collects the payment.

So unless the protocol binds the proof to its producer, anyone who observes a submitted (or pending) proof can copy it, present it as their own, and steal the fulfillment reward — without doing any of the work. The gap is the missing, unforgeable link between *the proof* and *the prover who should be paid*.

Any fix has to establish that binding **before the proof becomes a public, actionable artifact**. There are only three places to anchor it:

- **Authorize** — pin the eligible prover in on-chain state before the proof exists (e.g. locking).
- **Hide / delay** — keep the proof unusable-by-others until it's attributed to the prover (e.g. commit-reveal).
- **Bind-in-proof** — make the proof itself commit to the prover (e.g. in-circuit, or a trusted attestation).

## Where it applies

By definition the problem exists only on the **open fulfillment paths**, where the protocol has no prior on-chain binding of the prover:

- **Never-locked** (priced-at-fulfill, e.g. `priceAndFulfill`) → **vulnerable**: full payment goes to whoever submits.
- **After lock deadline** (was-locked, anyone may fulfill) → **vulnerable**: whoever submits is recorded as the fulfiller. The auction price is ~0 here, but the real prize is the **slashing reward**: the late fulfillment overwrites `lock.prover` with the submitter, so when `slash` is later called (after the full deadline) that submitter receives **50% of the defaulted locker's collateral**. So this path is *not* low-value — the stake reward is typically larger than a single proof's price, capped at 50% of the locker's collateral.

**Not a problem:**

- **Locked, before lock deadline** → already protected. The lock pre-authorizes the prover (`lock.prover == batch.prover` is enforced), so a copied proof can't redirect payment. This is the **authorize** family, already in place.

(`R0BoundlessAssessorAdapter` is out of scope here: its prover binding lives in the STARK journal, so rebinding requires regenerating the proof — a *soft*, proving-time deterrent rather than a hard guarantee.)

## The solution space

Four properties you'd want from a fix:

- **Hard** — security does not depend on the attacker's resources or speed. (Opposite = *soft*: a margin an attacker erodes with faster proving or more hashpower — e.g. R0's ~14s, or PoW.)
- **Trustless** — no off-chain party you have to trust.
- **Open** — permissionless: anyone who actually produced a proof can fulfill and get paid, with no pre-registration (the never-locked "whoever finishes first wins" model survives).
- **Cheap** — no extra transaction, no added block latency, no recurring compute.

You can't have all four — each option gives up exactly one:

| Option (family) | Hard | Trustless | Open | Cheap |
|---|:---:|:---:|:---:|:---:|
| **Commit-reveal** (hide / delay) | ✅ | ✅ | ✅ | ❌ +1 tx, +1 block |
| **Lock / claim** (authorize) | ✅ | ✅ | ❌ *closes the set* | ⚠️ +collateral to be grief-safe |
| **Attestation** (bind-in-proof, trusted) | ✅ | ❌ *trust attestor* | ✅ | ✅ |
| **R0 in-circuit / PoW** (bind / delay, soft) | ❌ *soft margin* | ✅ | ✅ | ✅ |

**Commit-reveal is the only option that is hard *and* trustless *and* open** — it buys all three by spending latency. R0-in-circuit and PoW are the same idea at the cheap end (a soft time/compute margin an attacker can erode); attestation is hard + open but reintroduces a trusted signer; locking is hard + trustless but closes the open set (and needs collateral to be grief-safe).

## The fix in this PR: commit-reveal on the open paths

Before settling any never-locked or was-locked fill, `fulfill` requires a prior `commitFulfillment(keccak256(abi.encode(fulfillmentBatches)))` recorded in a **strictly earlier block**:

```
 block N     commitFulfillment( keccak256(abi.encode(fulfillmentBatches)) )   ← only a hash; seal stays hidden
 block N+1   fulfill / priceAndFulfill / submitRoot…  → commitment at N (< N+1) ✓ → settles

 a front-runner who first sees the seal at N+1 cannot have a block-N commitment for it,
 and cannot fabricate one in the same block (strictly-earlier rule) → copied proof rejected
```

- **Binding the seals** makes the commitment un-precomputable — you can't commit without already holding the proof.
- The **strictly-earlier-block** rule (`COMMIT_REVEAL_MIN_BLOCKS = 1`) defeats same-block / sequencer-reordering attempts. One block suffices on a single sequencer; it's a named constant so it can be raised on reorg-prone chains.
- Applied **uniformly** to all open-path fulfillment regardless of assessor — the market can't cheaply tell which assessor a fill used at settle time, and this also hardens R0's soft proving-time margin. The **locked-before-deadline path is unaffected and stays single-tx.**

This is the hide/delay anchor with the cheapest possible clock (block height): hard, trustless, open — paying only ~1 block of latency on the two open paths.

## Changes

- `BoundlessMarket.sol`: `commitFulfillment`, `_hasOpenPathFill`, `_consumeCommitment`, and the gate in `fulfill`. New storage slot `fulfillmentCommitBlock` appended after `imageUrl`.
- `IBoundlessMarket.sol`: `commitFulfillment` + `MissingFulfillmentCommitment` (so the SDK/broker bindings expose it). Regenerated SDK artifact + `bytecode.rs`.
- Storage-layout parity preserved: slots 0/1/2 (shared with the legacy impl via the delegatecall fallback) are unchanged; `verify-storage-layout.py` passes.
- Contract size: **+394 B** (22,890 B runtime; 1,686 B under the EIP-170 limit).

## Follow-ups (not in this PR)

- **Broker/SDK**: brokers must `commitFulfillment` one block ahead before fulfilling never-locked / was-locked orders. Separate broker-crate change.

Builds on #1982 (router decoupling) and #2005 (`OnChainAssessor`).


unboxedtype:
The new commit/reveal flow blocks the straightforward version of the attack: an attacker can’t just copy a pending open-path fulfillment, change batch.prover, and reveal it, because they don’t have a prior commitment for that modified batch. But there seems to be another route, we've just edited this into a step-by-step plan to make it more explicit.


1. The honest prover prepares a valid open-path fulfillment for a never-locked request. 
The honest prover calls `commitFulfillment()` for the exact `FulfillmentBatch[]`.

2. In a later block, the honest prover submits the reveal transaction with `priceAndFulfill()`.
Attacker sees this reveal transaction in the mempool. Before the reveal is included, the attacker calls `lockRequest()` for the same request. The request is now locked to the attacker, so it is no longer treated as an open-path fulfillment.

3. Honest prover’s `priceAndFulfill()` is processed after the attacker’s lock. The honest fulfillment does not revert, but it returns `RequestIsLocked` as the payment error. The honest prover receives no payment, even though their proof was accepted as a delivered proof.

4. Attacker copies the honest prover’s proof material. Attacker creates a new fulfillment batch with `batch.prover = attacker`. Attacker recomputes the `OnChainAssessor` signature for their own prover address.

5. Attacker calls `fulfill()` as the locked prover. Because the proof material is not bound to the prover/payment recipient, verification passes. The attacker receives the request payment and gets their lock collateral back.


cc: @jonastheis

jonastheis:
**Confirmed, the route is valid.** A locked-before-deadline fill carries no commitment requirement (`_hasOpenPathFill` returns false for it), and `_fulfillAndPayLocked` only gates payment on `lock.prover`/`isProverPaid`, not on who delivered the proof. So an attacker who front-runs the lock becomes `lock.prover`, re-binds the copied proof, and collects the fee — exactly as described. The honest prover's reveal still marks the request `fulfilled` but leaves the payment slot open, which the attacker then claims. The commit-reveal gate from this PR does not cover this, because both the lock and the locked-path fulfillment need no commitment.

**Scope: this is reachable on exactly one path.** For the bypass to work the path must be *both* (a) an open path — no prior `lock.prover` binding — *and* (b) still lockable, so the attacker can acquire that binding. Only the **never-locked, pre-lock-deadline** path is both:

| Open path | Lockable now? | Payment at stake | Lock-then-copy bypass |
|---|---|---|---|
| never-locked, pre-lock-deadline | **yes** | client price | **yes** |
| never-locked, post-lock-deadline | no (`RequestLockIsExpired`) | none — `priceAt` = 0 | n/a (nothing to steal) |
| was-locked (lock expired) | no (locked flag is sticky, set with `\|=` and never cleared) | stake reward + Δprice | no — the lock can't be acquired |

So the commit-reveal gate remains load-bearing for the **was-locked** path (where you cannot lock first to protect yourself, and the attacker cannot acquire a lock either). The new route only touches the never-locked-pre-deadline window.

**On that one window, this is inherent to fulfilling without a lock, not a defect we can patch away.** A proof is a bearer artifact — it proves the claim, not who produced it — so the only thing that grants a prover the exclusive right to payment is *holding the lock*. The standard flow, both as our recommendation and as implemented in the proving software, is **lock first, then fulfill**: a held lock binds the request to you via `lock.prover` and cannot be acquired by anyone else, which fully closes this route. Choosing to fulfill a still-lockable request *without* locking is, by definition, giving up that exclusivity — and the front-running exposure is the cost of that choice.

**We deliberately decline to enforce a contract-level fix here, because the cure is worse than the disease.** The only way to protect a never-locked fulfillment from a later lock is to let an *earlier* fulfillment commitment override a *later* lock. But that directly destroys the exclusivity guarantee that makes locking useful: a prover who locks in good faith — unable to see any commitment, since it's just a hash — could be bumped at reveal by anyone who committed earlier, wasting their proving work and opening a new griefing surface against honest lockers. Trading a front-run that has a complete, standard, zero-extra-cost defense (lock first) for a permanently weaker, fuzzier exclusivity guarantee is a net loss.

**Disposition:** documented as a deliberate trade-off of the no-lock path. We've added a NatSpec note on `priceRequest` (the "fulfill without locking" entry point) spelling out that this path is non-exclusive by definition and that locking first is the exclusive, front-running-immune flow. No contract change; the commit-reveal gate stays in place for the was-locked path it actually protects.

Note this is a continuation of the [prior audit](https://github.com/boundless-xyz/boundless-security/blob/main/audits/2025_04_Veridise%20(Boundless%20Core).pdf)'s V-RISC0-VUL-006 (Low/Usability, Acknowledged), where the no-lock-no-payment case was already accepted with "always lock before fulfilling — the broker's default".

cc @unboxedtype

PR #2056, opened 2026-07-01
BM-3074: fix(examples): unbreak nightly examples (ZkvmOps, via_ir, Send)

jonastheis:
## Summary

The [Nightly Examples](https://github.com/boundless-xyz/boundless/actions/workflows/nightly-examples.yml) workflow has been failing on `main` after recent merges. The examples are separate cargo/foundry workspaces built only by the nightly schedule (never on PRs), so three independent regressions landed green and broke the nightly. This PR fixes all three.

### 1. `cargo build` — `NotProvided: ZkvmOps is not satisfied`
The risc0 SDK decoupling (#2039) made `Client` / `ClientBuilder` / the request builder generic over a `ZkvmOps` type parameter, defaulting to `NotProvided`. `new_request()` / `submit()` / `upload_program()` now require `Z: ZkvmOps`. The main workspace (CLI) was migrated; the examples were not.

**Fix:** add `.with_zkvm(Risc0ZkvmOps::new().await)` to each example's client builder (matching `crates/boundless-cli/src/config.rs`).

### 2. `forge test` — `Stack too deep` in `BoundlessRouter.sol`
The BoundlessRouter change (#1982) added `FulfillmentBatch.sol`, which transitively pulls the router implementation into the smart-contract-requestor example's contract compile. That example's `foundry.toml` lacked `via_ir`, so solc 0.8.26 hit "Stack too deep".

**Fix:** enable `via_ir` + optimizer in the example's `foundry.toml`, matching the root contracts config.

### 3. `cargo nextest` — `Send is not general enough` (request-stream test)
The new request builder's higher-ranked (`for<'a>`) bounds make `run()`'s future fail `Send` inference, so the test's `tokio::spawn(run(...))` no longer compiles.

**Fix:** drive `run()` inline via `select!` (polled on the current task, so no `Send` bound; dropping it on timeout cancels it) — preserving the test's behavior.


bobbobbio (approved):
Thanks for fixing this. 

Not an expert in the smart-contracts stuff, but that fix looks reasonable to me too FWIW

jonastheis:
Latest manual run on the branch passed: https://github.com/boundless-xyz/boundless/actions/runs/28570032143

PR #2067, opened 2026-08-03
fix(order-stream): make WebSocket cleanup generation-aware

### LINKED ISSUE #2066: [order-stream] Stale WebSocket cleanup can remove a replacement connection
## Summary

The order-stream WebSocket implementation supports reconnecting with the same address by replacing the existing entry in `state.connections`.

However, both WebSocket cleanup and broadcast cleanup remove connections by address only. This creates a race where cleanup belonging to an older connection can remove a newer replacement connection.

## Race condition

1. Connection A is stored at `connections[address]`.
2. Connection B reconnects with the same address and replaces A.
3. A later exits, or a broadcast task detects A's sender as closed.
4. Cleanup calls `connections.remove(&address)`.
5. The entry for B is removed instead of A.

As a result, B's WebSocket may remain open while no longer being present in the server-side connection map. Subsequent orders will not be broadcast to B, and connection capacity statistics become inaccurate.

## Additional observation

`drop(old_connection.sender)` does not necessarily terminate the old WebSocket task because `websocket_connection` still retains its own `sender_channel` clone.

## Proposed fix

Assign every connection a unique ID or generation number:

```rust
struct ClientConnection {
    id: u64,
    sender: mpsc::Sender<String>,
}
```

Cleanup should remove the entry only when the stored connection ID matches the ID of the task performing cleanup. Broadcast snapshots should also retain and validate this ID.

The replacement path should additionally use an explicit cancellation mechanism, or avoid retaining an extra sender clone, so that the old WebSocket task is reliably terminated.

## Tests

Add deterministic tests covering:
- stale cleanup of A after B replaces A;
- stale broadcast cleanup of A after B replaces A;
- confirmation that B remains in the connection map.

grandpig:
## What

Make WebSocket connection cleanup identity-aware.

Fixes https://github.com/boundless-xyz/boundless/issues/2066

## Why

Connection cleanup currently removes entries by address only. When a client reconnects with the same address, cleanup from the previous connection can remove the replacement connection. Broadcast cleanup has the same race.

## Changes

- Add a unique connection ID/generation.
- Remove connections only when the stored ID matches the cleaning task.
- Make broadcast cleanup generation-aware.
- Ensure the replaced WebSocket task is explicitly terminated.
- Add deterministic regression tests.

## Testing

- `cargo test -p order-stream`

### DIFF
--- a/crates/order-stream/src/lib.rs
+++ b/crates/order-stream/src/lib.rs
@@ -486,9 +486,9 @@ impl AppState {
     }
 
     /// Removes connection for a given address.
-    pub(crate) async fn remove_connection(&self, addr: &Address) {
+    pub(crate) async fn remove_connection(&self, addr: &Address, connection_id: u64) {
         let mut connections = self.connections.write().await;
-        connections.remove(addr);
+        ws::remove_connection_if_current(&mut connections, addr, connection_id);
     }
 }
 
--- a/crates/order-stream/src/ws.rs
+++ b/crates/order-stream/src/ws.rs
@@ -29,17 +29,49 @@ use boundless_market::{
 use futures_util::{SinkExt, StreamExt};
 use rand::{seq::SliceRandom, Rng};
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{
+    atomic::{AtomicU64, Ordering},
+    Arc,
+};
 use tokio::{sync::mpsc, task::JoinHandle};
+use tokio_util::sync::CancellationToken;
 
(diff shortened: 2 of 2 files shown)


grandpig:
@capossele Just following up on this in case it got buried. This fixes the reconnect cleanup race so a stale WebSocket task can't remove its replacement connection, including the same race in broadcast cleanup. Deterministic regression tests are included. Thanks!

grandpig:
@austinabell Hi, Could you please review this PR at your convenience? Thank you very much.

PR #2068, opened 2026-08-19
fix(order-stream): preserve Redshift telemetry field compatibility

zloglevel:
## Summary

- Add migration 005 to align `request_completions` with the current broker telemetry payload.
- Expose `committed_to_application_proof_duration_secs` and `actual_total_proving_time_secs`.
- Preserve `proving_duration_secs` and `actual_proving_time_secs` as compatibility aliases.
- Fall back to legacy JSON keys so telemetry ingested before migration 3 remains queryable.
- Remove reliance on the obsolete `preflight_cache_hit` field.
- Extend the Redshift telemetry integration test to validate the new fields and aliases.

## Context

Migration 002 still referenced fields that are no longer emitted by the current Rust telemetry structs. Migration 003 corrected the schema, but removed historical completion column names that may still be used by dashboards and BI queries.

This migration is intentionally additive and does not modify already-applied migrations.

## Testing

- `cargo fmt --all -- --check`
- `git diff --check`
- `RISC0_SKIP_BUILD_KERNELS=1 RISC0_SKIP_BUILD=1 cargo test -p order-stream --test telemetry --no-run`
- `RISC0_SKIP_BUILD_KERNELS=1 RISC0_SKIP_BUILD=1 cargo test -p order-stream --test telemetry telemetry_end_to_end -- --nocapture`

The Redshift end-to-end assertions were compiled successfully. The live Redshift test was skipped locally because `ORDER_STREAM_URL` and Redshift credentials were not configured.

### DIFF
--- a/crates/order-stream/tests/telemetry.rs
+++ b/crates/order-stream/tests/telemetry.rs
@@ -231,7 +231,8 @@ async fn telemetry_end_to_end() {
     let mut eval_found = false;
     for attempt in 1..=max_attempts {
         let row = sqlx::query(
-            "SELECT broker_address, outcome FROM telemetry.request_evaluations \
+            "SELECT broker_address, outcome, preflight_duration_ms \
+             FROM telemetry.request_evaluations \
              WHERE request_id = $1 \
              LIMIT 1",
         )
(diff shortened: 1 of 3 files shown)


zloglevel:
@capossele Could you please take a look when you have a chance? Thanks!

zloglevel:
@capossele  Could you please take a look when you have a chance? Thanks!

